PR #74, opened 2014-11-22
Lazy distance comparison for NonSizedIterable

morinmorin:
I often need to compare the size of a (possibly large) bidirectional range `rng` with a small number `n`. When the range is large, computing the "full" length (i.e. `distance(rng)`) is inefficient for this purpose; it's enough to compute "partial" length at most `n` or `n + 1`. 

I came up with the following solutions to this problem: 
### A. "drop at most" range adaptor

Oven's [`dropped(n)`](http://p-stade.sourceforge.net/oven/doc/html/oven/range_adaptors.html#oven.range_adaptors.dropped) range adaptor drops at most `n` elements; it does not penetrate the end of the range even if the range has size less than `n`. (To differentiate it from range-v3's `drop` adaptor, I call it `drop_at_most`.) With this adaptor, we can make comparison without computing the "full" length. For example, 
  `distance(rng) < n`  <--> `empty(rng | drop_at_most(n - 1))`
  `distance(rng) > n`  <--> `!empty(rng | drop_at_most(n))`
### B. "bounded distance" function

`bounded_distance(rng, bound)` returns the size of `rng` if the size is less than `bound`, and returns `bound` otherwise. This function increments an iterator at most `bound` times. For example, 
  `distance(rng) < n`  <--> `bounded_distance(rng, n) < n`
  `distance(rng) > n`  <--> `bounded_distance(rng, n + 1) > n`
### C. "lazy distance" proxy

`lazy_distance(rng)` returns a proxy that can be compared with a number. Actual computation of "partial" length is done when a proxy is compared. For example, 
  `distance(rng) < n`  <--> `lazy_distance(rng) < n`
  `distance(rng) > n`  <--> `lazy_distance(rng) > n`
<br>

However, IMHO, Option A is not so intuitive and Option B is verbose (we need to write `n` twice). I like Option C _very much_, but this approach might be too complex for a simple task and it might be difficult to get into the Standard. 

Then I sought another solution and found that the [GHC API](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/ghc/Util.html) has interesting function: 

```
listLengthCmp :: [a] -> Int -> Ordering
```

I use this function interface as a fourth solution. 
### D. "lazy distance comparison" function

`distance_compare(rng, n)` returns positive (negative) integers if the size of `rng` is larger (smaller) than `n` and returns 0 if the size equals `n`. For example, 
  `distance(rng) < n`  <--> `distance_compare(rng, n) < 0`
  `distance(rng) > n`  <--> `distance_compare(rng, n) > 0`
<br>

Option D is less readable than Option C, but it has good balance between the conciseness of the approach and the readability. 

This PR implemented `distance_compare` (Option D) and tests. What do you think?

### DIFF
--- a/include/range/v3/distance.hpp
+++ b/include/range/v3/distance.hpp
@@ -90,10 +90,18 @@ namespace ranges
         struct distance_compare_fn : iter_distance_compare_fn
         {
         private:
-            template<typename Rng>
+            template<typename Rng,
+                CONCEPT_REQUIRES_(!is_infinite<Rng>())>
             int impl_r(Rng &rng, range_difference_t<Rng> n, concepts::Iterable*) const
             {
                 return iter_distance_compare(begin(rng), end(rng), n);
+            }
+            template<typename Rng,
+                CONCEPT_REQUIRES_(is_infinite<Rng>())>
+            int impl_r(Rng &rng, range_difference_t<Rng> n, concepts::Iterable*) const
+            {
+                // Infinite ranges are always compared to be larger than a finite number.
+                return 1;
             }
             template<typename Rng>
             int impl_r(Rng &rng, range_difference_t<Rng> n, concepts::SizedIterable*) const
--- a/test/distance.cpp
+++ b/test/distance.cpp
@@ -54,8 +54,16 @@ void test_infinite_range(Rng&& rng)
     CHECK(distance_compare(rng,-1) > 0);
     CHECK(distance_compare(rng, 1) > 0);
     CHECK(distance_compare(rng, (std::numeric_limits<range_difference_t<Rng>>::min)()) > 0);
-    // Comparing an infinite range with a huge number might take a significant amount of time.
-    // CHECK(distance_compare(rng, (std::numeric_limits<range_difference_t<Rng>>::max)()) > 0);
+    if (is_infinite<Rng>::value) {
+        // For infinite ranges that can be detected by is_infinite<Rng> traits,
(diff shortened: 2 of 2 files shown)


morinmorin:
I just learned in #141 that `InfiniteRange` (ranges that satisfy `is_infinite<Rng>`) does not necessarily have an infinite number of elements. The optimization in `distance_compare` (i.e. `distance_compare(rng, n)` for `InfiniteRange` _early-exits_ with a _positive_ return value) works badly for ranges such as
```
view::ints(0) | view::take_while([](int i){ return i < 5; })
```
Should we remove the optimization or need `PossiblyInfiniteRange` things?

ericniebler:
Ouch, yeah. I'm starting to be of the opinion that `is_infinite` should return an enum: `yes`, `no`, and `maybe`.

ericniebler:
> I just learned in #141 that `InfiniteRange` (ranges that satisfy `is_infinite<Rng>`) does not necessarily have an infinite number of elements. The optimization in `distance_compare` (i.e. `distance_compare(rng, n)` for `InfiniteRange` *early-exits* with a *positive* return value) works badly for ranges such as [...]

Commit f3dfcd1 fixes this issue. `is_infinite<Rng>` is now true *only* when the range is in fact infinite.

morinmorin:
Nice! I'll submit a PR that adds `distance_compare` testcase for "maybe" infinite ranges (which are actually finite) soon.

PR #104, opened 2015-02-20
Removing unused template parameters from for_each

anders-sjogren:
Template parameters V and X seem to be no longer used, since their former use is nowadays pushed into IndirectInvokable concept.


ericniebler:
Thanks

anders-sjogren:
I'm glad I could help!

PR #105, opened 2015-02-20
Use meta implementation from its repository

gnzlbg:
This switches the range-v3 implementation of meta to use that of the meta-repository.


gnzlbg:
I tested this under MacOS with libc++, and the `FindMeta.cmake` script is able to find meta correctly without hints in `/usr/local/include` and `/usr/include` without problems. 

This should work under linux too (travis-ci will say if it works once the newest pull-request are merged into the meta repository). 

I hope that CMake will do the right thing on Windows. If it doesn't it will always pull the latest meta from github. A workaround is to specify the meta directory directly with `cmake -DMETA_INCLUDE_DIR=/path/to/meta/include` but a longer term solution is to add specific Windows `HINTS` to `find_path`.

ericniebler:
What happens when I update the meta repository, but users of range-v3 don't know to pull meta?

gnzlbg:
> What happens when I update the meta repository, but users of range-v3 don't know to pull meta?

Nothing. Unless they update range-v3 afterwards and don't update meta, in which case the compiler will remind them that they should update meta too.

ericniebler:
I prefer none of the above. Meta stays where it is.

gnzlbg:
Ok. I will make a pull request with the documentation changes I made in meta to range-v3. That is going to help me keeping both in sync.

ericniebler:
Yeah, submodules get messy. There's no real way to make it transparent for users. I appreciate the extra work it takes to keep the two repos in sync, but range-v3 currently has no dependencies that need to managed explicitly, and that's worthwhile, IMO. Thanks.

gnzlbg:
I can totally understand that, I also appreciate having no dependencies.

PR #111, opened 2015-03-10
Add missing header fill_n.hpp to algorithm.hpp.

cartec:
Fairly self-explanatory - I assume this omission is unintentional.


ericniebler:
Nice spot, thanks.

PR #114, opened 2015-03-17
biicode support

Manu343726:
Note there's no travis deployment. I don't want to pollute your test for something that you have to modify  anyway. Use the `.travis.yml` from Meta as a template, it should be exactly the same. Ask me if you have any doubts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,21 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
 endif()
 
 add_subdirectory(doc)
+
+if(BIICODE)
+    ADD_BIICODE_TARGETS()
+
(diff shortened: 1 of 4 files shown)


ericniebler:
You seem to be suggesting that I need to do something else, is that right? What exactly?

Manu343726:
In order to have the biicode block up to date, you might want to add deployment commands to the travis build, exactly as I shown you on Meta. See the short biicode description I left you on the Meta PR. Note the Meta deployment is just an example, and is not actually working since you have my password encrypted from my repo. You can use my block if you want, just create your own biicode account and I will add you as a block collaborator (Here the process is very similar to GitHub's one). Then encrypt your password (I.e. execute command 'travis encrypt BIICODE_PASSWORD="your password"') and substitute the current secure: entry just bellow BIICODE_USER with that one. And of course change BIICODE_USER value to your biicode account.

PR #117, opened 2015-03-23
Add missing headers to algorithm.hpp.

cartec:
algorithm.hpp should include is_sorted, is_sorted_until, and minmax_element.


ericniebler:
Thanks.

PR #127, opened 2015-03-31
add explicit defaulted copy constructor to proxy

gnzlbg:
No behavior is affected by this patch, this pr just makes the currently defaulted implicitly-declared copy constructor explicitly-declared instead. 

AFAIK `basic_iterator` will be rewritten "soon" anyways.


ericniebler:
Why is this needed?

gnzlbg:
Because the implicit generation of the default copy constructor is deprecated (in C++ 14 and above).

ericniebler:
Ah, thanks.

PR #129, opened 2015-03-31
allow selectively disabling deprecated warnings

gnzlbg:
The macro `RANGES_DISABLE_DEPRECATED_WARNINGS` allows disabling the deprecated warnings by redefining `RANGES_DEPRECATED(MSG)` to do nothing when this is defined.

This allows to disable the deprecation warnings selectively:

```c++
#include <ranges/v3/all.hpp>
#define RANGES_DISABLE_DEPRECATED_WARNINGS
#include <ranges/v3/some_deprecated_functionality_not_included_by_all.hpp>
#undef RANGES_DISABLE_DEPRECATED_WARNINGS
```

### DIFF
--- a/include/range/v3/detail/config.hpp
+++ b/include/range/v3/detail/config.hpp
@@ -56,16 +56,4 @@
 #endif
 #endif
 
-#if __cplusplus > 201103
-#define RANGES_DEPRECATED(MSG) [[deprecated(MSG)]]
-#else
-#if defined(__clang__) || defined(__GNUC__)
-#define RANGES_DEPRECATED(MSG) __attribute__((deprecated(MSG)))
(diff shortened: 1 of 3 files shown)


gnzlbg:
This only works the first time you include a given file, the:

`#include <ranges/v3/detail/deprecated.hpp>` 

in the filter implementation is under the filter include guards.

ericniebler:
It's good to give users a way to turn off deprecation warnings, but not selectively. Let's leave the definition of `RANGES_DEPRECATED` in config.hpp. Yes, I know you want to un-deprecate `view::filter`. That's a separate issue.

gnzlbg:
Ok, this is closed by pull-request #130 .

Issue #131, opened 2015-04-02
Some tests do not compile under clang/libc++ in C++14


gnzlbg:
The following tests do not compile with clang-trunk/libc++-trunk in C++14 mode:

- `action/push_front.cpp`
- `action/push_back.cpp`
- `view/transform.cpp`

The errors for `push_front`/`push_back` are of the form: 

```shell
In file included from range-v3/test/action/push_front.cpp:16:
range-v3/include/range/v3/action/push_front.hpp:91:21: error: static_assert failed "The object to be inserted with action::push_front must either be convertible
      to the range's value type, or else it must be a range of elements that are convertible to the range's value type."
                    CONCEPT_ASSERT_MSG(meta::or_<
                    ^                  ~~~~~~~~~~
range-v3/include/range/v3/utility/concepts.hpp:744:28: note: expanded from macro 'CONCEPT_ASSERT_MSG'
#define CONCEPT_ASSERT_MSG static_assert
                           ^
include/libcxx/__functional_base:415:12: note: in instantiation of function template specialization
      'ranges::v3::adl_push_front_detail::push_front_fn::operator()<ranges::v3::reference_wrapper<std::__1::list<int, std::__1::allocator<int> >, false>, const
      ranges::v3::for_each_view<ranges::v3::take_view<ranges::v3::view::ints_fn>, (lambda at range-v3/test/action/push_front.cpp:93:54)> &, 42, 0>' requested
      here
    return _VSTD::forward<_Fp>(__f)(_VSTD::forward<_Args>(__args)...);
           ^
include/libcxx/__config:381:15: note: expanded from macro '_VSTD'
#define _VSTD std::_LIBCPP_NAMESPACE
              ^
4 errors generated.
```

The errors for push_back are similar.

The problem seems to be that: `const
      ranges::v3::for_each_view<ranges::v3::take_view<ranges::v3::view::ints_fn>, (lambda at range-v3/test/action/push_back.cpp:68:54)> &` does not model `Iterable<T&&>`.

The errors for transform are of the form:

```shell
In file included from range-v3/test/view/transform.cpp:12:
In file included from include/libcxx/string:439:
In file included from include/libcxx/algorithm:628:
In file included from include/libcxx/memory:604:
In file included from include/libcxx/iterator:343:
include/libcxx/__functional_base:415:12: error: attempt to use a deleted function
    return _VSTD::forward<_Fp>(__f)(_VSTD::forward<_Args>(__args)...);
           ^
[...]

PR #134, opened 2015-04-07
Constexpr support

gnzlbg:
Adds C++14 constexpr support

- adds macros to detect the language version: 
    - `RANGES_CXX_GREATER_THAN_11` for C++>11 (e.g. C++14/17/...)
    - `RANGES_CXX_GREATER_THAN_14` for C++>14 (e.g. C++17/...)

- adds macro to enable relaxed constexpr support in C++>11:
   - `RANGES_RELAXED_CONSTEXPR` 
   - note in C++11 the macro is defined to `inline`

- add CMake option to choose with which language std the tests should be compiled:
   - `-DRANGES_CXX_STD=value` => `-std=c++value`,  `value={11,14,17,1y,1z}` (defaults to `11`).

- set up travis to test in both C++11 and C++14
- set up travis to test with:
  - gcc-4.9 (libstdc++)
  - clang-3.4, clang-3.5, clang-3.6, clang-3.7 (libc++)

- set up valgrind with CMake and travis 
  - note: all tests are run both with and without valgrind in all configurations

- add utilities: 
   - constexpr friendly array (used for testing)
   - constexpr reverse_iterator (used for rbegin/rend/...)

**Lessons learned**

- [x] `constexpr` functions are instantiated in unevaluated contexts (yes, really!)
    - Solution: don't use `-> decltype(...)` for SFINAE. Seriously, just don't, use `enable_if` or similar instead. 
    - Otherwise... different branches will be taken at compile time in C++14 is some of the functions involved are marked `constexpr`. This is particularly bad for customization points since some users might mark theirs `constexpr` and others won't. The alternatives are more verbose but work fine and are in general more explicit (concepts, enable_if, ...). 
    - In a nutshell: your code shall not break when removing the trailing return types (`-> decltype(...)`) in C++14.

**Currently work-in-progress ("// TODO: [constexpr]")**:

- [ ] Set up travis to test range-v3 under MacOS using AppleClang + libc++.
- [ ] -Wundefined-inline: `constexpr` implies `inline` which introduces some visibility warnings.
- [x] Rethink general workarounds:
    - [x] `std::bind` is not `constexpr` (a one parameter parameter "binder" `binder_` is implemented in functional)
    - [x] `constexpr` functions are instantiated in unevaluated contexts (out of luck here)
- [x] Open issues (haven't given them enough thought):
    - [x] Should only non-mutable algorithms/views be implemented? 
        - No, mutable algorithms/views can also be made constexpr and used in a "non-constexpr"y way inside `constexpr` functions.
        - Old note: mutating `begin` and `end` member functions cannot mutate the algorithm state if this is `constexpr`. Update: but this is not `constexpr` if not marked as such inside a `constexpr` function. 
    - [x] Should actions be implemented?
        - Yes, same rationale as for mutable algorithms/views.
- [ ] Clean-up: fix all todos, rebase, remove all whitespace changes, ...
  - [ ] remove advance_to_fn from range_fwd (is superseded by advance)
  - [ ] check license in new files
  - [ ] check for uses of `std::ref` instead of `ranges::ref`

Make most of range-v3 constexpr:

- [ ] core:
  - [x] `begin`/`end`/`rbegin`/`rend`...
  - [x] `at`/`size`/`empty`/`front`/`back`
  - [x] `move` + `indirect_move`
  - [x] `swap` (doesn't add std::swap to ADL anymore because it is not constexpr)
  - [ ] `distance`
     - rethink distance as customization point
  - [ ] note: add issue `data` non-member function 
  - [ ] `range_adaptor`: discuss with eric the mutable member:  `/*mutable*/ base_range_t rng_;` 
- [x] utility:
  - [x] `variant` impl is not constexpr and works (at least for the tests)
    - Note: feels like duck-tape, I think it is better to reconsider the design of `tagged_variant` from scratch. For example Eggs.Variant is a 1:1 constexpr-compliant replacement to `tagged_variant`, with:

       ```c++
           template<typename... Ts>
           using tagged_variant = eggs::variant<Ts...>;
           using eggs::variants::get;
           template<std::size_t I, typename...Ts, typename U>
           RANGES_RELAXED_CONSTEXPR
           void set(tagged_variant<Ts...> &var, U &&u)
           {
               var = std::forward<U>(u);
           }
           /// + replace all calls to is_valid() with nothing (it has an explicit operator bool that does that)
       ```
       I've added an `explicit operator bool() -> is_valid()` to ease the transition if desired, but it is currently not used.
    - note: if all types are trivially destructible, then the destructor is defaulted and variant is constexpr
  - [x] `optional` depends on `variant`
  - [x] `semiregular` depends on `variant` and `optional`
  - [x] `common_tuple` / `common_pair`
     - Note these should be reconsidered with `constexpr` in mind
- [ ] algorithms:
   - [x] `all_of`/`any_of`/`none_of`
   - [x] `adjacent_find`
   - [x] `binary_search`
   - [x] `copy` / `copy_if` / `copy_n` / `copy_backward`
   - [x] `count`/`count_if`
   - [x] `equal` (implementation + tests)
   - [x] `fill`/`fill_n`
   - [x] `find`/`find_if`/ `find_if_not`
   - [x] `find_end`
   - [x] `find_first_of`
   - [x] `for_each`
   - [x] `generate` / `generate_n`
   - [x] `heap_algorithm`(`is_heap_until`/`is_heap`/`push_heap` / `pop_heap` / `make_heap`/`sort_heap`)
   - [x] `inplace_merge` (unbuffered)
        - [ ] Add issue about exposing the non-allocating `inplace_merge` unbuffered in the API
   - [x] `is_partitioned`
   - [x] `is_sorted`/`is_sorted_until`
   - [x] `lexicographical_compare`
   - [x] `lower_bound`/`upper_bound`/`equal_range`
   - [x] `merge` / `merge_move`
   - [x] `min_element`/`max_element`/`minmax_element`
   - [x] `move` / `move_backward`
   - [x] `mismatch`
   - [ ] `nth_element`
   - [ ] `partial_sort` / `partial_sort_copy`
   - [ ] `partition` / `partition_copy` / `partition_move` 
   - [ ] `partition_point`
   - [ ] `permutation`
   - [ ] `random shuffle`
   - [ ] `remove` / `remove_copy` / `remove_copy_if`
   - [ ] `replace` / `replace_if` / `replace_copy` / `replace_copy_if` 
   - [ ] `reverse` / `reverse_copy`
   - [x] `rotate` / `rotate_copy`
   - [ ] `search` / `search_n`
   - [ ] `set_algorithm`
   - [ ] `shuffle`
   - [ ] `sort`
   - [ ] `stable_partition`
   - [ ] `stable_sort`
   - [ ] `swap_ranges`
   - [ ] `transform`
   - [ ] `unique` / `unique_copy`
- [ ] views:
   - [ ] `all`: 
   - [x] `iota` (implementation + tests)
   - [x] `generate`/`generate_n` (implementation + tests)
   - [x] `join` (implementation + tests)
   - [x] `repeat`/`repeat_n` (implementation + tests)
   - [x] `take` (implementation + tests)
   - [x] `take_exactly`
   - [x] `transform` (implementation + tests)
       - [ ] note: issue #131 is not fixed yet
   - [x] `filter`/ `remove_if`:
       - Old note: amortized mutating `begin` and `end` cannot mutate the range if it is ~~~a constant expression (i.e. const)~~~ constant, but they can mutate a non-constant range within a constant expression.
   - [x] `reverse` (implementation + tests)
   - [x] `zip` (implementation + tests)
- [ ] actions ?


ericniebler:
Closing due to no activity.

PR #136, opened 2015-04-17
Add missing include: empty.hpp

ncrookston:
Missing include in join.hpp:
```c++
#include <range/v3/view/join.hpp>

int main() {}
```
yields
```bash
crookston@crookston:~$ clang++-3.5 simple.cpp -I dev/range-v3/include/ -std=c++11
In file included from simple.cpp:1:
dev/range-v3/include/range/v3/view/join.hpp:282:30: error: no member named
      'empty' in namespace 'ranges'
                    (ranges::empty(this->mutable_base()) ? 0 :
                     ~~~~~~~~^
1 error generated.
```

Thanks for this library!


ericniebler:
Thank you.

PR #137, opened 2015-04-22
Initial C++14 support

gnzlbg:
- adds macros to detect the language version:
    - RANGES_CXX_GREATER_THAN_11 for C++>11 (e.g. C++14/17/...)
    - RANGES_CXX_GREATER_THAN_14 for C++>14 (e.g. C++17/...)

- adds macro to enable relaxed constexpr support in C++>11:
    - RANGES_CXX14_CONSTEXPR (in the spirit of BOOST_CXX14_CONSTEXPR)
    - Note: constexpr implies inline, so the macro is defined to
      inline in C++11 to retain the same visibility.

- add CMake option to choose with which language std the tests should be
  compiled:
    - DRANGES_CXX_STD=value, where "value" is the revision of the
      standard (value={11,14,17,1y,1z}) and defaults to 11. It is
      used to set the compiler-flag "-std=c++value".

- extend the travis test matrix to:
    - test in both C++11 and C++14
    - test multiple versions of clang starting at the minimum
      supported version: 3.4, 3.5, 3.6, trunk.
    - test with gcc 4.9
    - run all test of release builds under valgrind

- set algorithm tests:
    - further subdivide set algorithm tests (set_difference,
      set_intersection, set_symmetric_difference, and set_union)
      to avoid memory errors under travis.
    - fixed a bug in all the set algorithm tests where a
      bidirectional iterator was being tested twice instead of
      also testing forward iterators.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,59 +1,234 @@
+# Adapted from various sources, including:
+# - Louis Dionne's Hana: https://github.com/ldionne/hana
+# - Paul Fultz II's FIT: https://github.com/pfultz2/Fit
 language: cpp
-
 script: cmake
 
-compiler:
- - clang
- - gcc
-
-env:
-  global:
-    - GCC_VERSION="4.9"
-  matrix:
-    - BUILD_TYPE=Debug
-    - BUILD_TYPE=Release
-    - ASAN=On
-    
+# Test matrix:
+# - Build matrix per compiler: C++11/C++14 + Debug/Release
+# - Optionally: AddressSanitizer (ASAN)
+# - Valgrind: all release builds are also tested with valgrind
+# - clang 3.4, 3.5, 3.6, trunk
+#   - Note: 3.4 and trunk are tested with/without ASAN,
+#     the rest is only tested with ASAN=On.
+# - gcc 4.9, 5.0
+#
 matrix:
-  exclude:
-    - compiler: gcc
-      env: ASAN=On
+  include:
+    # Test clang-3.4: C++11, Buidd=Debug/Release, ASAN=On/Off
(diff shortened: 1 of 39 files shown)


gnzlbg:
I think it is better to start committing some useful parts of the constexpr branch as soon as they are solid. Smaller pull-request are easier to review. 

This is just the basic infrastructure (needed macros) and testing support.

I've explicitly left in clang-3.5 since it segfaults on meta and I don't know what is the right thing to do here. C++14 is only tested for clang > 3.4.

ericniebler:
I've been super-busy, but what is going on with the Travis build? It seems to take an age and a half.

gnzlbg:
I agree, it takes ~2h for the tests to run, thats a lot. 

Each job takes about ~30min and travis can run 4 of those simultaneously, so it can run ~8 jobs per hour. If we cut clang-3.5, there will be ~18 jobs, which is ~2-2.5 hours.

gnzlbg:
We could just tests clang-3.4, clang-trunk and gcc 4.9, and always test with ASan. That results in 8 jobs which should take 1h:

- clang-3.4 (libc++):
   - c++11-debug-asan
   - c++11-release-asan-valgrind
- clang-trunk (libc++): 
  - c++11-debug-asan
  - c++11-release-asan-valgrind
  - c++14-debug-asan
  - c++14-release-asan-valgrind
- gcc-4.9 (libstdc++):
  - c++11-debug
  - c++11-release-valgrind

gnzlbg:
It took ~1h to run the last commit but somehow travis reports the sum of all the jobs time which is 8x30min ~4h... I'll just rebase this, we can uncomment the other tests later if needed.

gnzlbg:
The job that failed in the PR build didn't fail on my branch build.

PR #139, opened 2015-04-28
enable relaxed constexpr in core

gnzlbg:
- adds constexpr to:
  - begin, end, rbegin, rend, cbegin, cend, crbegin, crend
    - note: i've hijacked ADL for C-arrays and initializer lists in ranges::rbegin/rend/crbegin/crend 
      because std::rbegin/rend/crbegin/crend are not constexpr (while std::begin/std::end are...). 
        - a constexpr reverse_iterator is added to utility to make this work.
  - advance, next, prev, distance, iter_size, swap, indirect_swap
    - note 1: i've hijacked std::advance for pointers since it is not constexpr
    - note 2: for the test iterators I have to define advance overloads that forward
      to range-v3 ones. I think it would be better to allow customizing advance's implementation
      via a different name (e.g. advance_customization_point), and not import std::advance at all. 
  - front, back, empty, size, at

- adds constexpr tests to test/constexpr_core.cpp
   - test iterators are constexpr
   - test::array is fully tested, C-arrays and initializer_lists are partially tested (since swap for them
     is not constexpr and iter_swap cannot work). view::iota tests can be added in the future.

- adds reverse_iterator to utilities
- adds testing utility array
- fix: remove c++1y test for view::join, it is already tested in C++14 mode.

### DIFF
--- a/include/range/v3/range_access.hpp
+++ b/include/range/v3/range_access.hpp
@@ -82,63 +82,63 @@ namespace ranges
             };
 
             template<typename Rng>
-            static auto begin_cursor(Rng & rng, long)
+            static RANGES_CXX14_CONSTEXPR auto begin_cursor(Rng & rng, long)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 rng.begin_cursor()
             )
             template<typename Rng>
-            static auto begin_cursor(Rng & rng, int)
+            static RANGES_CXX14_CONSTEXPR auto begin_cursor(Rng & rng, int)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 static_cast<Rng const &>(rng).begin_cursor()
             )
             template<typename Rng>
-            static auto end_cursor(Rng & rng, long)
+            static RANGES_CXX14_CONSTEXPR auto end_cursor(Rng & rng, long)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 rng.end_cursor()
             )
             template<typename Rng>
-            static auto end_cursor(Rng & rng, int)
+            static RANGES_CXX14_CONSTEXPR auto end_cursor(Rng & rng, int)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 static_cast<Rng const &>(rng).end_cursor()
             )
 
             template<typename Rng>
-            static auto begin_adaptor(Rng & rng, long)
+            static RANGES_CXX14_CONSTEXPR auto begin_adaptor(Rng & rng, long)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 rng.begin_adaptor()
             )
             template<typename Rng>
-            static auto begin_adaptor(Rng & rng, int)
+            static RANGES_CXX14_CONSTEXPR auto begin_adaptor(Rng & rng, int)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 static_cast<Rng const &>(rng).begin_adaptor()
             )
             template<typename Rng>
-            static auto end_adaptor(Rng & rng, long)
+            static RANGES_CXX14_CONSTEXPR auto end_adaptor(Rng & rng, long)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 rng.end_adaptor()
             )
             template<typename Rng>
-            static auto end_adaptor(Rng & rng, int)
+            static RANGES_CXX14_CONSTEXPR auto end_adaptor(Rng & rng, int)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 static_cast<Rng const &>(rng).end_adaptor()
             )
 
             template<typename Cur>
-            static auto current(Cur const &pos)
+            static RANGES_CXX14_CONSTEXPR auto current(Cur const &pos)
             RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
             (
                 pos.current()
             )
             template<typename Cur>
-            static auto next(Cur & pos)
+            static RANGES_CXX14_CONSTEXPR auto next(Cur & pos)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 pos.next()
@@ -150,7 +150,7 @@ namespace ranges
                 pos.done()
             )
             template<typename Cur>
-            static auto equal(Cur const &pos0, Cur const &pos1)
+            static RANGES_CXX14_CONSTEXPR auto equal(Cur const &pos0, Cur const &pos1)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 pos0.equal(pos1)
@@ -162,18 +162,19 @@ namespace ranges
                 end.equal(pos)
             )
             template<typename Cur>
-            static auto prev(Cur & pos)
+            static RANGES_CXX14_CONSTEXPR auto prev(Cur & pos)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 pos.prev()
             )
             template<typename Cur, typename D>
-            static auto advance(Cur & pos, D n)
+            static RANGES_CXX14_CONSTEXPR auto advance(Cur & pos, D n)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 pos.advance(n)
             )
             template<typename Cur>
+            RANGES_CXX14_CONSTEXPR
             static auto distance_to(Cur const &pos0, Cur const &pos1)
             RANGES_DECLTYPE_AUTO_RETURN
             (
@@ -242,12 +243,12 @@ namespace ranges
             using single_pass_t = typename single_pass<Cur>::type;
 
             template<typename Cur, typename S>
-            static Cur cursor(basic_iterator<Cur, S> it)
+            static RANGES_CXX14_CONSTEXPR Cur cursor(basic_iterator<Cur, S> it)
             {
                 return std::move(it.pos());
             }
             template<typename S>
-            static S sentinel(basic_sentinel<S> s)
+            static RANGES_CXX14_CONSTEXPR S sentinel(basic_sentinel<S> s)
             {
                 return std::move(s.end());
             }
--- a/include/range/v3/utility/basic_iterator.hpp
+++ b/include/range/v3/utility/basic_iterator.hpp
@@ -70,7 +70,7 @@ namespace ranges
             {
                 using type = Val;
                 template<typename I>
-                static type apply(I const & i)
+                static RANGES_CXX14_CONSTEXPR type apply(I const & i)
                 {
                     return *i;
                 }
@@ -89,14 +89,17 @@ namespace ranges
                             iterator_value_t<I>,
                             private_>;
                     I const it_;
+                    RANGES_CXX14_CONSTEXPR
                     explicit proxy(I i)
                       : it_(std::move(i))
                     {}
                     friend struct operator_brackets_const_proxy;
                 public:
+                    RANGES_CXX14_CONSTEXPR
                     proxy(proxy const&) = default;
+                    RANGES_CXX14_CONSTEXPR
                     proxy const & operator=(proxy &) const = delete;
-                    operator Ref() const
+                    RANGES_CXX14_CONSTEXPR operator Ref() const
                     {
                         return *it_;
                     }
@@ -105,12 +108,12 @@ namespace ranges
                         return *it_;
                     }
                     CONCEPT_REQUIRES(Convertible<Ref, value_t>())
-                    operator value_t() const
+                    RANGES_CXX14_CONSTEXPR operator value_t() const
                     {
                         return *it_;
                     }
                 };
-                static type apply(I i)
+                static RANGES_CXX14_CONSTEXPR type apply(I i)
                 {
                     return type{std::move(i)};
                 }
@@ -129,12 +132,13 @@ namespace ranges
                             iterator_value_t<I>,
                             private_>;
                     I const it_;
+                    RANGES_CXX14_CONSTEXPR
                     explicit proxy(I i)
                       : it_(std::move(i))
                     {}
                     friend struct operator_brackets_proxy;
                 public:
-                    operator Ref() const
+                    RANGES_CXX14_CONSTEXPR operator Ref() const
                     {
                         return *it_;
                     }
@@ -143,25 +147,29 @@ namespace ranges
                         return *it_;
                     }
                     CONCEPT_REQUIRES(Convertible<Ref, value_t>())
-                    operator value_t() const
+                    RANGES_CXX14_CONSTEXPR operator value_t() const
                     {
                         return *it_;
                     }
+                    RANGES_CXX14_CONSTEXPR
                     proxy (proxy const&) = default;
+                    RANGES_CXX14_CONSTEXPR
                     proxy const & operator=(proxy&) const = delete;
                     // BUGBUG assign from common reference? from rvalue reference?
+                    RANGES_CXX14_CONSTEXPR
                     proxy const & operator=(iterator_value_t<I> const & x) const
                     {
                         *it_ = x;
                         return *this;
                     }
+                    RANGES_CXX14_CONSTEXPR
                     proxy const & operator=(iterator_value_t<I> && x) const
                     {
                         *it_ = std::move(x);
                         return *this;
                     }
                 };
-                static type apply(I i)
+                static RANGES_CXX14_CONSTEXPR type apply(I i)
                 {
                     return type{std::move(i)};
                 }
@@ -190,7 +198,9 @@ namespace ranges
             private:
                 mutable value_type value_;
             public:
+                RANGES_CXX14_CONSTEXPR
                 postfix_increment_proxy() = default;
+                RANGES_CXX14_CONSTEXPR
                 explicit postfix_increment_proxy(I const& x)
                   : value_(*x)
                 {}
@@ -198,7 +208,7 @@ namespace ranges
                 // (*r++).mutate(), but it imposes fewer assumptions about the
                 // behavior of the value_type.  In particular, recall that
                 // (*r).mutate() is legal if operator* returns by value.
-                value_type& operator*() const
+                RANGES_CXX14_CONSTEXPR value_type& operator*() const
                 {
                     return value_;
                 }
@@ -216,7 +226,9 @@ namespace ranges
                 mutable value_type value_;
                 I it_;
             public:
+                RANGES_CXX14_CONSTEXPR
                 writable_postfix_increment_proxy() = default;
+                RANGES_CXX14_CONSTEXPR
                 explicit writable_postfix_increment_proxy(I x)
                   : value_(*x)
                   , it_(std::move(x))
@@ -225,16 +237,19 @@ namespace ranges
                 // value_type(*r++) can work.  In this case, *r is the same as
                 // *r++, and the conversion operator below is used to ensure
                 // readability.
+                RANGES_CXX14_CONSTEXPR
                 writable_postfix_increment_proxy const & operator*() const
                 {
                     return *this;
                 }
                 // So that iter_move(r++) moves the cached value out
+                RANGES_CXX14_CONSTEXPR
                 friend value_type && indirect_move(writable_postfix_increment_proxy const &ref)
                 {
                     return std::move(ref.value_);
                 }
                 // Provides readability of *r++
+                RANGES_CXX14_CONSTEXPR
                 operator value_type &() const
                 {
                     return value_;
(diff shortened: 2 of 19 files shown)


include/range/v3/utility/addressof.hpp
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            template <typename T>
+            typename std::enable_if<
+                addressof_detail::has_addressof_operator<T>::value, T*>::type
+            operator()(T& r) const noexcept
+            {
+                return std::addressof(r);
+            }
+        };
```
ericniebler: Ogh. This is exactly the sort of ugly hack expression SFINAE *should* solve. Is there no better way?

include/range/v3/utility/array.hpp
```diff
@@ -0,0 +1,304 @@
...
+template <size_t I, class T, size_t N>
+class tuple_element<I, ranges::array<T, N> >
+{
+ public:
+    using type = T;
+};
+
+}  // namespace std
+
+#endif // RANGES_V3_UTILITY_ARRAY_HPP
```
ericniebler: Oh, no. No no. If `constexpr` support means wholesale duplication of std functionality, I don't want it. This has to go.

include/range/v3/utility/reverse_iterator.hpp
```diff
@@ -0,0 +1,203 @@
...
+        template <typename It, CONCEPT_REQUIRES_(BidirectionalIterator<It>{})>
+        RANGES_CXX14_CONSTEXPR
+        reverse_iterator<It> make_reverse_iterator(It i)
+        {
+            return reverse_iterator<It>(i);
+        }
+
+    }  // namespace v3
+}  // namespace ranges
+#endif  // RANGES_V3_UTILITY_REVERSE_ITERATOR_HPP
```
ericniebler: Some things:

1) I *really* don't like duplicating std functionality. I've done in in isolated places (e.g. some std iterators weren't modeling the Iterator concepts of the Palo Alto report), but I did it grudgingly.
2) The iterator adaptors live in utility/iterator.hpp
3) Most of this verbiage can be replaced with an appropriately-defined cursor together with the `basic_iterator` utility.
4) Don't use `iterator_traits`. New-style iterators don't define nested `pointer` or `reference` type aliases or `operator->`.

include/range/v3/utility/swap.hpp
```diff
@@ -42,10 +53,26 @@ namespace ranges
...
+                std::is_move_assignable<T>::value
+             >::type
+            swap(T& x, T& y)
+            noexcept(std::is_nothrow_move_constructible<T>::value &&
+                     std::is_nothrow_move_assignable<T>::value)
+            {
+                T t(std::move(x));
+                x = std::move(y);
+                y = std::move(t);
+            }
```
ericniebler: `std::move` isn't `constexpr` in C++11.

ericniebler: This just doesn't work. It's perfectly acceptable for users to explicitly specialize `swap` on user-defined types in namespace `std`. Routing around `std::swap` just because it happens not to be `constexpr` is fundamentally broken.

ericniebler: ...And it seems to me that if you try to call `swap` on a type that happens to have `std` as an associated namespace, it's going to find the one in `std` anyway. This overload creates an ambiguity error, does it not?

include/range/v3/utility/reverse_iterator.hpp
ericniebler: Ah. If you remove `operator->` then you can nuke the weird `ranges::addressof`. Cool.

include/range/v3/utility/array.hpp
gnzlbg: I guess you mean the duplication of `std::array` in general (correct me if I'm wrong). 

The duplication is not necessary, but it simplifies testing so much that it is worth having. As stated in the pull request, I'm happy to move this to the test directory. The issue it solves is that besides C-arrays, and initializer lists, I need to test against a constexpr friendly user defined sequence. This array class is way simpler than `view::ints` to get right, and adds almost no dependencies. At the beginning I was using `view::ints` to test everything, but that pulls more of range-v3 into each test and requires a lot of `view`-related functionality to be constexpr. I didn't want to submit all those changes just yet. I've done it once, and I prefer to tackle this step by step, in the following order: core, algorithms, views, and finally actions.

So I guess I agree, it sucks that `std::array` is broken, otherwise I would use it. But this is better than the other alternatives I've tried, so I would like to keep it.

ericniebler: I would suggest a minimal constexpr-friendly `array` container in the test directory. Containers don't belong in range-v3 proper.

include/range/v3/utility/reverse_iterator.hpp
gnzlbg: > 1) I really don't like duplicating std functionality. I've done in in isolated places (e.g. some std iterators weren't modeling the Iterator concepts of the Palo Alto report), but I did it grudgingly.

I don't like it either.

> 2) The iterator adaptors live in utility/iterator.hpp
> 3) Most of this verbiage can be replaced with an appropriately-defined cursor together with the basic_iterator utility.

This is libc++'s `reverse_iterator` implementation. I'll write a reverse iterator using `basic_iterator` and a cursor and make sure it passes all libc++ tests. And put that into `utility/iterator.hpp`.

> 4) Don't use iterator_traits. New-style iterators don't define nested pointer or reference type aliases or operator->.
> Ah. If you remove operator-> then you can nuke the weird ranges::addressof. Cool.

Cool, didn't knew about this. I'll remove those then (`addressof` will come back for variant tho).

ericniebler: > I'll write a reverse iterator using basic_iterator and a cursor and make sure it passes all libc++ tests.

Obviously, if the libc++ tests check for the presence of `operator->` or whatever, those tests will fail and that's ok. Some simple sanity checks would do IMO.

ericniebler:
> maybe importing `rbegin`/`rend` should be removed, since i've hijacked them anyways.

A decision needs to be made about whether `rbegin`/`rend` are customization points. If they are, we can't remove the `using` declarations you refer to without causing potential ambiguity errors.

include/range/v3/utility/swap.hpp
gnzlbg: Thanks for taking a look at this. I didn't knew users could specialize `std::swap` in the `std` namespace. I thought the preferred way was to add a swap function to the namespace of the user-defined type and let ADL do the rest. If explicit specializations of `std::swap` are to be considered, then I need to reconsider this approach (more below).

First, what I want and tried to do is to let ADL pick the swap function in the namespace of the user-defined type. For literal types, or if no swap function is found, it should pick this swap function and not the one in namespace std. As far as I can tell, this overload does not create any ambiguity. It would if I were to import `std::swap`. 

If I have to allow for explicit specializations of std swap, then I need to import std::swap. I don't see a way to import those without importing `std::swap`. The whole thing is not doomed tho, since I can probably add a constexpr swap to the ranges namespace that will be picked for types living on it, and users can add constexpr swap functions for their own types if desired. However, I don't see a way to add a constexpr swap for C-arrays if I have to import the std::swap here (I've just realized that I forgot to add the swap overload for C-arrays that calls `swap_ranges` but I can fix this.). What do you think?
 
> std::move isn't constexpr in C++11.

The function is only constexpr in C++14 or greater.

gnzlbg:
> A decision needs to be made about whether rbegin/rend are customization points.

I'm inclined to think that they are customization points. Until now I always asked myself the question "Does it make sense to customize this function for C-arrays?". Maybe this is the wrong question, but for begin, end, size, rbegin, and rend the answer is yes. If it makes sense for C-arrays, it might make sense for some other user-defined type that one cannot easily modify. 

>If they are, we can't remove the using declarations you refer to without causing potential ambiguity errors.

Then they stay. I've only "hijacked" them for C-arrays and `std::initializer_lists`.

include/range/v3/utility/swap.hpp
gnzlbg: So I just learned that yes, it is common to explicitly specialize std::swap since you cannot use ADL for hooking up a swap for e.g. `std::reference_wrapper<T>`. I'll rethink this.

ericniebler: I think you're mistaken about the ambiguity error. I just tried the following with your branch:

~~~~cpp
#include <utility>
#include <complex>
#include <range/v3/core.hpp>

template<typename T>
struct S {
    T t;
};

int main()
{
    S<std::complex<float>> s,t;
    ranges::swap(s,t);
}
~~~~

With clang, this is the error I get:

~~~
/cygdrive/c/Users/eric/Code/range-v3/libs/range/v3/scratch/range.cpp:13:5: error: no matching function for call to object of type 'const ranges::v3::adl_swap_detail::swap_fn'
    ranges::swap(s,t);
    ^~~~~~~~~~~~
/cygdrive/c/Users/eric/Code/range-v3/include/range/v3/utility/swap.hpp:94:17: note: candidate template ignored: substitution failure [with T = S<std::complex<float> > &, U = S<std::complex<float> > &]: no type named 'type' in
      'meta::v1::detail::_if_<std::integral_constant<bool, false> >'
                operator()(T && t, U && u) const noexcept(is_nothrow_swappable<T, U>::value)
                ^
1 error generated.
~~~

With my branch, it compiles fine. Reread my comment above about types with `std` as an associated namespace and ADL. This is the major problem with your approach. I'm quite certain that any effort to redefine `swap` will fail this way.

include/range/v3/utility/array.hpp
gnzlbg: Done.

include/range/v3/utility/swap.hpp
gnzlbg: You are right. I've reverted the swap changes, and added this test to `test/utility/swap.cpp`.

After sleeping over it, I think the best thing is to just import std swap. Since ADL is used users can implement constexpr versions of swap for their own types if so desired. 

The only case that might be worth considering is that of C-arrays, but that can be done in the `swap_fn` object before performing ADL.

gnzlbg: I worked out a new implementation that defaults to ADL and falls back to std::swap. 

Users can decide if they want ADL + std::swap, ADL + ranges::swap, or just ranges::swap without ADL (this is useful for std library types).

include/range/v3/utility/reverse_iterator.hpp
gnzlbg: @ericniebler I've mostly done this in the last commit, and am pretty amazed at how easy it was to do it with cursor and basic iterator (even without docs). 

There are only two minor quirks missing. One is a warning during "concept checking" that occurs in the test constexpr_core.cpp. The other is a situation that used to compile with `std::reverse_iterator` but doesn't compile with the `basic_iterator` implementation. It is related to  `reverse_iterator<non_const_it> - reverse_iterator<const_it>`, where  `reverse_iterator<const_it>` has to be converted to `reverse_iterator<non_const_it>` when using `basic_iterator`, which fails.

include/range/v3/utility/swap.hpp
ericniebler: Please just put swap back the way it was.

gnzlbg: Done. 

I've had to remove the constexpr `iter_swap` and `swap` tests since `std::swap` will be picked up for, among others, fundamental types. The associated namespaces of fundamental types are empty, and there is not much that can be done about this. As a consequence, e.g. `iter_swap(int*, int*)` won't be constexpr. I don't know how large of a hit this will be, but I guess it will be a pretty large hit (see the commented out code in the `constexpr_core.cpp` tests). One can work around this for a user defined type (e.g. see `test/array.hpp`), but that workaround is just wrong (I just left it for exposition only).

include/range/v3/utility/addressof.hpp
ericniebler: So, you can just use `reinterpret_cast` here because that's not `constexpr`?

include/range/v3/utility/swap.hpp
```diff
@@ -23,6 +34,14 @@
...
 #include <range/v3/utility/static_const.hpp>
 
+RANGES_CXX14_CONSTEXPR
+void swap(int& a, int& b) {
+    int tmp = a;
+    a = b;
+    b = tmp;
+}
+
+
```
ericniebler: Remove this.

include/range/v3/utility/iterator.hpp
```diff
@@ -483,25 +526,119 @@ namespace ranges
...
+
+        namespace detail
+        {
+            template<typename I> struct reverse_cursor;
+        }
+
+        template<typename I>
+        using reverse_iterator = basic_iterator<detail::reverse_cursor<I>,
+                                                detail::reverse_cursor<I>>;
+
```
ericniebler: These belong in range_fwd.hpp

include/range/v3/utility/iterator.hpp
```diff
@@ -483,25 +526,119 @@ namespace ranges
...
+        namespace detail
+        {
+
+        template<typename I>
+        struct reverse_cursor
+        {
```
ericniebler: Formatting

include/range/v3/utility/swap.hpp
ericniebler: I understand. We can revisit this in the future.

include/range/v3/utility/addressof.hpp
gnzlbg: > So, you can just use reinterpret_cast here because that's not constexpr?

Yes, but then I would be reimplementing `std::addressof` which depending on the std library might be doing a bit more than that (e.g. libc++ interfaces with Objective-C++ pointers...).

include/range/v3/utility/swap.hpp
gnzlbg: Oops, fixed.

include/range/v3/utility/iterator.hpp
gnzlbg: Fixed.

include/range/v3/utility/iterator.hpp
gnzlbg: Fixed.

include/range/v3/utility/addressof.hpp
gnzlbg: It might be worth it to also tests here if the user-provided addressof operator is constexpr or not.

gnzlbg:
> Also, the fact that basic_iterator still provides operator-> (and hence needs std::addressof) is merely an oversight. It ought to be deleted.
> 4) Don't use iterator_traits. New-style iterators don't define nested pointer or reference type aliases or operator->.

I've removed `operator->`, `pointer`,  and also `pointer` from the specialization of `std::iterator_traits<basic_iterator<C, S>>`, but removing reference gives too many errors. Having a crippled `std::iterator_traits` is not something anybody will expect.

I fixed the conversion tests that were not passing for `reverse_iterator`. The fix is not finished tho since I still need to constrain the `basic_iterator::operator-(other_basic_iterator)` more. The issue was:

```c++
char* a;
const char * b;
// this worked:
b - a;
// since it expands to b.operator-(a)
// and a (char*) can be converted to b (const char*)
// this failed:
a - b; 
// since b (const char*) cannot be converted to a (char*)
```

I cannot constraint operator- with `Convertible` since the types are not convertible to each other, but they can be substracted from each other. I also don't know which difference_type to choose, probably the widest of both is the right solution.



t break

ericniebler:
>> Also, the fact that basic_iterator still provides operator-> (and hence needs std::addressof) is merely an oversight. It ought to be deleted.
>> 4) Don't use iterator_traits. New-style iterators don't define nested pointer or reference type aliases or operator->.
>
> I've removed operator->, pointer, and also pointer from the specialization of std::iterator_traits<basic_iterator<C, S>>, but removing reference gives too many errors. Having a crippled std::iterator_traits is not something anybody will expect.

Sorry, I meant the *iterators* don't need nested `pointer` or `reference` type aliases. `iterator_traits` still needs them, though.

> I fixed the conversion tests that were not passing for reverse_iterator. The fix is not finished tho since I still need to constrain the basic_iterator::operator-(other_basic_iterator) more. The issue was:
> 
> ~~~~
> char* a;
> const char * b;
> // this worked when using 
> // basic_iterator<random_access_iterator<char*>> and ...<<const char*>>:
> b - a;
> // since it expands to b.operator-(a)
> // and a (char*) can be converted to b (const char*)
> // this failed:
> a - b; 
> // since b (const char*) cannot be converted to a (char*)
> ~~~~
>
> I cannot constrain operator- with Convertible since the types are not convertible to each other, but they can be substracted from each other. I also don't know which difference_type to choose, probably the widest of both is the right solution.

I'll see what can be done about that.

include/range/v3/range_access.hpp
```diff
@@ -162,19 +162,19 @@ namespace ranges
...
-            static auto advance(Cur & pos, D n)
+            static RANGES_CXX14_CONSTEXPR auto advance(Cur & pos, D n)
             RANGES_DECLTYPE_AUTO_RETURN
             (
                 pos.advance(n)
             )
-            template<typename Cur>
-            static auto distance_to(Cur const &pos0, Cur const &pos1)
+            template<typename Cur0, typename Cur1>
+            static RANGES_CXX14_CONSTEXPR auto distance_to(Cur0 const &pos0, Cur1 const &pos1)
```
gnzlbg: I had to change this to allow for computing the distance between iterators of different types that are not convertible to each other but can still be "substracted" (`const char*`, `char`).

ericniebler: I fixed this problem in `basic_iterator`. Apologies for the merge conflicts. :-(

gnzlbg: Thanks! No big deal and everything works fine now :)

include/range/v3/utility/basic_iterator.hpp
```diff
@@ -411,15 +436,18 @@ namespace ranges
                     detail::cursor_concept_t<Cur>>;
 
             using detail::mixin_base<Cur>::get;
-            Cur &pos() noexcept
+
+          public:
+            RANGES_CXX14_CONSTEXPR Cur &pos() noexcept
```
ericniebler: `pos` should not be public.

include/range/v3/utility/basic_iterator.hpp
```diff
@@ -709,7 +754,6 @@ namespace std
                 ::ranges::detail::as_std_iterator_category<
                     typename iterator::iterator_category,
                     reference>>;
-        using pointer = typename iterator::pointer;
```
ericniebler: This line needs to be restored.

include/range/v3/utility/iterator.hpp
```diff
@@ -480,25 +523,111 @@ namespace ranges
...
+
+                RANGES_CXX14_CONSTEXPR reverse_cursor(I it) : it_(std::move(it)) {}
+                template<typename U, CONCEPT_REQUIRES_(Convertible<U, I>{})>
+                RANGES_CXX14_CONSTEXPR reverse_cursor(reverse_cursor<U> const& u)
+                        : it_(u.base()) {}
+
+                template<typename U, CONCEPT_REQUIRES_(Convertible<I, U>{})>
+                RANGES_CXX14_CONSTEXPR operator reverse_cursor<U>() {
+                    return U{it_};
+                }
```
ericniebler: You're liable to get conversion ambiguities if you have both a converting constructor and a conversion operator. Remove the conversion operator and keep the constructor.

include/range/v3/utility/iterator.hpp
```diff
@@ -480,25 +523,111 @@ namespace ranges
...
+                {
+                    it_ -= n;
+                }
+                template<class U, CONCEPT_REQUIRES_(RandomAccessIterator<I>())>
+                RANGES_CXX14_CONSTEXPR
+                iterator_difference_t<I>
+                distance_to(reverse_cursor<U> const &that) const
+                {
+                    return it_ - that.base();
+                }
```
ericniebler: This doesn't need to be a template. The machinery in `basic_iterator` makes sure the comparison is symmetric by forcing a conversion.

include/range/v3/utility/swap.hpp
```diff
@@ -10,6 +10,17 @@
...
+// (http://libcxx.llvm.org).
+//
+//===----------------------------------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
ericniebler: Remove this comment.

ericniebler:
After you've finished with those last few edits, can you rebase your pull request onto master and squash all your commits into one? [Instructions](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)

ericniebler:
Looks very good, thanks.

PR #141, opened 2015-05-01
fix take_while sentinel constraint

gnzlbg:
A const sentinel of `view::take_while` required a const predicate and a const range. 
Was that intended? (See new tests).

### DIFF
--- a/include/range/v3/view/take_while.hpp
+++ b/include/range/v3/view/take_while.hpp
@@ -61,7 +61,7 @@ namespace ranges
             {
                 return {pred_};
             }
-            CONCEPT_REQUIRES(Callable<Pred const, range_iterator_t<Rng const>>())
+            CONCEPT_REQUIRES(Callable<Pred const, range_iterator_t<Rng>>())
             sentinel_adaptor<true> end_adaptor() const
             {
                 return {pred_};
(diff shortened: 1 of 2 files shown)


ericniebler:
The take_while test is failing in debug mode now. Can you take a look?

gnzlbg:
>The take_while test is failing in debug mode now. Can you take a look?

Since `generate` returns an infinite range, `generate | take_while` is infinite, so one cannot take the distance of it. I think the fix is just removing the `CHECK`. 

One can always write `view::ints(1) | view::take_while([](int) { return true; })` but I think that using `take_while` to make an infinite range into a finite one is sometimes the intended behavior. So maybe an alternative could be to introduce a `view::I_promise_this_is_finite` or something.

ericniebler:
> One can always write `view::ints(1) | view::take_while([](int) { return true; })` but I think that using `take_while` to make an infinite range into a finite one is sometimes the intended behavior. So maybe an alternative could be to introduce a `view::I_promise_this_is_finite` or something.

It's a general problem: how to handle ranges that are *possibly* infinite (but probably not). A `view::finite` adaptor is not a bad suggestion. I'll think about it.

PR #143, opened 2015-05-05
constexpr algorithms

gnzlbg:
add constexpr support to the following core facilities:
  - distance / enumerate
  - range / sized_range / make_range / range::get
  - range_iterface
add constexpr support to the following facilities in utility/functional:
  - as function
  - with_braced_init_args
  - operators
  - reference_wrapper
  - ref_fn / ref
  - add greater as a synonym for operator> (requires WeaklyOrdered and EqualityComparable)
add c++14 constexpr support to utility/move
add c++14 constexpr support to safe_int and add constexpr swap function to safe_int
bugfix: set_difference test

The following algorithms can be used within constexpr functions with minimal to no changes to the 
algorithm implementation: 
- adjacent_find
- all_of
- any_of
- binary_search
- copy
- copy_backward
- copy_if
- copy_n
- count
- count_if
- equal
- equal_range
- fill
- fill_n
- find
- find_end
- find_first_of
- find_if
- find_if_not
- for_each
- generate
- generate_n
- includes
- inplace_merge (unbuffered)
- is_heap
- is_heap_until
- is_partitioned
- is_sorted
- is_sorted_until
- lexicographical_compare
- lower_bound
- make_heap
- max_element
- merge
- merge_move
- min_element
- minmax_element
- mismatch
- move
- move_backward
- none_of
- partition
- partition_copy
- partition_move
- partition_point
- push_heap
- random_shuffle (for user-provided constexpr random number generators)
- remove
- remove_copy
- remove_copy_if
- remove_if
- replace
- replace_copy
- replace_copy_if
- replace_if
- reverse_copy
- rotate
- rotate_copy
- search
- search_n
- set_difference
- set_intersection
- set_symmetric_difference
- set_union
- shuffle* (for user-provided constexpr uniform random number generators)
- sort_heap
- stable_partition
- stable_sort
- swap_ranges
- transform
- unique
- unique_copy
- upper_bound

[*] shuffle: requires a constexpr uniform_int_distribution that has been adapted from 
libc++ (with minimal changes), but no real changes to the algorithm itself. For testing purposes, the linear_congruential_engine from libc++ has been adapted (with minimal changes).

The following algorithms haven’t been adapted yet because goto is not allowed within constexpr functions and as a consequence the algorithms would have to be rewritten: 
- nth_element
- is_permutation

This is doable, and I partially did so for stable_partition, but I would like to know what you think.

The following algorithms are constexpr but use iter_swap. Since iter_swap requires indirect_swap and thus swap to be constexpr, they won’t be constexpr for most types (e.g. int, they work for `safe_int<int>` because it has a constexpr swap): 
- sort_heap
- pop_heap
- inplace_merge
- next_permutation
- partial_sort
- partial_sort_copy
- partition
- prev_permutation
- random_shuffle
- reverse
- shuffle
- stable_partition
- swap_ranges

I’ve tested these with safe_int by adding a friend swap function to safe_int. It might be better to add this functions for testing only, and not in general to safe int, since it overrides the default swap for the underlying type and does not perform ADL. 

On the usage of temporary buffers (note this discussion is now taking place in [this issue](https://github.com/ericniebler/range-v3/issues/168)):

I think it is :
- wrong for algorithms to allocate memory by default,
- even wronger that users cannot opt out of this behavior,
- bad that users cannot pass these algorithms preallocated buffers.

In libstdc++ and libc++ get_temporary_buffer just calls new in a loop. 
So even tho I can understand the good intentions behind get_temporary_buffer, I think that the way this is handled by the STL right now is far from optimal. 

The algorithms that use it directly are:
- stable_partition
- inplace_merge

And the algorithms that use these algorithms are:
- stable_sort

I’ve provided these algorithms with a constexpr member function called `inplace` that uses no buffer.
It will be better to provide them with another member function called `with_buffer` that allows the user to specify a buffer (such a stack allocated buffer). 

I'd rather leave this as is (with inplace) and open an issue where we can discuss this on its own, since I think is one of the things that is worth to get right on an STL2 (and maybe try to unify get_temporary_buffer with Allocator).

### DIFF
--- a/include/range/v3/utility/move.hpp
+++ b/include/range/v3/utility/move.hpp
@@ -32,6 +32,7 @@ namespace ranges
             {
                 template<typename T,
                     typename U = meta::eval<std::remove_reference<T>>>
+                constexpr
                 U && operator()(T && t) const noexcept
                 {
                     return static_cast<U &&>(t);
@@ -48,6 +49,7 @@ namespace ranges
             /// \ingroup group-utility
             /// \sa `move_fn`
             template<typename T>
+            RANGES_CXX14_CONSTEXPR
             meta::eval<std::remove_reference<T>> && operator|(T && t, move_fn move) noexcept
             {
                 return move(t);
@@ -66,10 +68,37 @@ namespace ranges
         /// \cond
         namespace adl_move_detail
         {
+
+            constexpr struct valid_expr_t
+            {
+                template<typename ...T>
+                void operator()(T &&...) const;
+            }  valid_expr {};
+
+            struct IndirectlyMovable_
+            {
+                template<typename I,
+                         typename R = decltype(*std::declval<I>()),
+                         typename U = meta::eval<std::remove_reference<R>>//,
+                >
+                auto requires_(I const& i) -> decltype(
+                    valid_expr(
+                        aux::move_t<R>{aux::move(*i)}
+                    ));
+            };
+            template<typename I>
+            using IndirectlyMovable
+            = concepts::models<IndirectlyMovable_, I>;
+
             // Default indirect_move overload.
             template<typename I,
-                typename R = decltype(*std::declval<I>()),
-                typename U = meta::eval<std::remove_reference<R>>>
+                  typename R = decltype(*std::declval<I>()),
+                  typename U = meta::eval<std::remove_reference<R>>,
+                  typename std::enable_if<IndirectlyMovable<I>{}, int>::type = 0        //,
+            //          typename E
+            // = decltype(aux::move_t<R>{aux::move(*std::declval<I const&>())})
+            >
+            RANGES_CXX14_CONSTEXPR
             aux::move_t<R> indirect_move(I const &i)
                 noexcept(std::is_reference<R>::value ||
                     std::is_nothrow_constructible<detail::decay_t<U>, U &&>::value)
@@ -80,6 +109,7 @@ namespace ranges
             struct indirect_move_fn
             {
                 template<typename I>
+                RANGES_CXX14_CONSTEXPR
                 auto operator()(I const &i) const
                     noexcept(noexcept(indirect_move(i))) ->
                     decltype(indirect_move(i))
--- a/include/range/v3/algorithm/adjacent_find.hpp
+++ b/include/range/v3/algorithm/adjacent_find.hpp
@@ -39,6 +39,7 @@ namespace ranges
                 CONCEPT_REQUIRES_(ForwardIterator<I>() && IteratorRange<I, S>() &&
                     IndirectCallableRelation<C, Project<I, P>>())>
             I
+            RANGES_CXX14_CONSTEXPR
             operator()(I begin, S end, C pred_ = C{}, P proj_ = P{}) const
             {
                 auto &&pred = as_function(pred_);
@@ -57,6 +58,7 @@ namespace ranges
                 typename I = range_iterator_t<Rng>,
                 CONCEPT_REQUIRES_(ForwardIterable<Rng>() &&
                     IndirectCallableRelation<C, Project<I, P>>())>
+            RANGES_CXX14_CONSTEXPR
             range_safe_iterator_t<Rng>
             operator()(Rng &&rng, C pred = C{}, P proj = P{}) const
             {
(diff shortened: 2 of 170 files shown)


gnzlbg:
Some comments:
- I still have to clean this, I will do so during this afternoon hopefully.
- Right now there are some `Wundefined-internal` warnings due to constexpr implying inline (e.g. in functional indirected), I'll take care of these.
- I've added an inplace member function that avoids the temporary buffer in some algorithms, I can call this function anything else (inplace_merge.inplace feels like "really really please do it in place"). Maybe "without_buffer" or "unbuffered".
- I've avoided the issue of std::swap not being constexpr, maybe it is worth it to open this as a separate issue where alternatives can be discussed. I don't think we will find a good alternative, but I have thought of some less bad alternatives (e.g. an extension header that one has to explicitly include)
- I've made swap a friend of safe_int for _my_ convenience but I still need to clean this. Is it ok if I just move it into a test utility within the test directory? 
- There are some functional utilities that could improve the constexpr experience in general but they are large features (constexpr bind, constexpr mem_fn). They are not worth pursuing for me, but might be useful for e.g. using projections from within constexpr functions. Sprout provides constexpr implementations of these based on libc++ for anyone interested.
- I've added an inplace member function that avoids the temporary buffer in some algorithms (other names could be unbuffered, or without_buffer). As an alternative approach I can think of either tag dispatching or having functions of different names, but as stated in the pull-request I think we need a better solution for algorithms that might improve their complexity guarantees if extra memory is available (allow users to provide a buffer/require no buffer, a default, unifying this with the Allocator concepts...). Shall I leave these out of this pull-request?

gnzlbg:
It seems that travis broke on master when sort_patterns was merged, I fixed it in the last rebase here, so hopefully all will be green in this PR soon.

ericniebler:
Just FYI: I'm at the committee meeting this week and C++Now next. I will have no time to look at this in the next 2 weeks.

Also, if you know what travis is complaining about, can you submit a separate PR? If it's small and simple, I'll merge that ASAP.

gnzlbg:
Cool! Have a lot of fun!

gnzlbg:
In case somebody mentions it, there is a paper about allowing goto in constexpr functions that lacked a very good motivation. In this PR I had to go through a couple of algorithms that do use goto. I liked how is_permutation and nth_element use it. From looking at their implementation, I cannot really find a good reason for banning goto from constexpr functions. 

The other algorithms that use goto are stable_partition and stable_sort.

include/range/v3/utility/move.hpp
```diff
@@ -66,10 +68,37 @@ namespace ranges
...
+            using IndirectlyMovable
+            = concepts::models<IndirectlyMovable_, I>;
+
             // Default indirect_move overload.
             template<typename I,
-                typename R = decltype(*std::declval<I>()),
-                typename U = meta::eval<std::remove_reference<R>>>
+                  typename R = decltype(*std::declval<I>()),
+                  typename U = meta::eval<std::remove_reference<R>>,
+                  typename std::enable_if<IndirectlyMovable<I>{}, int>::type = 0        //,
```
gnzlbg: The only truly broken thing that I still have to fix is figuring out why GCC builds break if I constrain `indirect_move` in any way I've tried:

- the current approach with enable if (using IndirectlyMovable, I know I shouldn't do this)
- constraining directly by just adding a new type to the template parameter list that does `decltype(aux::move_t<R>(aux::move(*std::declval<I const&>())))`,
- constraining using expression SFINAE... (this was my preferred way since it is the simplest thing possible)

They all work with all versions of clang, and they all break with GCC. I am completely bluffed. I'll install gcc 4.9 and look into this again this week since trying to debug this through travis takes too much time.

There are some other things to discuss like the `inplace` method on algorithms that use `get_temporary_buffer` and if `nth_element` and `is_permutation` should be rewritten to make them constexpr but once the direction in these issues is clear implementing them should be easy.

ericniebler:
Closing due to no activity.

gnzlbg:
Yes sorry about that. This is hitting an internal compiler error in gcc 5.0... and haven't had more time to look into it.

gnzlbg:
FYI I tried to prepare a small PR where I enabled gcc 5 under travis but while gcc 4.9 succeeds to compile range-v3 in c++11 mode, at least the gcc 5 used by travis fails with ICEs:

 https://travis-ci.org/gnzlbg/range-v3/builds/66921286

ericniebler:
Did you submit the bug report? Only one way to get bugs fixed, unless you
want to do it yourself. ;-)

gnzlbg:
I know, not yet, too bad an excuse for lack of time.

PR #144, opened 2015-05-05
disable sort_patterns in cxx11

gnzlbg:
problem: perf/sort_patterns requires c++14 and breaks in travis under C++11.
solution: disable sort_patterns in C++11 

I'll clean sort_patterns and backport it to C++11 later on.


gnzlbg:
I think this fixes the issue, but lets wait first what travis says.

PR #145, opened 2015-05-11
fix: variant apply visitor under c++14

gnzlbg:
SFINAE's apply_visitor since under C++14 consexpr it must be "turtles all the way down" with decltype.

### DIFF
--- a/include/range/v3/utility/variant.hpp
+++ b/include/range/v3/utility/variant.hpp
@@ -251,12 +251,13 @@ namespace ranges
                 apply_visitor(Fun &&fun, Var &var)
                   : var_(var), fun_(std::forward<Fun>(fun))
                 {}
-                template<typename T, std::size_t N>
-                void operator()(T &&t, meta::size_t<N> u) const
-                {
-                    using result_t = result_of_t<Fun(T &&, meta::size_t<N>)>;
-                    this->apply_(std::forward<T>(t), u, std::is_void<result_t>{});
-                }
+                template<typename T, std::size_t N,
+                         typename result_t = result_of_t<Fun(T &&, meta::size_t<N>)>>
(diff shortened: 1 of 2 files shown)


gnzlbg:
I had this in my constexpr branch but missed it since I didn't include the variant tests (making variant fully constexpr is a PR on its own).

include/range/v3/utility/variant.hpp
```diff
@@ -251,12 +251,13 @@ namespace ranges
...
-                {
-                    using result_t = result_of_t<Fun(T &&, meta::size_t<N>)>;
-                    this->apply_(std::forward<T>(t), u, std::is_void<result_t>{});
-                }
+                template<typename T, std::size_t N,
+                         typename result_t = result_of_t<Fun(T &&, meta::size_t<N>)>>
+                auto operator()(T &&t, meta::size_t<N> u) const
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    this->apply_(std::forward<T>(t), u, std::is_void<result_t>{})
```
ericniebler: I seem to recall that gcc has a hard time with `this` appearing in contexts like these. Apparently not since the tests passed. ???

gnzlbg: Hm I do not recall anything in particular about gcc and this but in the
constexpr algorithms branch the only tests that fail are gcc because it
somehow doesn't like the constrained implementation of move :/
I really don't know what is wrong there.

gnzlbg: I've just double checked that this function is actually being tested, and indeed it is exercised in the following tests:
	144 - test.view.concat
	161 - test.view.remove_if
	174 - test.example.fibonacci

PR #149, opened 2015-05-24
in action::unique, change sort_fn parameter in bind() to unique_fn

rhalbersma:


ericniebler:
Oops, thanks.

rhalbersma:
Note that this apparently was not caught by the test suite. I discovered this by accident while trying to create my own action for random_shuffle and copying the implementation from unique. I am not familiar enough with your code to suggest adding test cases for the bind() functions inside actions.

PR #150, opened 2015-05-24
Reserve capacity before inserting a SizedRange into a std::vector.

CaseyCarter:
It's a special case and not a generic optimization, but `std::vector` is worth special-casing. (In response to [this SO question](http://stackoverflow.com/questions/30425417/how-to-write-a-range-v3-action-for-random-shuffle).)

### DIFF
--- a/include/range/v3/action/concepts.hpp
+++ b/include/range/v3/action/concepts.hpp
@@ -90,6 +90,13 @@ namespace ranges
                 );
             };
 
+            struct ReserveAndAssignable : refines<Reservable(_1)> {
+                template <class C, class I>
+                auto requires_(C&& c, I&& i) -> decltype(
+                    concepts::valid_expr((c.assign(i, i), 42))
+                );
+            };
+
             struct RandomAccessReservable : refines<Reservable> {
                 template <class C>
                 auto requires_(C&& c) -> decltype(
@@ -100,6 +107,9 @@ namespace ranges
 
         template <class C>
         using Reservable = concepts::models<concepts::Reservable, C>;
+
+        template <class C, class I>
+        using ReserveAndAssignable = concepts::models<concepts::ReserveAndAssignable, C, I>;
 
         template <class C>
         using RandomAccessReservable = concepts::models<concepts::RandomAccessReservable, C>;
--- a/include/range/v3/action/concepts.hpp
+++ b/include/range/v3/action/concepts.hpp
@@ -77,6 +77,33 @@ namespace ranges
         template<typename T>
         using Container = concepts::models<concepts::Container, T>;
 
+        namespace concepts
+        {
+            struct Reservable : refines<Container> {
+                template <class T>
+                using size_type = decltype(std::declval<const T&>().size());
+
+                template <class C>
+                auto requires_(C&& c) -> decltype(
+                    concepts::model_of<Integral, size_type<C>>(),
+                    concepts::valid_expr((c.reserve(c.size()), 42))
+                );
+            };
+
+            struct RandomAccessReservable : refines<Reservable> {
+                template <class C>
+                auto requires_(C&& c) -> decltype(
+                    concepts::model_of<RandomAccessIterator, decltype(begin(c))>()
+                );
+            };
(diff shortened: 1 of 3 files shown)


CaseyCarter:
Yes, I thought of the allocator issue during "code review" while I was in the shower. Fixed.

CaseyCarter:
A general solution that requires size() and reserve() should work with `vector`, `basic_string`, `unordered_map`, and `unordered_set`.

ericniebler:
The concept wouldn't be hard to write. (ContiguousLayout isn't right. BulkInsertableContainer, maybe? Reservable?) I'd prefer if this were more general. `to_container` should also get the same treatment.

CaseyCarter:
I'm leaning toward Reservable as well:
```
template <class T>
concept bool Reservable =
  LvalueContainerLike<T> &&
  requires(C c) {
    { c.size() } -> Integral;
    c.reserve(c.size());
  };
template <class T>
concept bool RandomReservable =
  Reservable<T> &&
  RandomAcessIterator<IteratorType<T>>;
```
RandomReservable would be a useful refinement for cases like `insert` that want to reserve capacity but need to workaround iterator invalidation.

CaseyCarter:
Let me know if you want any design changes, and I'll submit another pull request with a cleaner commit log. ReserveAndAssignable is a terrible name.

include/range/v3/action/concepts.hpp
```diff
@@ -77,6 +77,43 @@ namespace ranges
...
+        namespace concepts
+        {
+            struct Reservable : refines<Container> {
+                template <class T>
+                using size_type = decltype(std::declval<const T&>().size());
+
+                template <class C>
+                auto requires_(C&& c) -> decltype(
+                    concepts::model_of<Integral, size_type<C>>(),
+                    concepts::valid_expr((c.reserve(c.size()), 42))
```
ericniebler: I prefer this is written as:

```.cpp
template<typename C>
auto requires_(C&& c) -> decltype(
    concept::valid_expr(
        concepts::model_of<Integral, size_type<C>>(),
        (c.reserve(c.size()), 42)
    ));
```

include/range/v3/action/concepts.hpp
```diff
@@ -77,6 +77,43 @@ namespace ranges
...
+                template <class C, class I>
+                auto requires_(C&& c, I&& i) -> decltype(
+                    concepts::valid_expr((c.assign(i, i), 42))
+                );
+            };
+
+            struct RandomAccessReservable : refines<Reservable> {
+                template <class C>
+                auto requires_(C&& c) -> decltype(
+                    concepts::model_of<RandomAccessIterator, decltype(begin(c))>()
```
ericniebler: Why not just `refines<Reservable, RandomAccessRange>` ?

include/range/v3/action/insert.hpp
```diff
@@ -78,22 +78,46 @@ namespace ranges
...
+                typename C = range_common_iterator_t<Rng>,
+                CONCEPT_REQUIRES_(RandomAccessReservable<Cont>() && SizedRange<Rng>())>
+            auto insert(Cont && cont, I p, Rng && rng) ->
+                decltype(unwrap_reference(cont).insert(begin(unwrap_reference(cont)), C{begin(rng)}, C{end(rng)}))
+            {
+                auto const index = p - begin(unwrap_reference(cont));
+                unwrap_reference(cont).reserve(unwrap_reference(cont).size() + size(rng));
+                return unwrap_reference(cont).insert(begin(unwrap_reference(cont)) + index, C{begin(rng)}, C{end(rng)});
+            }
+
```
ericniebler: I try to avoid negating concepts in requires clauses. Feels wrong stylistically. I also prefer keeping the count of public member functions down. This can be cleaned up by keeping the existing overloads and moving the implementation into some private members, dispatching to the correct one depending on the value of `RandomAccessReservable<Cont>() && SizedRange<Rng>()`.

CaseyCarter:
Closing and re-pulling.

PR #151, opened 2015-05-25
Add shuffle action.

FloopCZ:
In my opinion, since there is a sort action, there should also be a shuffle action. Requesting review.

### DIFF
--- a/include/range/v3/action/shuffle.hpp
+++ b/include/range/v3/action/shuffle.hpp
@@ -35,7 +35,8 @@ namespace ranges
             {
             private:
                 friend action_access;
-                template<typename Gen>
+                template<typename Gen,
+                    CONCEPT_REQUIRES_(UniformRandomNumberGenerator<Gen>())>
                 static auto bind(shuffle_fn shuffle, Gen && gen)
                 RANGES_DECLTYPE_AUTO_RETURN
                 (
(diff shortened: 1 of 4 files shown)


rhalbersma:
I wonder [where you got this idea](http://stackoverflow.com/questions/30425417/how-to-write-a-range-v3-action-for-random-shuffle#comment48946078_30426249) ;) 

BTW, if this PR gets accepted, you probably should also systematically replace random_shuffle from the various unit tests (dogfooding and all that...).

include/range/v3/action/shuffle.hpp
```diff
@@ -0,0 +1,108 @@
...
+        /// @{
+        namespace action
+        {
+            struct shuffle_fn
+            {
+            private:
+                friend action_access;
+                template<typename Gen>
+                static auto bind(shuffle_fn shuffle, Gen && gen)
+                RANGES_DECLTYPE_AUTO_RETURN
```
ericniebler: This `bind` function should be constrained with `UniformRandomNumberGenerator<Gen>`.

ericniebler:
That's the only nit. The rest looks good.

FloopCZ:
Thank you for your feedback, it really should be constrained. Fixed and commited.

PR #152, opened 2015-05-25
Add [[deprecated]] attribute to random_shuffle.

rhalbersma:
N4140: [depr.alg.random.shuffle] states:
"The function templates random_shuffle are deprecated."

Also add -Wno-error=deprecated-declarations to CMakeLists.txt.
This will override -Werror in case this would be added as a compiler option,
so that the random_shuffle unit tests will continue to compile in that case.


ericniebler:
Thanks

PR #153, opened 2015-05-25
Reservation support (Was #150)

CaseyCarter:
* Implement concepts Reservable, RandomAccessReservable, ReserveAndAssign to identify containers that can reserve capacity.
* Use RandomAccessReservable in insert.
* Use ReserveAndAssignable in to_container.

### DIFF
--- a/include/range/v3/action/insert.hpp
+++ b/include/range/v3/action/insert.hpp
@@ -76,23 +76,70 @@ namespace ranges
                 return unwrap_reference(cont).insert(p, n, std::forward<T>(t));
             }
 
+            namespace impl {
+                template<typename C, typename I, typename S>
+                using ReserveConcept =
+                    meta::fast_and<
+                        RandomAccessReservable<C>,
+                        SizedIteratorRange<I, S>>;
+
+                template<typename Cont, typename P, typename I, typename S,
+                    typename C = common_iterator<I, S>>
+                auto insert(Cont && cont, P p, I i, S j, std::false_type) ->
+                    decltype(unwrap_reference(cont).insert(p, C{i}, C{j}))
+                {
+                    return unwrap_reference(cont).insert(p, C{i}, C{j});
+                }
+
+                template<typename Cont, typename P, typename I, typename S,
+                    typename C = common_iterator<I, S>>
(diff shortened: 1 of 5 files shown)


include/range/v3/action/insert.hpp
```diff
@@ -76,23 +76,76 @@ namespace ranges
                 return unwrap_reference(cont).insert(p, n, std::forward<T>(t));
             }
 
+            namespace impl {
```
ericniebler: In range-v3, the detail namespace is called `detail` everywhere. Also, to avoid cluttering the Doxygen output with details, I wrap details with `/// \cond` and `/// \endcond`.

include/range/v3/action/insert.hpp
```diff
@@ -76,23 +76,76 @@ namespace ranges
...
+                using ReserveConcept =
+                    meta::fast_and<
+                        RandomAccessReservable<C>,
+                        SizedIteratorRange<I, S>>;
+
+                template<typename Cont, typename P, typename I, typename S,
+                    typename C = common_iterator<I, S>,
+                    CONCEPT_REQUIRES_(LvalueContainerLike<Cont>() && Iterator<P>() && IteratorRange<I, S>())>
+                auto insert(Cont && cont, P p, I i, S j, std::false_type) ->
+                    decltype(unwrap_reference(cont).insert(p, C{i}, C{j}))
```
ericniebler: `insert` is a customization point. If these aren't hooking the customization point, they should be called something else, like `insert_` or `insert_impl`.

include/range/v3/action/insert.hpp
ericniebler: Also, use [the Allman brace style](https://en.wikipedia.org/wiki/Indent_style#Allman_style) please.

ericniebler:
Other than those nits, this looks pretty good. Thanks for doing this.

include/range/v3/action/insert.hpp
CaseyCarter: This namespace is embedded inside the `adl_insert_detail` namespace, which is already wrapped in `\cond`/`\endcond`.

ericniebler:
Thanks!

PR #154, opened 2015-05-25
Warning fix in test/utility/basic_iterator.cpp

CaseyCarter:
d is intentionally set-but-not-used.

### DIFF
--- a/test/utility/basic_iterator.cpp
+++ b/test/utility/basic_iterator.cpp
@@ -62,6 +62,7 @@ int main()
     d = (a != b);
 
     detail::ignore_unused(
+        d,
         a < b,
         a <= b,
         a > b,


rhalbersma:
Ha! Talk about serendipity :) I just submitted a pull request of 5 commits that systematically documents what -Weverything catches, including this one that you just caught.

CaseyCarter:
You're far more disciplined than I - I just fixed the one particular warning that jumped at out me during unit testing ;)

rhalbersma:
I have an incentive because I compile all my own projects at -Weverything, so when I include from range-v3 I currently use `-isystem` on its include path to suppress all warnings. Although getting rid of every `-Wno-xxx` is almost impossible, it is very nice to know what subtleties are lurking beneath the surface.

PR #155, opened 2015-05-25
increase warning levels

rhalbersma:
On Clang, -Weverything will catch an enormous amount of warnings. This patch documents all the disabled warnings that lets the code compile warning free.

On gcc, -Wall -Wextra -pedantic is as high as it gets without individually adding
every possible warning. Frequent compilation with Clang will catch enough anyway.

There are some further minor differences in warning syntax between Clang / gcc.


test/constexpr_core.cpp
```diff
@@ -179,8 +179,8 @@ RANGES_CXX14_CONSTEXPR auto test_non_member_f(Sequence1234&& a) -> bool {
 RANGES_CXX14_CONSTEXPR auto test_array() -> bool {
     test::array<int, 4> a{{1, 2, 3, 4}};
 
-    auto beg = ranges::begin(a);
-    auto three = ranges::next(beg, 2);
+    // auto beg = ranges::begin(a);
+    // auto three = ranges::next(beg, 2);
```
gnzlbg: These lines are doing something (the code will only compile if begin and next are constant expressions). By uncommenting these lines, this is no longer tested.

The fix here is not to comment out these lines, but to uncomment the ones below. 
I'll see if I can do this in the constexpr algorithms branch by using safe_int instead of int all over the place in constexpr test.

ericniebler:
Looks like gcc doesn't like this one. Can you please take a look?

test/constexpr_core.cpp
gnzlbg: I just tested it and is possible, so i'm fine with commenting out these lines for now. 
I'll switch to safe_int in the constexpr algorithm PR and uncoment them without triggering the warning.

rhalbersma:
sorry, my git-fu is not up to par yet, and I messed up by pushing random_shuffle commits to this warning-level PR. I will close and repull.

PR #158, opened 2015-05-28
replace deprecated random_shuffle with shuffle

rhalbersma:
including use of ranges::random_shuffle and std::random_shuffle
except for definition and unit tests for ranges::random_shuffle

PR #159, opened 2015-05-28
(was #158) replace deprecated random_shuffle with shuffle

rhalbersma:
including use of ranges::random_shuffle and std::random_shuffle
except for definition and unit tests for ranges::random_shuffle

include <random> in make_heap.cpp

include <random> in min_element.cpp


ericniebler:
Thanks!

PR #161, opened 2015-05-29
silence -Wmissing-noreturn

rhalbersma:
replace RANGES_ASSERT(false) with `throw std::logic_error("class::function")` and add `[[noreturn]]`
this helps in dead code detection and increases the general warning level


include/range/v3/utility/variant.hpp
```diff
@@ -67,21 +67,21 @@ namespace ranges
...
+                [[noreturn]] void copy(std::size_t, variant_data const &) const
                 {
                     RANGES_ASSERT(false);
                 }
-                bool equal(std::size_t, variant_data const &) const
+                [[noreturn]] bool equal(std::size_t, variant_data const &) const
                 {
                     RANGES_ASSERT(false);
                     return true;
                 }
```
gnzlbg: Functions marked with `[[noreturn]]` cannot return anything. One would also need to remove the `return` statements in these functions: 

```c++
[[noreturn]] bool equal(std::size_t, variant_data const &) const
{
    RANGES_ASSERT(false);
}
```

rhalbersma: I will update later today after I set up my own travis to test more thoroughly. Perhaps a `return throw 42, true;` would also be an idea?

gnzlbg: I don't know. I never liked this. One should never call this functions, right now this is only checked in debug mode. I would rather have something like a call to std::terminate(), but trackable. So that means either just 

    throw "you should never call this because.... at line....";

or maybe just always call:

   __builtin_debug_trap();

which triggers an assertion and will make your debugger stop at the right place.

ericniebler:
I'm not a fan of this change. Throwing an exception on logic errors is generally a bad idea, as is removing the asserts. I would prefer adding a new macro, `RANGES_ENSURE`, that defaults to `RANGES_ASSERT` followed by a call to `std::terminate`.

ericniebler:
Even better would be to find a way to structure the code so that these functions can be removed, but I'm not sure it's possible.

ericniebler:
With my `RANGES_ENSURE` suggestion, I meant something more like:

```c++
#define RANGES_ENSURE(X) \
    ((X) ? void() : (RANGES_ASSERT(!#X), std::terminate()))
```

...and used like:

```c++
[[noreturn]] int some_function(/*...*/)
{
    RANGES_ENSURE(false);
}
```

If someone wants to `#define` `RANGES_ENSURE` to, say, throw an exception, they can.

rhalbersma:
I don't exactly see why you'd want `RANGES_ASSERT(!(X))` in the `X==false` branch instead of `RANGES_ASSERT((X))`, because wouldn't that delegate directly to `std::terminate` in that case?

ericniebler:
Look more closely. The assert is `RANGES_ASSERT(!#X)`. The `#` is significant.

rhalbersma:
OK, this requires a bit of care, because `-Wstring-conversion` is lurking around the corner, so I had to protect boolean conversions on error strings with the `true && MSG` trick (documented in Clang). Would two macros `RANGES_ENSURE` and `RANGES_ENSURE_MSG` be OK?

    #include <cassert>
    #include <stdexcept>

    #define RANGES_BOOLIFY(MSG) \
        (true && (MSG))
    
    #define RANGES_ENSURE(COND) \
        ((COND) ? void() : (assert(!RANGES_BOOLIFY(#COND)), std::terminate()))

    #define RANGES_ENSURE_MSG(COND, MSG) \
        ((COND) ? void() : (assert(!RANGES_BOOLIFY(MSG)), std::terminate()))

    int main() 
    {
        //RANGES_ENSURE(false);
       RANGES_ENSURE_MSG(false, "should never get here");
    }

ericniebler:
You can implement `RANGES_ENSURE` (unconditionally!) in terms of `RANGES_ENSURE_MSG`, and give people the option to provide their own `RANGES_ENSURE_MSG`. I suppose we could do the same with `RANGES_ASSERT`.

rhalbersma:
ah yes, you mean with 

    #define RANGES_ENSURE(COND) \
        RANGES_ENSURE_MSG(COND, #COND)

OK, to summarize, compared to current master: I'll replace every `RANGES_ASSERT(false)` with ` `RANGES_ENSURE(false)` and add `[[noreturn]]`?

ericniebler:
Yes, I think that would be an improvement. When you make this change, please rebase your pull request on master so it's one commit instead of *N*. Thanks.

ericniebler:
Thanks.

PR #169, opened 2015-06-11
Get range trait definitions from Range, not View.

CaseyCarter:
Not a functional change, simply uses a different access path to the same entities for documentation purposes. I.e., to make it clear that the range traits don't depend on View concepts.


ericniebler:
Oh right. Thanks. I'll merge when travis is done.

Issue #164, opened 2015-06-07
Should range_facade support containers?


CaseyCarter:
Commit 35457471d9f2a437abb48e9896c2f35d7dc71e6b "fixes for and misc clean up of mutable ranges" included a [change to `range_access`](https://github.com/ericniebler/range-v3/commit/35457471d9f2a437abb48e9896c2f35d7dc71e6b#diff-fbed09fe503554aa56f090fcaf93a5e5L85) to prefer `const` flavors of `begin_cursor`/`end_cursor` when they are available.  This unfortunately makes `range_facade` unusable to implement containers since a container with only `const` iterators is not ideal.

It would appear that supporting non-view ranges is an intended use case for the `range_foo` utilities, as they were not renamed to `view_foo` in the great renaming. If that is in fact the case then we need to fix this const-preference issue. I experimentally ripped out the const-preference overloads (caseycarter/range-v3@466d3fcce9d5d25f3b808fa5d7590ee3f1480b9a) to see what broke, so I could start thinking about a fix that would work for both mutable views *and* containers. Oddly enough [nothing broke](https://travis-ci.org/CaseyCarter/range-v3/builds/65751655).

Under the assumption that this change must have been made for a reason, I won't submit a pull request to simply tear it back out. I need guidance on how to proceed from here: why was this change made? Why is it necessary to prefer `const` `begin_cursor`/`end_cursor` for ranges that have both `const` and non-`const` flavors?

ericniebler:
Views have reference semantics. Just as a `char *` and a `char *const`, when they are dereferenced, yield a `char&`, so too a top-level const qualifier on a view should not effect the return types of `begin()` and `end()`. The logic you refer to in `range_facade` is to give views that property, even if the view they wrap does not have that property. (The reason you didn't see anything break when you changed the logic was because there are no views in range-v3 that are sensitive to top-level const qualification in this way.)

Should `range_facade` and `range_adaptor` be renamed to `view_facade` and `view_adaptor`? Yep, I missed those. Thanks.

PR #170, opened 2015-06-12
View fixes

CaseyCarter:
Partially addresses #164 - future work will (hopefully) separate the view-specific functionality from the generic range functionality. range-v3 doesn't contain any containers - see what I did there? - but it would be nice to have generic range functionality that makes it easy to write containers that integrate well with the library.

Note that I added range_interface into the renaming since it inherits basic_view.

### DIFF
--- a/example/calendar.cpp
+++ b/example/calendar.cpp
@@ -199,7 +199,7 @@ auto chunk(std::size_t n) {
 // Flattens a range of ranges by iterating the inner
 // ranges in round-robin fashion.
 template<class Rngs>
-class interleave_view : public range_facade<interleave_view<Rngs>> {
+class interleave_view : public view_facade<interleave_view<Rngs>> {
     friend range_access;
     std::vector<range_value_t<Rngs>> rngs_;
     struct cursor;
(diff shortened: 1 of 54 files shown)


gnzlbg:
> but it would be nice to have generic range functionality that makes it easy to write containers that integrate well with the library.

Hi Casey, I am not sure if I know what you mean by this.

To write a container I found the following range-v3 utilities useful: 
- concept checking and range concepts: to check if a container models a particular type of range, and
- iterator: to write the container iterators correctly (these are just amazing). 

There are other general utilities in the library that are really useful, but they are generally useful and not in particular for writing containers.

CaseyCarter:
> Hi Casey, I am not sure if I know what you mean by this.

I should have been more specific: I meant range_interface/range_facade. The algorithms/concepts and even views are extremely useful, e.g., `container_constructor(size_t n, const T& t)` is trivially implemented by delegating to `container_constructor(ranges::view::repeat_n(ranges::ref(t), n))`. `basic_iterator` is absolutely amazing.

`view_interface` and `view_facade` specifically are a small refactor away from providing default implementations of empty/begin/end/front/back/size/minimal stream insertion/pythonic slicing that work for containers as well as views.

ericniebler:
> `basic_iterator` is absolutely amazing.

It is lifted directly from Boost.Iterators. It is not my innovation.

PR #174, opened 2015-06-16
Update Concepts for decomposition design

CaseyCarter:
As discussed in ericniebler/stl2#10.

Note that `ConstructibleRef` is a positively terrible name for "Constructible object or reference type."

Also, I'm not convinced that `CopyConstructible<Foo> && Movable<Foo>` is a meaningful constraint for algorithms that want to memoize the return value of a projection a la adjacent_difference. I can't imagine there are many useful types that are copy constructible and move assignable but not copy assignable - I don't think there would be anything lost by simply requiring `Copyable`.

### DIFF
--- a/include/range/v3/utility/concepts.hpp
+++ b/include/range/v3/utility/concepts.hpp
@@ -335,10 +335,13 @@ namespace ranges
 
             struct Destructible
             {
-                template<typename T>
-                auto requires_(T&&) -> decltype(
+                template<typename T, typename = meta::if_<std::is_object<T>>>
+                auto requires_(T&& t, T* p = nullptr) -> decltype(
                     concepts::valid_expr(
-                        concepts::is_true(std::is_destructible<T>{})
+                        (t.~T(), 42),
+                        concepts::has_type<T*>(&t),
+                        (delete p, 42),
+                        (delete[] p, 42)
                     ));
             };
 
@@ -348,27 +351,28 @@ namespace ranges
                 template<typename T, typename ...Us>
                 auto requires_(T&&, Us&&...) -> decltype(
                     concepts::valid_expr(
-                        concepts::is_true(std::is_constructible<T, Us...>{})
+                        T(std::declval<Us>()...),
+                        new T(std::declval<Us>()...)
                     ));
             };
 
-            struct DefaultConstructible
-              : refines<Destructible>
+            struct Initializable
             {
-                template<typename T>
-                auto requires_(T&&) -> decltype(
+                template<typename T, typename ...Us>
+                auto requires_(T&&, Us&&...) -> decltype(
                     concepts::valid_expr(
-                        concepts::is_true(std::is_default_constructible<T>{})
+                        T(std::declval<Us>()...),
+                        concepts::is_true(meta::or_<
+                            std::is_reference<T>, models<Constructible, T, Us...>>())
                     ));
             };
 
             struct MoveConstructible
-              : refines<Destructible>
             {
                 template<typename T>
                 auto requires_(T&&) -> decltype(
                     concepts::valid_expr(
-                        concepts::is_true(std::is_move_constructible<T>{})
+                        concepts::model_of<Constructible, T, T&&>()
                     ));
             };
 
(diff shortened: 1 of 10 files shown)


CaseyCarter:
Well, this is odd. [This commit compiled fine in my Travis](https://travis-ci.org/CaseyCarter/range-v3/builds/67081671). Looking at the [error details](https://travis-ci.org/ericniebler/range-v3/jobs/67081691), it seems that particular build node doesn't have clang++-3.7. Restart the job, maybe?

gnzlbg:
@CaseyCarter It seems travis couldn't get llvm-3.7 from llvm.org package repository (error 404). 
You can retry on travis, let me know if this persists.

CaseyCarter:
I rebased the commit onto master to trigger a new Travis build - we'll hope it works this time.

CaseyCarter:
...and I'm a moron and just closed the pull request.

gnzlbg:
@CaseyCarter seems to be working fine now. Some times travis builds fails for reasons out of our control (out of memory, internet connection...). When one of the jobs fails it is possible to restart that single job. For PRs, I think both the submitter and the receiver can do it.

CaseyCarter:
@gnzlbg The cancel/retry button is not there - I assume that authority is owner-only.

include/range/v3/utility/concepts.hpp
```diff
@@ -655,8 +626,13 @@ namespace ranges
...
 
-        template<typename T>
-        using DefaultConstructible = concepts::models<concepts::DefaultConstructible, T>;
+        template<typename T, typename...Us>
+        using ConstructibleRef =
+            meta::apply<
+                meta::if_<std::is_reference<T>,
+                    meta::quote<std::is_constructible>,
+                    meta::quote<Constructible>>,
+                T, Us...>;
```
ericniebler: Can you explain this to me?

ericniebler: Is it because Constructible now requires `new T(args...)`? That will fail with reference types. It makes me worry we haven't gotten the Constructible concept right. But assuming it's ok, why can't we replace uses of ConstructibleRef with `is_constructible`?

ericniebler:
@CaseyCarter says:
> The cancel/retry button is not there - I assume that authority is owner-only.

I can see the retry button.

include/range/v3/utility/concepts.hpp
CaseyCarter: Everything in the hierarchy between Destructible - Regular is specific to object types. Constructible will reject reference types because of (a) the `new` requirement, (b) Destructible's requirement for `t.~T()`, and (c) Destructible's formation of a pointer-to-T.

ConstructibleRef - which I probably should have named `ConstructibleObjectOrReference`- exists to constrain variables of type T to have either an object type that is Constructible from Args or a reference type that can be bound to Args. It's used by the constructors of pair and tuple to ensure both that the members can be constructed from the constructor arguments and that the constructed pair/tuple will be Destructible. If we simply replace ConstructibleRef uses with is_constructible the guarantee of a Destructible tuple/pair is lost, violating my "Construction implies destruction" design axiom.

ericniebler: I see your point. It's a little confusing that some concepts expect object types and others don't (e.g. `Swappable`). And `ConstructibleObjectOrReference` is still an awful name. References are not constructed; they are initialized. So maybe `Initializable`? Egad. :-P

ericniebler: And should `Constructible` subsume `Initializable`? In this pull request, `ConstructibleRef` is a bit of a bolt-on afterthought.

CaseyCarter: > And should Constructible subsume Initializable? In this pull request, ConstructibleRef is a bit of a bolt-on afterthought.

In Concepts, an equivalent definition would be:
```
template <class T, class...Args>
concept bool ConstructibleRef =
  Constructible<T, Args...> ||
  (std::is_reference<T>::value && requires(Args&&...args) {
    T(forward<Args>(args)...);
  });
```
from which it is clear that Constructible<T, Args...> subsumes ConstructibleRef<T, Args...>. The formulation I used for range-v3 performs the conditional early to avoid instantiating the failing branch for "performance" and to avoid a hard error in an earlier version of Destructible that wasn't guarded with `meta::if_<std::is_object<T>>`.

CaseyCarter: Would you prefer a "formal" concept declaration a la:
```
struct Initializable
{
  template <typename T, typename ...Us,
      typename = meta::if_c<!std::is_reference<T>::value>>
  auto requires_(T&&, Us&&...) -> decltype(
      concepts::valid_expr(
          concepts::model_of<Constructible, T, Us...>()
      ));

  template <typename T, typename ...Us,
      typename = meta::if_c<std::is_reference<T>::value>>
  auto requires_(T&&, Us&&...) -> decltype(
      concepts::valid_expr(
          T(std::declval<Us>()...)
      ));
};
```

ericniebler: I was thinking more in terms of `Constructible` being implemented in terms of `Initializable`, but adding `Destructible` and dynamic allocation.

ericniebler:
```c++
struct Initializable
{
    template<typename T, typename ...Us>
    auto requires_(T&&, Us&&...) -> decltype(
        concepts::valid_expr(
            T(std::declval<Us>()...)
        ));
};

struct Constructible
    : refines<Initializable, Destructible(_1)>
{
    template<typename T, typename ...Us>
    auto requires_(T&&, Us&&...) -> decltype(
        concepts::valid_expr(
            new T(std::declval<Us>()...)
        ));
};
```

ericniebler: Oh, I see. Then we don't get `Destructible` checked for non-reference types.

ericniebler: Maybe ...

```c++
struct Initializable
{
    template<typename T, typename ...Us,
        typename TT = meta::if_<std::is_reference<T>, int, T>>
    auto requires_(T&&, Us&&...) -> decltype(
        concepts::valid_expr(
            T(std::declval<Us>()...),
            concepts::model_of<Destructible, TT>()
        ));
};
```

CaseyCarter: > Maybe ... `struct Initializable`

AFAICS that's equivalent to my definition of Initializable, but a bit less readable.

ericniebler: OK, if we want this, it should be expressed the other way 'round like you said:

```c++
struct Constructible
  : refines<Destructible(_1)>
{
    template<typename T, typename ...Us>
    auto requires_(T&&, Us&&...) -> decltype(
        concepts::valid_expr(
            T(std::declval<Us>()...),
            new T(std::declval<Us>()...)
        ));
};

struct Initializable
{
    template<typename T, typename ...Us>
    auto requires_(T&&, Us&&...) -> decltype(
        concepts::valid_expr(
            T(std::declval<Us>()...),
            concepts::is_true(
                meta::or_<std::is_reference<T>, models<Constructible, T, Us...>>())
        ));
};
```

CaseyCarter: Agreed. I was just typing that same definition of Intializable as you posted it - I'd call that convergence.

ericniebler:
Ah, you've updated the pull request. I've done the same on a branch. The only other thing I needed to change was that `back_insert_iterator` and `ostream_iterator` no longer modeled the updated `SemiRegular` concept. See e013530f. (This was exposed by the calendar example, but there should probably be separate tests for these components.)

CaseyCarter:
Poor Travis - I think I've pushed 10 different commits onto this branch now.

PR #177, opened 2015-06-19
Remove breakage of Assignable introduced in the decomposition patch.

CaseyCarter:
Partially reverts #174.

### DIFF
--- a/include/range/v3/utility/concepts.hpp
+++ b/include/range/v3/utility/concepts.hpp
@@ -393,7 +393,7 @@ namespace ranges
                 template<typename T, typename U>
                 auto requires_(T&& t, U&& u) -> decltype(
                     concepts::valid_expr(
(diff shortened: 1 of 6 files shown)


include/range/v3/utility/concepts.hpp
```diff
@@ -413,9 +413,9 @@ namespace ranges
                 template<typename T>
                 auto requires_(T&&) -> decltype(
                     concepts::valid_expr(
-                        concepts::model_of<Assignable, T, T&>(),
-                        concepts::model_of<Assignable, T, const T&>(),
-                        concepts::model_of<Assignable, T, const T&&>()
+                        concepts::model_of<Assignable, T&, T&>(),
+                        concepts::model_of<Assignable, T&, const T&>(),
+                        concepts::model_of<Assignable, T&, const T&&>()
```
ericniebler: For future reference, I put const after the type, like `T const &`. It's not worth changing, though.

PR #183, opened 2015-07-06
view::cycle

gnzlbg:
Turns a finite `ForwardRange` into an infinite range that cycles on itself. 

### Example

```c++
auto il = {0, 1, 2};
RANGES_FOR(auto i, il | view::cycle) { std::cout << i << " "; }
```

prints "0 1 2 0 1 2 0 1 2 0 1 2...". 

### Notes on the design 

This designs chooses to make two iterators compare equal only if they point to the same element of the sequence within the same cycle. Example: in the sequence above both iterators `begin` and `begin  + 3` point to the same element (`0`); since they are "in different cycles" they are not equal. 

The rationale is that this allows iterating from `[begin, begin + M)` where `M` is larger than the number of elements in the sequence.

### TODO

- The implementation of `cycled_view<Rng>::cursor<IsConst>::advance` is pretty ugly.

- The implementation of `cycled_view<Rng>::cursor` can be optimized for `RandomAccessRange`s where the iterator can be obtained from the position in `O(1)`.

### DIFF
--- a/include/range/v3/view/cycle.hpp
+++ b/include/range/v3/view/cycle.hpp
@@ -1,0 +1,207 @@
+/// \file cycle.hpp
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2015
+//  Copyright Gonzalo Brito Gadeschi 2015
+//  Copyright Casey Carter 2015
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_CYCLE_HPP
+#define RANGES_V3_VIEW_CYCLE_HPP
+#include <utility>
+#include <type_traits>
+#include <meta/meta.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/size.hpp>
+#include <range/v3/distance.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/view_facade.hpp>
+#include <range/v3/view/all.hpp>
+#include <range/v3/view/view.hpp>
+#include <range/v3/utility/iterator.hpp>
+#include <range/v3/utility/static_const.hpp>
+
+namespace ranges
+{
+
+    inline namespace v3
+    {
+
+        /// \addtogroup group-views
+        ///@{
+        template <typename Rng>
+        struct cycled_view : view_facade<cycled_view<Rng>, infinite>
+        {
+
+        private:
+            CONCEPT_ASSERT(ForwardRange<Rng>());
+            friend range_access;
+            Rng rng_;
+
+        public:
+            cycled_view() = default;
+            explicit cycled_view(Rng rng) : rng_(std::move(rng))
+            {
+            }
+
+        private:
+            template <bool IsConst>
+            struct cursor
+            {
+            private:
+                template <typename T>
+                using constify_if = meta::apply<meta::add_const_if_c<IsConst>, T>;
+                using cycled_view_t = constify_if<cycled_view>;
+            public:
+                using difference_type = range_difference_t<Rng>;
+                using reference = range_reference_t<constify_if<Rng>>;
+
+            private:
+                cycled_view_t *rng_;
+                range_iterator_t<constify_if<Rng>> it_;
+
+            public:
+                cursor() = default;
+                explicit cursor(cycled_view_t &rng)
+                    : rng_(&rng), it_(ranges::begin(rng.rng_))
+                {
+                }
+
+                constexpr bool done() const
+                {
+                    return false;
+                }
+
+                reference current() const
+                {
+                    return *it_;
+                }
+
+                void next()
+                {
+                    const auto end = ranges::end(rng_->rng_);
+                    if (it_ != end)
+                    {
+                        ++it_;
+                    }
+                    if (it_ == end)
+                    {
+                        it_ = ranges::begin(rng_->rng_);
+                    }
+                }
+
+                bool equal(cursor const &pos) const
+                {
+                    RANGES_ASSERT(rng_ == pos.rng_);
+                    return it_ == pos.it_;
+                }
+
+                CONCEPT_REQUIRES(BidirectionalRange<Rng>{})
+                void prev()
+                {
+                    if (it_ == ranges::begin(rng_->rng_))
+                    {
+                        it_ = ranges::end(rng_->rng_);
+                    }
+                    --it_;
+                }
+
+                CONCEPT_REQUIRES(RandomAccessRange<Rng>())
+                void advance(difference_type n)
+                {
+                    if (n == difference_type{0})
+                    {
+                        return;
+                    }
+
+                    const auto begin = ranges::begin(rng_->rng_);
+                    const auto end = ranges::end(rng_->rng_);
+
+                    auto const N = end - begin;
+                    if (N == 0)
+                    {
+                        return;
+                    }
+
+                    auto pos = it_ - (begin + n);
+                    if (pos < 0 || pos >= N)
+                    {
+                      pos %= N;
+                      if (pos < 0)
+                      {
+                        pos += N;
+                      }
+                    }
+                    it_ = begin + pos;
+                }
+
+                CONCEPT_REQUIRES(RandomAccessRange<Rng>())
+                difference_type distance_to(cursor const &that) const
+                {
+                    RANGES_ASSERT(that.rng_ == rng_);
+                    return that.it_ - it_;
+                }
+            };
+
+            cursor<false> begin_cursor()
+            {
+                return cursor<false>{*this};
+            }
+            cursor<true> begin_cursor() const
+            {
+                return cursor<true>{*this};
+            }
+        };
+
+        namespace view
+        {
+
+            struct cycle_fn
+            {
+            private:
+                friend view_access;
+                template <class T>
+                using Concept = meta::and_<ForwardRange<T>, BoundedRange<T>>;
+
+            public:
+                template <typename Rng, CONCEPT_REQUIRES_(Concept<Rng>())>
+                cycled_view<all_t<Rng>> operator()(Rng &&rng) const
+                {
+                    return cycled_view<all_t<Rng>>{all(std::forward<Rng>(rng))};
+                }
+
+#ifndef RANGES_DOXYGEN_INVOKED
+                template <typename Rng, CONCEPT_REQUIRES_(!Concept<Rng>())>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(ForwardRange<Rng>(),
+                                       "The object on which view::cycle operates must be a "
+                                       "model of the ForwardRange concept.");
+                    CONCEPT_ASSERT_MSG(BoundedRange<Rng>(),
+                                       "To cycle a range object, its end iterator must be a "
+                                       "model of the ForwardIterator concept.");
+                }
+#endif
+            };
+
+            /// \relates cycle_fn
+            /// \ingroup group-views
+           namespace
+           {
+               constexpr auto &&cycle = static_const<view<cycle_fn>>::value;
+           }
+
+       } // namespace view
+       /// @}
+    } // namespace v3
+} // namespace ranges
+#endif
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -421,6 +421,14 @@ namespace ranges
         using counted_sentinel =
             basic_sentinel<detail::counted_sentinel>;
 
+        template<typename Rng>
+        struct cycled_view;
+
+        namespace view
+        {
+            struct cycle_fn;
+        }
+
         namespace detail
         {
             template<typename I> struct reverse_cursor;
--- a/test/view/CMakeLists.txt
+++ b/test/view/CMakeLists.txt
@@ -23,6 +23,9 @@ add_test(test.view.const, view.const)
 add_executable(view.counted counted.cpp)
 add_test(test.view.counted, view.counted)
 
+add_executable(view.cycle cycle.cpp)
+add_test(test.view.cycle, view.cycle)
+
 add_executable(view.delimit delimit.cpp)
 add_test(test.view.delimit, view.delimit)
 
--- a/test/view/cycle.cpp
+++ b/test/view/cycle.cpp
@@ -1,0 +1,302 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2015
+//  Copyright Gonzalo Brito Gadeschi 2015
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+
+#include <array>
+#include <list>
+#include <forward_list>
+#include <memory>
+#include <range/v3/range_for.hpp>
+#include <range/v3/algorithm/count_if.hpp>
+#include <range/v3/view/cycle.hpp>
+#include <range/v3/view/take.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/reverse.hpp>
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+
+using namespace ranges;
+
+template <typename Rng> void test_const_forward_range(Rng const &rng)
+{
+    auto r = rng | view::cycle;
+    static_assert(is_infinite<decltype(r)>{}, "");
+    static_assert(!BoundedRange<decltype(r)>{}, "");
+    CHECK(distance(r | view::take_exactly(0)) == 0);
+    CHECK(distance(r | view::take_exactly(1)) == 1);
+    CHECK(distance(r | view::take_exactly(2)) == 2);
+    CHECK(distance(r | view::take_exactly(3)) == 3);
+    CHECK(distance(r | view::take_exactly(4)) == 4);
+    CHECK(distance(r | view::take_exactly(6)) == 6);
+    CHECK(distance(r | view::take_exactly(7)) == 7);
+    CHECK(count_if(r | view::take_exactly(7), [](int) { return true; }) == 7);
+
+    ::check_equal(r | view::take_exactly(0), std::array<int, 0>{});
+    ::check_equal(r | view::take_exactly(1), {0});
+    ::check_equal(r | view::take_exactly(2), {0, 1});
+    ::check_equal(r | view::take_exactly(3), {0, 1, 2});
+    ::check_equal(r | view::take_exactly(4), {0, 1, 2, 0});
+    ::check_equal(r | view::take_exactly(6), {0, 1, 2, 0, 1, 2});
+    ::check_equal(r | view::take_exactly(7), {0, 1, 2, 0, 1, 2, 0});
+
+    CHECK(distance(r | view::take(0)) == 0);
+    CHECK(distance(r | view::take(1)) == 1);
+    CHECK(distance(r | view::take(2)) == 2);
+    CHECK(distance(r | view::take(3)) == 3);
+    CHECK(distance(r | view::take(4)) == 4);
+    CHECK(distance(r | view::take(6)) == 6);
+    CHECK(distance(r | view::take(7)) == 7);
+    CHECK(count_if(r | view::take(7), [](int) { return true; }) == 7);
+
+    ::check_equal(r | view::take(0), std::array<int, 0>{});
+    ::check_equal(r | view::take(1), {0});
+    ::check_equal(r | view::take(2), {0, 1});
+    ::check_equal(r | view::take(3), {0, 1, 2});
+    ::check_equal(r | view::take(4), {0, 1, 2, 0});
+    ::check_equal(r | view::take(6), {0, 1, 2, 0, 1, 2});
+    ::check_equal(r | view::take(7), {0, 1, 2, 0, 1, 2, 0});
+}
+
+template <typename Rng> void test_const_forward_reversed_range(Rng const &rng)
+{
+    test_const_forward_range(rng);
+
+    auto r = rng | view::reverse | view::cycle;
+    static_assert(is_infinite<decltype(r)>{}, "");
+    static_assert(!BoundedRange<decltype(r)>{}, "");
+
+    CHECK(distance(r | view::take_exactly(0)) == 0);
+    CHECK(distance(r | view::take_exactly(1)) == 1);
+    CHECK(distance(r | view::take_exactly(2)) == 2);
+    CHECK(distance(r | view::take_exactly(3)) == 3);
+    CHECK(distance(r | view::take_exactly(4)) == 4);
+    CHECK(distance(r | view::take_exactly(6)) == 6);
+    CHECK(distance(r | view::take_exactly(7)) == 7);
+    CHECK(count_if(r | view::take_exactly(7), [](int) { return true; }) == 7);
+
+    ::check_equal(r | view::take_exactly(0), std::array<int, 0>{});
+    ::check_equal(r | view::take_exactly(1), {2});
+    ::check_equal(r | view::take_exactly(2), {2, 1});
+    ::check_equal(r | view::take_exactly(3), {2, 1, 0});
+    ::check_equal(r | view::take_exactly(4), {2, 1, 0, 2});
+    ::check_equal(r | view::take_exactly(6), {2, 1, 0, 2, 1, 0});
+    ::check_equal(r | view::take_exactly(7), {2, 1, 0, 2, 1, 0, 2});
+
+    CHECK(distance(r | view::take(0)) == 0);
+    CHECK(distance(r | view::take(1)) == 1);
+    CHECK(distance(r | view::take(2)) == 2);
+    CHECK(distance(r | view::take(3)) == 3);
+    CHECK(distance(r | view::take(4)) == 4);
+    CHECK(distance(r | view::take(6)) == 6);
+    CHECK(distance(r | view::take(7)) == 7);
+    CHECK(count_if(r | view::take(7), [](int) { return true; }) == 7);
+
+    ::check_equal(r | view::take(0), std::array<int, 0>{});
+    ::check_equal(r | view::take(1), {2});
+    ::check_equal(r | view::take(2), {2, 1});
+    ::check_equal(r | view::take(3), {2, 1, 0});
+    ::check_equal(r | view::take(4), {2, 1, 0, 2});
+    ::check_equal(r | view::take(6), {2, 1, 0, 2, 1, 0});
+    ::check_equal(r | view::take(7), {2, 1, 0, 2, 1, 0, 2});
+}
+
+template <typename Rng> void test_mutable_forward_range_reversed(Rng &rng)
+{
+    test_const_forward_reversed_range(rng);
+    int count = 2;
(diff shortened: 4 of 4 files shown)


include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,259 @@
...
+
+            private:
+                friend struct sentinel<IsConst>;
+                cycled_view_t *rng_;
+                // TODO: RandomAccessRange's don't need to store the iterator
+                // since it can be obtained from `begin` in O(1).
+                // Note: Bidirectional and Forward ranges need to store the
+                // iterator for `current` to be O(1).
+                range_iterator_t<constify_if<Rng>> it_;
+                difference_type pos_;
```
ericniebler: I've learned from Sean and Andrew that cyclic iteration is well within the design of STL's iterators. In other words, you don't need to maintain `pos_` so that `difference_to` has something to return. In yet other words, it's OK if `(next(it, n) - it) != n`.

CaseyCarter:
> This designs chooses to make two iterators compare equal only if they point to the same element of the sequence within the same cycle.

I'm reasonably certain this is a mistake, in that it hides the fact that iterators alias the same element. I believe that cyclic sequences like `view::cycle` and `view::repeat` must have cyclic iterators as well to make that aliasing explicit. If you want to create a finite sequence from the cyclic sequence, you can always `view::take(M)`.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,259 @@
...
+                        {
+                            n -= rel_distance;
+                            it_ = begin + distance - n % distance;
+                            if (it_ == end)
+                            {
+                                it_ = begin;
+                            }
+                        }
+                    }
+                }
```
ericniebler: I strongly suspect this can be simpler.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,259 @@
...
+
+            template <bool IsConst>
+            struct sentinel
+            {
+                sentinel() = default;
+                constexpr bool equal(cursor<IsConst> const &) const
+                {
+                    return false;
+                }
+            };
```
ericniebler: If you put a `done` method on the cursor, you don't need to define a sentinel at all.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,259 @@
...
+            }
+
+            sentinel<false> end_cursor()
+            {
+                return sentinel<false>{};
+            }
+            sentinel<true> end_cursor() const
+            {
+                return sentinel<true>{};
+            }
```
ericniebler: .. and once you delete `sentinel`, you can delete these `end_cursor` overloads.

gnzlbg:
@CaseyCarter 

>>This designs chooses to make two iterators compare equal only if they point to the same element of the sequence within the same cycle.
>
> I'm reasonably certain this is a mistake, [...]  If you want to create a finite sequence from the cyclic sequence, you can always view::take(M).

I have mixed feelings about this too. I started with making iterators that alias the same element to compare equal. The problem I ran into was the following: 

```c++
auto il = {0, 1, 2};
auto rng = il | view::cyclic | view::take(6);  // [0, 1, 2, 0, 1, 2]
auto b = begin(rng); // alias 0
auto e = end(rng); // alias 0:   0, 1, 2, 0, 1, 2, 0
//                                                 ^one past the end
```

Here `b` and `e` alias the same element. If they compare equal, the condition `b == e` is `true`, and iteration breaks. I thus decided to make them compare differently, and at least for the tests I wrote everything worked out fine, but it still doesn't feel right. 

Do you have a better solution to this problem?  I am going to check `repeat`, since maybe Eric came already with a solution to this problem there.

include/range/v3/view/cycle.hpp
gnzlbg: neat, thanks!

include/range/v3/view/cycle.hpp
gnzlbg: The only reason I maintain `pos_` is to make sure in this example:

```c++
auto il = {0, 1, 2};
auto rng = il | view::cyclic | view::take(6);  // [0, 1, 2, 0, 1, 2]
auto b = begin(rng); // alias 0
auto e = end(rng); // alias 0:   0, 1, 2, 0, 1, 2, 0
//                                                 ^one past the end
```

that `b` and `e` compare equal only when `b` has already traversed the range twice. 

@CaseyCarter pointed out that there might be a better solution to this problem.

CaseyCarter:
`take_view::sentinel::equal` is defined:
```
bool equal(detail::counted_cursor<base_iterator> const &that) const
{
    return 0 == that.count() || that.base() == sent_;
}
```
so `take(n)` is "take at most n". You and I are both wrong, you want `take_exactly(n)`.

include/range/v3/view/cycle.hpp
gnzlbg: Yes me too. It is too complicated.

CaseyCarter: After discarding `pos_`, it becomes:
```
auto pos = it_ - begin + n;
auto const N = end - begin;
if (pos < 0 || pos >= N) {
  pos %= N;
  if (pos < 0) {
    pos += N;
  }
}
it_ = begin + pos;
```

CaseyCarter: One further examination, I think the reason `take` doesn't work is because your view is container-like: the type of the cursor (and hence iterator) differs based on the constness of the view, which is almost certainly incorrect. views have reference semantics - there should not be a non-`const` overload of `begin_cursor`/`end_cursor`.

gnzlbg: @CaseyCarter Thanks! I have not removed `pos_` yet since I haven't made the tests to work with `take_exactly` yet but I'll keep trying. I've just committed your implementation of advance, is the copyright correct?

> Removing pos_ also allows you use the underlying range's difference_type directly.

I wasn't sure about this. The underlying range's `difference_type` might be very short, but through `view::cycle` which is infinite one might want to pass a much larger `n` to `advance`, so I thought the `difference_type` of the cursor should be `max<range_difference_t<Rng>, std::ptrdiff_t>` at least. 

> views have reference semantics - there should not be a non-const overload of `begin_cursor`/`end_cursor`.

Maybe I misunderstood views completely, but how do you handle mutation/non-mutation then? IIUC a const view provides `const_iterator`s, while a non-const view provides `iterator`s.

include/range/v3/view/cycle.hpp
ericniebler: `cycle` is an infinite range. You can't decrement the end of an infinite range, period. `end` should be a sentinel type, and it should *never* compare equal to any iterator into the range. What am I missing?

include/range/v3/view/cycle.hpp
ericniebler: This is a common confusion. Views have reference semantics; there is always a (logical) indirection. You can mutate an int through an `int*` or through an `int*const`. The const doesn't matter at the top level when you have an indirection.

ericniebler:
> I am going to check repeat, since maybe Eric came already with a solution to this problem there.

I coded it cautiously since I didn't know Sean's and Andrew's views on cyclic iteration. I haven't had a chance to fix it yet.

include/range/v3/view/cycle.hpp
CaseyCarter: On Mon, Jul 6, 2015 at 1:22 PM, gnzlbg <notifications@github.com> wrote:

> @CaseyCarter <https://github.com/CaseyCarter> Thanks! I have not removed
> pos_ yet since I haven't made the tests to work with take_exactly yet but
> I'll keep trying. I've just committed your implementation of advance, is
> the copyright correct?
>
Yip. That's fine.

>  Removing pos_ also allows you use the underlying range's difference_type
> directly.
>
> I wasn't sure about this. The underlying range's difference_type might be
> very short, but through view::cycle which is infinite one might want to
> pass a much larger n to advance, so I thought the difference_type of the
> cursor should be max<range_difference_t<Rng>, std::ptrdiff_t> at least.
>
That's a good point. I retract that comment.


>  views have reference semantics - there should not be a non-const
> overload of begin_cursor/end_cursor.
>
> Maybe I misunderstood views completely, but how do you handle
> mutation/non-mutation then? IIUC a const view provides const_iterators,
> while a non-const view provides iterators.
>
The mutability of the iterators of a view follows from the mutability of
the elements of the range it denotes, just like a pointer to an array
element.


-- 
Casey Carter

ericniebler: > One further examination, I think the reason `take` doesn't work is because your view is container-like: the type of the cursor (and hence iterator) differs based on the constness of the view, which is almost certainly incorrect. views have reference semantics - there should not be a non-const overload of `begin_cursor`/`end_cursor`.

Actually there's nothing fundamentally wrong with a view that uses a top-level const this way. It's a common request, despite the fact that I believe it's always wrong. Most of the views are implemented to be sensitive to const *if the underlying view is also sensitive to const*. None of the views provided by range-v3 are.

Why doesn't `take` work on a `cycle` view?

include/range/v3/view/cycle.hpp
gnzlbg: Agreed, that should be achieved via the `done` method on the cursor always returning `false` right?

In the example above `end` is the end iterator of a `take` view (over a cyclic range). 

Basically the "problem" `pos_` solved was that if my sequence seq has N elements, I cycle it `auto rng = seq | view::cycled`, and then I take the slice `[begin(rng), begin(rng) + 3*N)`, the distance of that slice should be `3*N`. If pointing to the same element is sufficient for two iterators to compare equal, the slice `[begin(rng), begin(rng) + 3*N)` might have distance zero (since `begin(rng)  + M * N` points to `begin(rng)`).

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+        public:
+            cycled_view() = default;
+            explicit cycled_view(Rng rng) : rng_(std::move(rng))
+            {
+            }
+
+        private:
+            template <bool IsConst> struct cursor;
+            template <bool IsConst> struct sentinel;
```
ericniebler: Delete this line.

include/range/v3/view/cycle.hpp
ericniebler: > I thought that the distance of that slice should be `3*N`. If pointing to the same element is sufficient for two iterators to compare equal, the slice `[begin(rng), begin(rng) + 3*N)` might have distance zero (since `begin(rng) + M * N` points to `begin(rng)`).

I thought that, too. It's not right. Iterators make no such guarantee, as I've said.

CaseyCarter:
`take_exactly(n)` (and indirectly, `take(n)`) don't work with cyclic random access iterators when `n` is greater than or equal to the cycle length, since they believe that `(i + n) - i == n`, whereas the iterator requirements for `SizedIteratorRange` specify that `(i + n) - i` is the *smallest* value of `DifferenceType` `k` such that `(i + n) - i == k`.

rhalbersma:
> (I can't exactly blame someone for thinking that [i, i + n) denotes a sequence of n elements ;))

If that range were to yield `true` on a hypothetical trait `is_cyclical<Rng>` I would expect `size(rng) == n % cycle_length<Rng>` (or the equivalent if you want to have a runtime cycle length)

CaseyCarter:
The presentation in EoP 6 makes it clear what is going on here: For iterators `f` and `l`of the same type `I` and `n` of type `DifferenceType<I>` the `weak_range` `[f, n)` includes all of the iterators `f + i` where `0 <= i < n`. The sole reason to define `counted_range` is to refine `weak_range` with the additional axiom that there are no cycles in the iterator values. If `l == f + n`, the `bounded_range` `[f, l)` is equivalent to the `counted_range` `[f, n)` - but not necessarily to the `weak_range` `[f, n)` due to the possible existence of cycles.

So the problem arises because `take(r, n)` assumes that the `weak_range` `[begin(r), n)` is a `counted_range` which it can convert to a `bounded_range` `[begin(r), next(begin(r), n))` when `r` models `RandomAccessRange`. This is of course only true for the cyclic ranges `repeat` and `cycle(range)` when `n` is less than the cycle length [1 for `repeat` and `distance(range)` for `cycle(range)`].

[I made `view::repeat` model `RandomAccessView` for kicks](https://github.com/CaseyCarter/range-v3/commit/6bf713664bbf7aa7d1e06e15cd69edbcdc451afe), which interacts hilariously with `take(n)` . If `r` is the range `view::repeat(foo)`,  `begin(r) == begin(r) + n` for all `n`, so `take(n)` is always an empty range.

If we want cyclic ranges to work, we probably need a trait to differentiate them. Its notable that cyclic ForwardRanges are necessarily infinite since the increment operations on ForwardIterator are equality preserving, but ranges weaker than Forward can be both cyclic and finite.

gnzlbg:
> Its notable that cyclic ForwardRanges are necessarily infinite since the increment operations on ForwardIterator are equality preserving, but ranges weaker than Forward can be both cyclic and finite.

I've left ranges weaker than `ForwardRange` out for the moment.

> If we want cyclic ranges to work, we probably need a trait to differentiate them; there are no syntactic differences, so our choices are trait or preconditions on differently named algorithms.

@CaseyCarter @ericniebler I've added a trait `is_cyclic_range` to `range_concepts.hpp` that is always false but users can specialize for their own ranges if the default is wrong. I've specialized it to true for `view::cyclic`, `view::repeat`, and `view::repeat_n`. 

I was able to remove `pos_` and get the tests to work with `view::take_exactly`. As Casey pointed out above, the problem was with the specialization for `RandomAccessRange`, which assumed that the range is not cyclic. This specialization is now just active for `RandomAccessRange && !Cyclic`. 

The trait works for now, but this appears to me to be common enough to warrant a `CyclicRange` concept.

ilyapopov:
I believe the word "cyclic" is misleading in `is_cyclic_range`, because what really causes problem, is that there can be iterators within the range that compare equal. This can happen not only in cyclic ranges, but also, for example, in a result of hypothetical `random_sampling_with_repetitions`. 

Thus I think it better be `has_nonunique_iterators` or something like that.

ericniebler:
> I've added a trait `is_cyclic_range`

A cyclic range has cyclic iterators. So "cyclic" seems a property of the iterators to me, not the range.

ericniebler:
I think I prefer simply fixing `view::take` and `view::take_exactly` to work with cyclic ranges by removing the faulty assumption. If we can get away with not adding a new trait or concept, it's a win.

gnzlbg:
@ericniebler  `view::take`doesn't have any problems (I just committed some tests for it and it works fine). The only problem is that `view::take_exactly` uses the following optimization for `RandomAccessRanges`: 

```c++
range_iterator_t<Rng> end()
{
}
```

instead of a `counted_sentinel` (which is slower). This only works for "bijective" iterators[*], and since this is a semantic requirement I don't see how to enable this optimization without introducing a new trait. Do you have any ideas?


[*] those iterators that while traversing the sequence only points to each element once.

CaseyCarter:
> A cyclic range has cyclic iterators. So "cyclic" seems a property of the iterators to me, not the range.

Agreed. The existence of `repeat_n` disproves my statement that "cyclic ForwardRanges are necessarily infinite", which would more accurately be "cyclic sequences of ForwardIterators are necessarily infinite."

> I think I prefer simply fixing `view::take` and `view::take_exactly` to work with cyclic ranges by removing the faulty assumption.

> The only problem is that `view::take_exactly` uses the following optimization for `RandomAccessRanges`

I suppose we need to determine (a) if `take_exactly` is the only library component that assumes a weak range can be converted to a bounded range, and (b) if the optimization achieved via that assumption is significant enough to warrant adding a trait to preserve it.

ericniebler:
@gnzlbg says:
> instead of a `counted_sentinel` (which is slower)

How much slower? Have you measured? Is it worth complicating the interface by adding a new trait or concept? I would rather KISS and optimize later if it is necessary (and it probably isn't). As Sean says, not all problems are worth solving.

@CaseyCarter says:
I suppose we need to determine (a) if `take_exactly` is the only library component that assumes a weak range can be converted to a bounded range

Probably not. I didn't know I was making a faulty assumption until late in the game. :-(

> and (b) if the optimization achieved via that assumption is significant enough to warrant adding a trait to preserve it.

Right. I think the answer is no, at least in the case of `take_exactly`. It might be a different story elsewhere.

ericniebler:
Also: whether an iterator is cyclic is data-dependent in principle. Consider a linked list. If there is a cycle in the nodes, the iterator is "cyclic"; otherwise, it isn't. What should `is_cyclic` report for a linked list iterator? This doesn't work.

morinmorin:
Does `cycle_n(n)` make sense?

For ForwardRange, `rng | view::cycle_n(n)` would be more efficient than `rng | view::cycle | view::take_exactly(distance(rng) * n)`.

gnzlbg:
> EDIT: For reference, the reason I implemented take_exactly the way I did for RandomAccessIterators was so that it could be a bounded range, not for any performance reasons.

Without this, neither `take` nor `take_exactly` are `BoundedRange`s, and some views like `view::reverse` won't work on them. I don't see a way of "fixing" the specialization for `RandomAccessRange`s without introducing a trait and while keeping `take_exactly` as a `BoundedRange`.

> Also: whether an iterator is cyclic is data-dependent in principle. Consider a linked list. If there is a cycle in the nodes, the iterator is "cyclic"; otherwise, it isn't. What should is_cyclic report for a linked list iterator? This doesn't work.

Indeed.

gnzlbg:
@CaseyCarter 

> Requiring BoundedRange when it's only needed to ensure that prev is O(1) is limiting. It would be nice to have separate overloads: the current operator() for ForwardRange, and another operator() for BidirectionalRange that finds the end iterator up front before constructing the view (i.e., returns cycle_view<bounded_view<Rng>>).

Agreed. The same argument could be made for `view::reverse`. IIRC @ericniebler wasn't very fond of doing work on view construction, but for example `view::remove_if` finds the begin iterator the first time `begin` is called, so I guess `view::reverse` could find the `end` iterator the first time `end` is called for not bounded ranges, which would allow it to work with unbounded `take` and `take_exactly` views.

CaseyCarter:
In the specific case of `view::cycle` you could cache the value of the end iterator since you technically don't need to know it until it has been seen. Decrementing the first iterator of a range has undefined behavior, so the only way a client could legally decrement an iterator equal to the begin iterator of the adapted range is if they obtained that iterator by applying enough increments to wrap around to the beginning of the adapted range. i.e., you are guaranteed to see the end iterator of the adapted range before someone can decrement the begin iterator with defined behavior.

CaseyCarter:
>> I suppose we need to determine (a) if take_exactly is the only library component that assumes a weak range can be converted to a bounded range

> Probably not. I didn't know I was making a faulty assumption until late in the game. :-(

Glancing at the spec of `counted_iterator` while auditing requires expressions for conformance to the language I'm working on for specifying modifying/non-modifying expressions makes me nervous. If we want it to support cyclic iterators it probably needs a thorough review.

ericniebler:
@CaseyCarter says:
> It may be as simple as never comparing base iterators and only considering the count to be significant for ordering/difference.

That sounds promising.

ericniebler:
@gnzlbg says:
> `view::reverse` could find the end iterator the first time end is called for not bounded ranges, which would allow it to work with unbounded `take` and `take_exactly` views.

&lt;nod&gt;

ericniebler:
@CaseyCarter says:
> In the specific case of `view::cycle` you could cache the adapted range's end iterator since you technically don't need to know it until it has been seen.

Right.

> It may be worth the trouble to have a separate specialization of `cursor` to support `BidirectionalRange<Rng>() && !BoundedRange<Rng>()` that caches the end iterator.

Yes, I play games like this in the `stride` and `slice` views. The assumptions need to be checked there, too.

ericniebler:
Where are we with this? My understanding is that there were a few issues left to work out, so I'm holding off on merging this for now.

gnzlbg:
do not merge this (should i close it maybe?) As far as i understood view
reverse still need to be ugpraded and counted range/iterator need to be
upgraded too before a "view::cycle is possible". I can try to do this in
this PR but I think those two should be their own PR. We can reopen this
when those are done.

ericniebler:
Can either of you think of a random-access, *bounded* range that has cycles? Maybe the `take_exactly` optimization can be applied to those (and only those) kinds of ranges.

CaseyCarter:
If bounded ranges have a precondition that the "end" iterator is reachable from the "begin" - and they probably should - then a bounded range of ForwardIterators cannot have cycles. Since ForwardIterator inherits equality preserving `++` from Incrementable, cycles are "inescapable" and would violate the reachability precondition.

ericniebler:
> If bounded ranges have a precondition that the "end" iterator is reachable from the "begin" - and they probably should - then a bounded range of ForwardIterators cannot contain a cycle.

I think the definition of "range" (lowercase 'r') requires reachability. It's not obvious to me that it should. I can write a bounded"

ericniebler:
(Sorry, fat fingers, small device. Will finish that thought later)

gnzlbg:
> random-access, bounded range that has cycles?

No. A random access bounded range with cycles only works if the end of the range (one before the end actually) does not participate in the cycles. The moment the end participates in a cycle iteration exits too early. 

If a general implementation of such a range has to consider the case in which the range is part of the cycle, then it cannot be bounded anymore. 

So the only examples I was able to construct are for special cases of these general ranges. The simplest example of what I mean was already discussed above:

```c++
auto rng = view::iota(0, 4) | view::cycle | view::take(N);
```

which does not have a cycle involving the end for `N < 4` and thus is random-access, and bounded, for `N < 4`. Arguably nothing gets cycled for `N < 4`, but in this spirit we can consider the more contrived:

```c++
auto rng = view::iota(0, 4) | view::cycle_slice_n(/*begin:*/ 1, /*end:*/ 3, /*n:*/2);
// 0, 1, 2, 1, 2, 3
```

where the "end" of the `view::iota(0, 4)` does not take part in the cycle. This example has cycles, is random access, and could be bounded _if the end of the range is not allowed to be part of the slice_. Nobody would put such a restriction in a `cycle_slice_n` view.

ericniebler:
See 8ce7f6f. `view::repeat` is now random-access, `view::reverse` works with non-Bounded ranges, and `view::take_exactly` no longer assumes iterators are non-cyclic.

gnzlbg:
Thanks!

gnzlbg:
I have squashed the commits and all tests pass.

What should be done with input and infinite ranges?

Right now, the implementation of cycle requires a model of the ForwardRange and the BoundedRange concept, basically leaving those out. Since the view could be made conditionally single pass, this can be relaxed. Any thoughts?

ericniebler:
I don't think it's possible in general to cycle an Input range. Not supporting Bounded is overconstraining, and @CaseyCarter suggested one approach. But I'm OK with this being a separate pull. Let me take a quick look at the state of the code and get back to you.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_CYCLE_HPP
+#define RANGES_V3_VIEW_CYCLE_HPP
+#include <iostream>
```
ericniebler: Don't include `<iostream>`

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+            struct cursor
+            {
+            private:
+                template <typename T>
+                using constify_if = meta::apply<meta::add_const_if_c<IsConst>, T>;
+                using cycled_view_t = constify_if<cycled_view>;
+            public:
+                using difference_type = range_difference_t<Rng>;
+                using reference = range_reference_t<constify_if<Rng>>;
+                using single_pass = SinglePass<Rng>;
```
ericniebler: Delete this. `Rng` is not single-pass; you asserted that above.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+                void next()
+                {
+                    const auto end = ranges::end(rng_->rng_);
+                    if (it_ != end)
+                    {
+                        ++it_;
+                    }
+                    if (it_ == end)
+                    {
+                        it_ = begin(rng_->rng_);
```
ericniebler: I prefer that you qualify this call.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+                {
+                    return rng_ == pos.rng_ && it_ == pos.it_;
+                }
+
+                CONCEPT_REQUIRES(BidirectionalRange<Rng>{})
+                void prev()
+                {
+                    if (it_ == begin(rng_->rng_))
+                    {
+                        it_ = end(rng_->rng_);
```
gnzlbg: Here I'll use `ranges::begin` and `ranges::end` too.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+                    }
+                    if (it_ == end)
+                    {
+                        it_ = begin(rng_->rng_);
+                    }
+                }
+
+                bool equal(cursor const &pos) const
+                {
+                    return rng_ == pos.rng_ && it_ == pos.it_;
```
ericniebler: It's a bug if `rng_ != pos.rng_`. Turn that into an assertion.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+                }
+
+                CONCEPT_REQUIRES(BidirectionalRange<Rng>{})
+                void prev()
+                {
+                    if (it_ == begin(rng_->rng_))
+                    {
+                        it_ = end(rng_->rng_);
+                    }
```
ericniebler: Qualify these calls.

include/range/v3/view/cycle.hpp
```diff
@@ -0,0 +1,206 @@
...
+                    {
+                        return;
+                    }
+
+                    auto pos = it_ - begin + n;
+                    if (pos < 0 || pos >= N) {
+                      pos %= N;
+                      if (pos < 0) {
+                        pos += N;
+                      }
```
ericniebler: Indentation, brace-style here.

ericniebler:
I made a few extra clean-ups and merged to master. See c46dac7.

gnzlbg:
Thanks, looks way better now!

ericniebler:
f8c747f makes `view::cycle` work with non-Bounded ranges, using the technique @CaseyCarter describes above.

tvaneerd-rim:
My general impression is that there are two distinct properties of an iterator - where it is in the iteration (ie position/count), and what it actually points to.

I think (gut instinct)* equality should be concerned with position.  So the 10th iteration and the 13th are different, even if they point to the same object.

[*] its count is a salient attribute of the iterator.

ericniebler:
As appealing as that is from a logical perspective, it is impossible to realize without overhead. Consider the `repeat` view which has iterators that are empty apart from the repeated value (or a pointer to same) and for which `op++` is a no-op. To avoid cycles, you have add an int and make `op++` do work.  That also places an arbitrary bound on the number of times such an iterator can be incremented.

tvaneerd-rim:
I also wonder if this is a sign that we are still leaning too heavily on iterators instead of ranges.

ie which properties should be part of the range vs the iterator.  Maybe the range could be _slightly_ more heavy than the iterator, or we should ask the range more questions than the iterator.

Obviously high-level general comments.  Something doesn't feel right.
I need to find time to dig into the details more :-(

ericniebler:
I don't see how you can "fix" this by only making the range heavier but not the iterator. Are you suggesting we abandon iterators as the low-level primitives? If you're *not* suggesting that, then the `repeat` view must have heavy iterators like I described above.

tvaneerd-rim:
I haven't dug deep enough to make such a bold suggestion.  But it did cross my mind.  Or somehow not "dipping down" to the iterator level as soon/often.  But I need to live in the code a while before I could make a strong opinion.

Maybe repeat is just an edge case, but I find edge cases can _sometimes_ really shine light on whether you have the right model or not. Sometimes.

PR #184, opened 2015-07-06
fix missing includes forever

### LINKED ISSUE #182: Consider testing includes
Boost.Hana's CMake automatically tests that each header file is able to include itself by creating a test for each header file that just:

```c++
#include <header_file>
int main() { return 0; }
```

Is there interest in adding something like this to range-v3?

gnzlbg:
All kudos go to Louis Dione for writing the CMake scripts of Boost.Hana. 

This just adapts his scripts for range-v3; all mistakes introduced are only mine.

- fix missing includes caught by the tests.

- replace `# warning` with `#pragma message` because `-Werror, -Wpedantic` complains that 
  `# warning` is a language extension (this prevents some new tests from compiling).

Closes #182  .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,8 @@
+# Copyright Eric Niebler 2013-2015
+# Copyright Louis Dione 2015
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
 cmake_minimum_required(VERSION 2.8)
 set(CMAKE_LEGACY_CYGWIN_WIN32 0)
 
@@ -55,7 +60,6 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
 #   message(FATAL_ERROR "Unknown compiler.")
 endif()
 
-
 add_subdirectory(doc)
 
 if(BIICODE)
@@ -71,6 +75,33 @@ if(BIICODE)
 
     return()
 endif()
+
+#   range_v3_list_remove_glob(<list> <GLOB|GLOB_RECURSE> [globbing expressions]...)
+#
+# Generates a list of files matching the given glob expressions, and remove
+# the matched elements from the given <list>.
+#
+# Adapted from Boost.Hana: https://github.com/ldionne/hana/
+#
+macro(range_v3_list_remove_glob list glob)
+    file(${glob} _bhlrg10321023_avoid_macro_clash_matches ${ARGN})
+    list(REMOVE_ITEM ${list} ${_bhlrg10321023_avoid_macro_clash_matches})
+endmacro()
+
+# Test all headers
+include(CheckIncludeFileCXX)
+file(GLOB_RECURSE RANGE_V3_PUBLIC_HEADERS "${CMAKE_SOURCE_DIR}/include/*.hpp")
+range_v3_list_remove_glob(RANGE_V3_PUBLIC_HEADERS GLOB_RECURSE
+  "${CMAKE_SOURCE_DIR}/include/range/v3/detail/re_enable_warnings.hpp"
+  )
+
+foreach(_header IN LISTS RANGE_V3_PUBLIC_HEADERS)
+  file(RELATIVE_PATH _relative "${CMAKE_SOURCE_DIR}/include" "${_header}")
+  check_include_file_cxx("${_relative}" ${_relative}_FOUND "-I/${CMAKE_SOURCE_DIR}/include")
+  if(NOT ${_relative}_FOUND)
+    message(FATAL_ERROR "Compilation of header ${_relative} failed")
+  endif()
+endforeach()
 
 add_subdirectory(test)
 add_subdirectory(example)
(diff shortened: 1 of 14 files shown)


rhalbersma:
There is a CMake macro `check_include_file_cxx` that does this. Just `glob_recurse` to get all headers and feed them to this macro one by one.

rhalbersma:
Another thing: why not write another `include<header>` to also test for header guards?

ericniebler:
Can you please squash these two commits?

ericniebler:
> There is a CMake macro `check_include_file_cxx` that does this. Just `glob_recurse` to get all headers and feed them to this macro one by one.

Sounds like a worthwhile simplification to me.

gnzlbg:
> Can you please squash these two commits?

Done. 

> There is a CMake macro check_include_file_cxx that does this. Just glob_recurse to get all headers and feed them to this macro one by one.

I tried to look up what that macro does but the docs just say:

> Check if the include file exists.

Which is not the same, but the [source file](https://github.com/Kitware/CMake/blob/master/Modules/CheckIncludeFileCXX.cmake) uses `try_compile` to try and compile a file. 

It is not clear to me if it tries to compile the file by passing it `-I/path/to/include/file.hpp` or if it actually adds `#include <path/to/include/file.hpp>` to the source file.

ilyapopov:
Actually, that variable is used, but implicitly. In [CheckIncludeFileCXX.cmake](https://github.com/Kitware/CMake/blob/master/Modules/CheckIncludeFileCXX.cmake) on line 55 there is a call to `configure_file`, which takes [this template](https://github.com/Kitware/CMake/blob/master/Modules/CheckIncludeFile.cxx.in) and substitutes the variable `CHECK_INCLUDE_FILE_VAR` from the current scope. Then the resulting source file is passed to `try_compile` for compilation. Thus this macro does exactly  what we want.

gnzlbg:
@ilyapopov thanks for the explanation!

ericniebler:
@gnzlbg Will you be updating the patch with the CMake built-in?

gnzlbg:
@ericniebler Yes, don't merge this yet.

gnzlbg:
So I've got `check_include_file_cxx` to work. 

I don't know if I like this solution better yet, but it is simpler than the previous one.

What makes me doubt is that:
- one has to re-run cmake for this to work and it only works at configure time, 
- it takes a long time to compile all the headers and AFAIK this can't be parallelized.

Louis solution can be parallelized by just doing `make test.headers -j N`, and it can be run multiple times (`make test.headers`) but the list of files is only updated at configure time (so it isn't perfect either).

Anyhow I just wanted the missing includes problem fixed forever, and this achieves it. If it turns out to be too annoying or something we could introduce a cmake variable to enable/disable this in the future.

ericniebler:
Thanks!

ericniebler:
I'm now getting this error when I configure:

fatal: Not a git repository (or any parent up to mount point /cygdrive/c)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
-- Looking for C++ include meta/meta.hpp
-- Looking for C++ include meta/meta.hpp - not found
CMake Error at CMakeLists.txt:102 (message):
  Compilation of header meta/meta.hpp failed


-- Configuring incomplete, errors occurred!

ericniebler:
Reverted.

gnzlbg:
meta.hpp not found? Is this in linux, mac or windows?

ericniebler:
Cygwin on windows

gnzlbg:
And I assume meta is in the include/ directory of range-v3 right?

Typically CMake writes an error file in the build_dir/CMakeFiles/CMakeTmp/
. Those files might be useful to help figure out what goes wrong.

ericniebler:
That directory doesn't exist. Probably since I reverted the commit and reconfigured so I could work on something else.

gnzlbg:
The work around would be to add a cmake flag for it, and check the headers only if it's set. 

The cmake file uses the "portable" `check_include_file_cxx` way, maybe the original solution didn't have that problem. 

From that error message the only thing I get is that cmake was not able to either find meta.hpp or compile it. 

One could set up appveyor for range-v3, it's like travis, but with windows, and has mingw.

tamaskenez:
There's a surplus slash after the `-I` switch in the `check_include_file_cxx` command:

    check_include_file_cxx("${_relative}" ${_relative}_FOUND "-I/${CMAKE_SOURCE_DIR}/include")

If I remove it, it works again.
Another, maybe more elegant way to write it would be:

    set(CMAKE_REQUIRED_INCLUDES ${CMAKE_SOURCE_DIR}/include)
    ....
    check_include_file_cxx("${_relative}" ${_relative}_FOUND)

Let me know if I should send a PR with the fix.

gnzlbg:
Thanks @tamaskenez , 

I am giving this a try here: https://travis-ci.org/gnzlbg/range-v3/builds/7070363 . 

Can anybody check if that branch works on windows under Cygwin as is?

tamaskenez:
[a3c9527e](https://github.com/gnzlbg/range-v3/commit/a3c9527e83e505c1863d24ac952095fee494b042) works under Cygwin, cmake 3.1.2, gcc 4.9.2, includes found, all tests pass

gnzlbg:
Superseeded by #192 .

PR #187, opened 2015-07-10
value-initialized view::concat iterators must compare equal.

CaseyCarter:
ForwardIterator requires that value-initialized iterators compare equal, but value-initialized `tagged_variant`s are incomparable.

### DIFF
--- a/include/range/v3/view/concat.hpp
+++ b/include/range/v3/view/concat.hpp
@@ -238,7 +238,8 @@ namespace ranges
                 }
                 bool equal(cursor const &pos) const
                 {
-                    return its_ == pos.its_;
+                    return its_.is_valid() == pos.its_.is_valid() &&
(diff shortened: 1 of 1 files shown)


ericniebler:
I don't have the code in front of me, but why can't value initialized variants be made comparable? Or can they?

CaseyCarter:
`tagged_variant` default constructs to the empty state (`which() == (size_t)-1`) which is not valid for equality comparison; `operator==` asserts that the variant contains one of the alternatives (something like `which() <= sizeof...(Ts)`). We could obviously change `tagged_variant` so that empty variants compare equal if you prefer that design.

ericniebler:
Oh. I should implement the Lenexa variant and default construct to the first alternative.

CaseyCarter:
Superseded by #188.

PR #188, opened 2015-07-10
tagged_variant default constructs the first alternative if possible.

CaseyCarter:
`tagged_variant` constructors and `set` are now variadic. Note that the variadic implementation of `construct_fun` is an egregious hack with `const_cast` since the visitation implementation wants visitors to be immutable, but I want to perfect forward the constructor arguments.

Drive-by: reduce redundant code by using a common implementation for `tagged_variant` move/copy constructor/assignment, and another for detail::variant_data copy/move.
Drive-by: Remove unnecessary `operator!` from optional.

### DIFF
--- a/include/range/v3/utility/variant.hpp
+++ b/include/range/v3/utility/variant.hpp
@@ -44,7 +44,8 @@ namespace ranges
         struct tagged_variant_element;
 
         template<std::size_t N, typename Var>
-        using tagged_variant_element_t = typename tagged_variant_element<N, Var>::type;
+        using tagged_variant_element_t =
+            meta::_t<tagged_variant_element<N, Var>>;
 
         /// \cond
         namespace detail
@@ -67,11 +68,9 @@ namespace ranges
             template<>
             union variant_data<>
             {
-                [[noreturn]] void move(std::size_t, variant_data &&) const
-                {
-                    RANGES_ENSURE(false);
-                }
-                [[noreturn]] void copy(std::size_t, variant_data const &) const
+                template <typename That,
+                    CONCEPT_REQUIRES_(Same<variant_data,uncvref_t<That>>())>
+                [[noreturn]] void move_copy_construct(std::size_t, That &&) const
                 {
                     RANGES_ENSURE(false);
                 }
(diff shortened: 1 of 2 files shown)


include/range/v3/utility/variant.hpp
```diff
@@ -197,20 +190,30 @@ namespace ranges
...
+                  : t_{std::forward<Ts>(ts)...}
                 {}
                 template<typename U,
-                    meta::if_<std::is_constructible<U, T>, int> = 0>
+                    CONCEPT_REQUIRES_(Constructible<U, Ts...>())>
                 void operator()(U &u) const
                 {
-                    ::new((void*)std::addressof(u)) U(std::forward<T>(t_));
+                    // HACKHACKHACK: "workaround" the fact that the visitation
+                    // design does not allow for mutable visitors.
```
ericniebler: Would be nice to revisit this later.

ericniebler:
Thanks for doing this! It's nice to see variant getting some love.

Future work (for me or anybody so inclined): make this variant implementation as close to the Lenexa design as possible, since I really liked it (although few others seem to). However, I'd like to preserve the fact that variant::visit returns a variant.

gnzlbg:
I like the Lenexa variant too. Is there an implementation of it somewhere?

The only issue I have with it is that one cannot receive variants from the network/file I/O "as efficiently as it could get". But then neither does `std::vector` so after sleeping a week over it and discussing a lot about it on reddit I consider this issue to be minor.  

Those who need this will use `boost::container::vector` and some other variant type (similar but not equal to the Lenexa variant) and be done with the issue. 

If making those who need this happy complicates the life of everybody else then it is not worth it.

ericniebler:
> I like the Lenexa variant too. Is there an implementation of it somewhere?

I'm not aware of one. A motivated person would have an easy time adapting either mine or [Egg.Variant](https://github.com/eggs-cpp/variant).

CaseyCarter:
> A motivated person would have an easy time adapting either mine or Egg.Variant.

There's some great design documentation for egg.variant in [K-Ballo's blog](http://talesofcpp.fusionfenix.com/) that may be useful.

PR #196, opened 2015-07-14
bugfix: group_by and split broken for mutable views

gnzlbg:
Some views mutate its state on begin/end (e.g. `view::remove_if` on begin).  
This patch constrains `view::group_by` and `view::split` `begin_cursor` on `Range<const Rng>()`
to disable the const overload for these kinds of ranges (such that the non-const overload
is picked up instead).


ericniebler:
Great, thanks!

PR #199, opened 2015-07-21
Make repeat_n_view constructor constexpr.

CaseyCarter:
Perform the assert in the mem-initializer for n_ instead of the constructor body.


CaseyCarter:
To save you the trouble of looking: the Travis error is the zip view issue inherited from master.

gnzlbg:
Also: without this fix the code does not compile due to the missing semicolon.

gnzlbg:
Thanks!

PR #201, opened 2015-07-21
bugfix: missing semicolon

gnzlbg:


gnzlbg:
Closing this because #199 is a better solution (this one is constexpr in C++14 only while @CaseyCarter solution is also constexpr in C++11).

PR #202, opened 2015-07-22
Update core concepts to spec as of the current Ranges proposal:

CaseyCarter:
* Use `std::declval` instead of `ranges::concepts::val` in `Swappable`.
* Variadic `Same` (This is an extension; I'm sure we'll put it in the spec eventually.)
* Rename `Derived` to `DerivedFrom`.

  Drive-by: Replace several instances of tests for "single pass"-ness:
  ```
Derived<input_iterator_tag, Foo>
```
  with uses of `SinglePass` so that an input iterator tagged with
 ```
struct my_input_iterator_tag : input_iterator_tag {}
```
  is appropriately single-pass.

* `Convertible` changes:
  * Rename `Convertible` to `ImplicitlyConvertibleTo`
  * Add `ExplicitlyConvertibleTo`
  * Add `ConvertibleTo` that subsumes both `ImplicitlyConvertibleTo` and `ExplicitlyConvertibleTo`.
  This is half extension as well, but almost certainly required to specify the library. I don't think generic code should have to jump through hoops to require explicit & implicit conversion with equivalent semantics. (Most occurrences of `Convertible` in range-v3 actually mis-constrain explicit conversions.)

* Drive-by: Rename `Eraseable` to `Erasable`.

* Drive-by: Correct uses of `SinglePass<Range>` to `SinglePass<range_iterator_t<Range>>` (the second commit)

* Drive-by: Replace occurrences of `BoundedRange<Range> && !SinglePass<range_iterator_t<Range>>` with `BoundedRange<Range>` alone (the third commit). This one needs review to ensure I am not crazy.

### DIFF
--- a/include/range/v3/utility/concepts.hpp
+++ b/include/range/v3/utility/concepts.hpp
@@ -249,14 +249,22 @@ namespace ranges
 
             struct Same
             {
-                template<typename T, typename U>
-                auto requires_(T &&, U &&) -> decltype(
+                template<typename ...Ts>
+                struct same : std::true_type {};
+                template<typename T, typename ...Us>
+                struct same<T, Us...> : meta::fast_and<std::is_same<T, Us>...> {};
+                template<typename ...Ts>
+                using same_t = meta::_t<same<Ts...>>;
+
+                template<typename ...Ts>
+                auto requires_(Ts &&...) -> decltype(
                     concepts::valid_expr(
-                        concepts::is_true(std::is_same<T, U>{})
+                        concepts::is_true(same_t<Ts...>{})
                     ));
             };
 
-            struct Convertible
+            /// \cond
+            struct ImplicitlyConvertibleTo
             {
                 template<typename T, typename U>
                 auto requires_(T &&, U &&) -> decltype(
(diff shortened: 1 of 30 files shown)


include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -668,6 +668,9 @@ namespace ranges
 
         template<typename I, typename S = I>
         using sized_iterator_range_concept_t = meta::_t<sized_iterator_range_concept<I, S>>;
+
+        template<typename C>
+        using is_single_pass = meta::bool_<!DerivedFrom<C, ranges::forward_iterator_tag>()>;
```
ericniebler: What does this add beyond `SinglePass`?

include/range/v3/utility/concepts.hpp
```diff
@@ -249,14 +249,22 @@ namespace ranges
 
             struct Same
             {
-                template<typename T, typename U>
-                auto requires_(T &&, U &&) -> decltype(
+                template<typename ...Ts>
+                struct same : std::true_type {};
+                template<typename T, typename ...Us>
+                struct same<T, Us...> : meta::fast_and<std::is_same<T, Us>...> {};
```
gnzlbg: Thanks, I've needed this a couple of times and it was weird to use `Same` for two types and `meta` for the variadic case.

include/range/v3/utility/iterator_concepts.hpp
CaseyCarter: Only the ability to be used on the category alone, which is used by `postfix_increment_result` in basic_iterator.hpp. I didn't realize that `SinglePass` existed. I'll merge them.

ericniebler:
Great, thanks @CaseyCarter

PR #212, opened 2015-08-28
Code coverage with CI

kpp:
As I promised you at #cppsiberia I have integrated coveralls.io and CI.
Live example: https://coveralls.io/builds/3428943

You have to add hidden env variable named COVERALLS_IO_TOKEN into Travis


ericniebler:
According to Coveralls support, "Github doesn't have a read-only private repos OAuth scope." I can't authorize Coveralls to access my github repos, so unfortunately I can't accept this pull request. I greatly appreciate the effort you put into this, though.

PR #213, opened 2015-09-09
counted_iterator-related changes:

CaseyCarter:
* Implement `make_counted_iterator` from the ranges proposal
* Remove the remaining vestiges of `recounted(Iterator, Iterator)` that was removed from `counted_iterator` to support cyclic iterators in ce2d726e.
* Remove the `search` and `search_n` implementations for sized ranges of `RandomAccessIterator`. Without `recounted(I, I)`, they are no more efficient than the general sized range implementation.


ericniebler:
Thanks!

PR #219, opened 2015-10-01
Feature/support apple clang3.5

julian-becker:


.travis.yml
```diff
@@ -221,7 +229,7 @@ install:
   - if [ -n "$CLANG_VERSION" ]; then CXX_FLAGS="${CXX_FLAGS} -D__extern_always_inline=inline"; fi
   - if [ "$LIBCXX" == "On" ]; then CXX_FLAGS="${CXX_FLAGS} -stdlib=libc++ -I/usr/include/c++/v1/"; fi
   - if [ "$LIBCXX" == "On" ]; then CXX_LINKER_FLAGS="${CXX_FLAGS} -L/usr/lib/ -lc++"; fi
-  - cmake .. -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_EXE_LINKER_FLAGS="${CXX_LINKER_FLAGS}" -DRANGES_CXX_STD=$CPP -DMEMORYCHECK_COMMAND:FILEPATH=/usr/bin/valgrind
+  - cmake .. -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_EXE_LINKER_FLAGS="${CXX_LINKER_FLAGS}" -DRANGES_CXX_STD=$CPP
```
ericniebler: Why are we dropping valgrind support here?

julian-becker: valgrind support is not being dropped by this change (see build log [here](https://travis-ci.org/ericniebler/range-v3/jobs/83205405#L5236) starting at line 5236).
Removing the hard-coded path from the command line lets cmake choose the correct path for valgrind on both ubuntu and OS X. Otherwise, the os x build would fail, because there is no 'valgrind' executable at '/usr/bin/valgrind'. Instead it is located at '/usr/local/bin/valgrind' for os x.

ericniebler: Thanks for the explanation.

gnzlbg: I think the hardcoded valgrind paths were an artifact from before using
travis addons for packages.

julian-becker: I'm not very familiar with travis myself. Just experimented a little to make it work on OS X and this setting seemed to work.

PR #220, opened 2015-10-02
Update to current meta; update compiler support matrix in README.

### LINKED ISSUE #210: Fails to compile under C++1y mode of GCC 4.9.2 with libstdc++ version 20150327 (libstdc++.so.6.0.20)
I create a new file `test.cpp` in the root directory of a fresh range-v3 clone that only contains the following:

~~~~~.cpp
#include <range/v3/core.hpp>

~~~~~

This compiles cleanly with any of:

    clang++ -std=c++11 -stdlib=libc++ -isystem include test.cpp -c
    clang++ -std=c++1y -stdlib=libc++ -isystem include test.cpp -c
    g++     -std=c++11                -isystem include test.cpp -c

...but fails with...

    g++     -std=c++1y                -isystem include test.cpp -c

~~~~~
In file included from include/range/v3/core.hpp:17:0,
                 from test.cpp:1:
include/range/v3/begin_end.hpp:40:24: error: ‘std::rbegin’ has not been declared
             using std::rbegin;
                        ^
include/range/v3/begin_end.hpp:41:24: error: ‘std::rend’ has not been declared
             using std::rend;
                        ^
In file included from include/range/v3/view_facade.hpp:24:0,
                 from include/range/v3/core.hpp:21,
                 from test.cpp:1:
include/range/v3/utility/basic_iterator.hpp:374:22: error: ‘constexpr const T& ranges::v3::basic_mixin<Cur>::get() const’ cannot be overloaded
             T const &get() const noexcept
                      ^
include/range/v3/utility/basic_iterator.hpp:368:16: error: with ‘constexpr T& ranges::v3::basic_mixin<Cur>::get() const’
             T &get() noexcept
                ^
include/range/v3/utility/basic_iterator.hpp:396:22: error: ‘constexpr const S& ranges::v3::basic_sentinel<S>::end() const’ cannot be overloaded
             S const &end() const noexcept
                      ^
include/range/v3/utility/basic_iterator.hpp:391:16: error: with ‘constexpr S& ranges::v3::basic_sentinel<S>::end() const’
             S &end() noexcept
                ^
include/range/v3/utility/basic_iterator.hpp:440:47: error: ‘constexpr const Cur& ranges::v3::basic_iterator<Cur, S>::pos() const’ cannot be overloaded
             RANGES_CXX14_CONSTEXPR Cur const &pos() const noexcept
                                               ^
include/range/v3/utility/basic_iterator.hpp:436:41: error: with ‘constexpr Cur& ranges::v3::basic_iterator<Cur, S>::pos() const’
             RANGES_CXX14_CONSTEXPR Cur &pos() noexcept
                                         ^
~~~~~

Context
-------

 * ***Compiler*** : GCC Ubuntu 4.9.2-10ubuntu13
 * ***Stdlib*** : GNU libstdc++ version 20150327 (/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.20)
 * ***OS*** : Ubuntu 15.04
 * ***Kernel*** : 3.19.0-25-generic #26-Ubuntu SMP Fri Jul 24 21:17:31 UTC 2015 x86_64

CaseyCarter:
Disable constexpr_core since it's ill-formed. E.g., `ranges::begin` isn't `constexpr`, this test cannot possibly complete with -std=c++14. I assume the test is abandoned.

Also disables the flags in test/CMakeLists.txt to force static linking and debug info.

Fixes #210.

### DIFF
--- a/include/meta/meta_fwd.hpp
+++ b/include/meta/meta_fwd.hpp
@@ -15,6 +15,8 @@
 #ifndef META_FWD_HPP
 #define META_FWD_HPP
 
+#include <utility>
+
 #ifndef META_DISABLE_DEPRECATED_WARNINGS
 #ifdef __cpp_attribute_deprecated
 #define META_DEPRECATED(MSG) [[deprecated(MSG)]]
@@ -35,8 +37,12 @@ namespace meta
 {
     inline namespace v1
     {
+#ifdef __cpp_lib_integer_sequence
+        using std::integer_sequence;
+#else
         template <typename T, T...>
         struct integer_sequence;
+#endif
 
         template <typename... Ts>
         struct list;
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Supported Compilers
 The code is known to work on the following compilers:
 
 - clang 3.4.0 (C++14 support requires at least clang 3.5.0)
-- GCC 4.9.0
+- GCC 4.9.0 (C++14 support requires GCC 5.2. Complex `constexpr` usage may be unreliable due to bugs in GCC, e.g. [PR67813](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67813), as witnessed by the failure of the constexpr-core test.)
 
 **Development Status:** This code is fairly stable, well-tested, and suitable for casual use, although currently lacking documentation. No promise is made about support or long-term stability. This code *will* evolve without regard to backwards compatibility.
 
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -18,6 +18,7 @@
 #include <cstddef>
 #include <initializer_list>
 #include <type_traits>
+#include <utility>
 #include <meta/meta_fwd.hpp>
 
 #if defined(__clang__)
@@ -190,12 +191,12 @@ namespace meta
         /// An integral constant wrapper around the result of incrementing the wrapped integer \c
         /// T::type::value.
         template <typename T>
-        using inc = std::integral_constant<decltype(T::type::value), T::type::value + 1>;
+        using inc = std::integral_constant<decltype(T::type::value + 1), T::type::value + 1>;
 
         /// An integral constant wrapper around the result of decrementing the wrapped integer \c
         /// T::type::value.
         template <typename T>
-        using dec = std::integral_constant<decltype(T::type::value), T::type::value - 1>;
+        using dec = std::integral_constant<decltype(T::type::value - 1), T::type::value - 1>;
 
         /// An integral constant wrapper around the result of adding the two wrapped integers
         /// \c T::type::value and \c U::type::value.
@@ -215,7 +216,7 @@ namespace meta
         /// \c T::type::value and \c U::type::value.
         /// \ingroup math
         template <typename T, typename U>
-        using multiplies = std::integral_constant<decltype(T::type::value *U::type::value),
+        using multiplies = std::integral_constant<decltype(T::type::value * U::type::value),
                                                   T::type::value * U::type::value>;
 
         /// An integral constant wrapper around the result of dividing the two wrapped integers \c
@@ -225,8 +226,8 @@ namespace meta
         using divides = std::integral_constant<decltype(T::type::value / U::type::value),
                                                T::type::value / U::type::value>;
 
-        /// An integral constant wrapper around the remainder of dividing the two wrapped integers
-        /// \c T::type::value and \c U::type::value.
+        /// An integral constant wrapper around the result of negating the wrapped integer
+        /// \c T::type::value.
         /// \ingroup math
         template <typename T>
         using negate = std::integral_constant<decltype(-T::type::value), -T::type::value>;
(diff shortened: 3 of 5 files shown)


CaseyCarter:
I notably have NOT turned on Travis compiles for GCC5. For unknown reasons, [GCC5 debug builds fail on Travis,](https://travis-ci.org/CaseyCarter/range-v3/builds/83379896), despite that they work just fine for me with both the g++5 installed by my package manager (which claims to be "Ubuntu 5.1.0-0ubuntu11~14.04.1"), and hand-built from `^/tags/gcc_5_1_0_release`.

CaseyCarter:
> For unknown reasons

"Unknown reasons" are probably related to Travis using a 12.04 "precise" Ubuntu base, and me using 14.04 "trusty."

CaseyCarter:
> `ranges::begin` isn't `constexpr`

This is a mis-characterization. The `operator()` members of `ranges::begin` are in fact `constexpr`. The issue is that `ranges::begin(foo)` delegates to `begin(foo)` which in many cases - including some used in `constexpr-core` - is not `constexpr`. E.g., in `test_array`:

```c++
RANGES_CXX14_CONSTEXPR auto test_array() -> bool {
    test::array<int, 4> a{{1, 2, 3, 4}};

    auto beg = ranges::begin(a);
    // ...
}
```
`ranges::begin(a)` unconditionally invokes the non-`constexpr` function `std::begin`, so the program is ill-formed.

gnzlbg:
On Sat, Oct 3, 2015 at 5:19 AM, Casey Carter <notifications@github.com>
wrote:

> ranges::begin(a) unconditionally invokes the non-constexpr function
> std::begin, so the program is ill-formed.


Why? begin_fn::operator() delegates to impl which has two overloads. IIUC
the intent is that if the type does not implement a member function begin,
then ADL will be performed (which includes std::begin). However for test
array, a member begin is present, so impl(..., int) should be called. All
of these are constexpr and this snipped used to work in clang 3.5, 3.6 and
3.7.

So why is ranges::begin calling std::begin instead of test::array::begin?

There is a proposal (P0031R0) to improve std::array, such that test::array
won't be necessary in the future.

Just because std::array and most std:: types are not constexpr friendly
doesn't mean range-v3 shouldn't be. The constexpr core PR was just a first
step towards a constexpr enabled range-v3. There is a closed PR that makes
all the algorithms constexpr and adds constexpr tests for all of them, as
well as an older PR that also made most of the views constexpr.

CaseyCarter:
> Why?

Sorry, I misread the error. The problem is in `test_c_array`:
```
cd ~/dev/range-v3/test && clang++-3.7 -I ~/dev/range-v3/include -std=c++14 -stdlib=libc++ ./constexpr_core.cpp
In file included from ./constexpr_core.cpp:8:
./test_iterators.hpp:331:59: warning: returning reference to local temporary object
      [-Wreturn-stack-address]
    reference operator[](difference_type n) const {return it_[n];}
                                                          ^~~~~~
**[snip]**
./constexpr_core.cpp:246:19: error: static_assert expression is not an integral constant
      expression
    static_assert(test_c_array(), "");
                  ^~~~~~~~~~~~~~
/home/casey/dev/range-v3/include/range/v3/begin_end.hpp:122:28: note: non-constexpr
      function 'begin<int, 4>' cannot be used in a constant expression
                    return begin(static_cast<Rng &&>(rng));
                           ^
/home/casey/dev/range-v3/include/range/v3/begin_end.hpp:138:28: note: in call to
      'impl(a, 0)'
                    return begin_fn::impl(static_cast<Rng &&>(rng), 0);
                           ^
./constexpr_core.cpp:101:19: note: in call to '&value->operator()(a)'
    auto beg = It{ranges::begin(a)};
                  ^
./constexpr_core.cpp:117:16: note: in call to 'test_it(a)'
    return     test_it<input_iterator<int *>>(a)
               ^
./constexpr_core.cpp:208:10: note: in call to 'test_its(a)'
    if (!test_its(a)) { return false; }
         ^
./constexpr_core.cpp:246:19: note: in call to 'test_c_array()'
    static_assert(test_c_array(), "");
                  ^
/usr/include/c++/v1/iterator:1391:1: note: declared here
begin(_Tp (&__array)[_Np])
^
```
It would appear that the array overload of `std::begin` is not `constexpr` in my installed libc++ (`_LIBCPP_VERSION 1101`). That's right, **version 1101 from llvm 3.4**. Apparently the Ubuntu clang packages don't include libc++. I'll fix the pull.

include/meta/meta_fwd.hpp
```diff
@@ -35,8 +37,12 @@ namespace meta
 {
     inline namespace v1
     {
+#ifdef __cpp_lib_integer_sequence
+        using std::integer_sequence;
+#else
         template <typename T, T...>
         struct integer_sequence;
+#endif
```
ericniebler: Why are we special-casing for libc++ here? Wouldn't we need to do the same thing for libstdc++?

CaseyCarter: libc++ macros are generally `LIBCPP_XXX`; `__cpp_lib_XXX` is the format for SD-6 feature test macros for library features. `__cpp_lib_integer_sequence` came from [N4440](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4440.html#recs.cpp14).

ericniebler: Thanks for the clarification.

PR #233, opened 2015-11-02
Implement algorithms min, max, and minmax from the Ranges TS.

CaseyCarter:
Fixes # 232.

### DIFF
--- a/include/range/v3/algorithm.hpp
+++ b/include/range/v3/algorithm.hpp
@@ -45,10 +45,13 @@
 #include <range/v3/algorithm/is_sorted_until.hpp>
 #include <range/v3/algorithm/lexicographical_compare.hpp>
 #include <range/v3/algorithm/lower_bound.hpp>
+#include <range/v3/algorithm/max.hpp>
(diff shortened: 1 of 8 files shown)


gnzlbg:
Shouldn't these have overloads for revalue references (that either are
deleted or do the right thing)?

CaseyCarter:
> Shouldn't these have overloads for rvalue references (that either are
deleted or do the right thing)?

Yes, they probably should. There was a flurry of discussion a few weeks ago on one of the reflectors about fixing `min` and `max` that didn't really go anywhere. The proper solution would probably be to accept the parameters as forwarding references, and return something like `remove_rvalue_reference_t<common_reference_t<T&&, U&&>>`, e.g.:
```c++
template <class T, class U, class P = identity,
  IndirectCallableRelation<projected<const decay_t<T>*, P>,
    projected<const decay_t<U>*, P>> C = less<>>
remove_rvalue_reference_t<common_reference_t<T&&, U&&>>
min(T&& t, U&& u, C comp = C{}, P proj = P{}) {
  // handwaving
}
```
Feel free to investigate ;)

PR #236, opened 2015-11-08
Correct critical error in is_permutation

CaseyCarter:
I just found this terrible error in cmcstl2, and realizing that `is_permutation` came from range-v3 I figured it should be fixed here as well.


ericniebler:
LOL

PR #237, opened 2015-11-08
Make GCC 4.9 happy with variant

CaseyCarter:
G++ 4.9 doesn't like:
```c++
template<typename T, typename U>
auto operator()(T &&t, U &&) const ->
    decltype(fn()(std::forward<T>(t)))
{
    return fn()(std::forward<T>(t));
}
```
In the new variant code when `fn` is a non-static-member function. This patch changes that syntax in several places to:
```c++
template<typename T, typename U>
auto operator()(T &&t, U &&) const
RANGES_DECLTYPE_AUTO_RETURN
(
    this->fn()(std::forward<T>(t))
)
```
which 4.9 strangely enough *does* accept.


ericniebler:
Huh, strange. But thanks.

Issue #235, opened 2015-11-06
Mom, Travis-CI is hating on me again!


ericniebler:
https://travis-ci.org/ericniebler/range-v3/builds/89329436

CaseyCarter:
[Found the bug](https://github.com/ericniebler/range-v3/pull/237).

Incidentally, I strongly dislike the header-checking feature of CMakeLists.txt. It greatly slows down the speed of changing build configuration, which is a mild annoyance, and it makes it impossible to debug header file issues on Travis.

ericniebler:
Agreed

gnzlbg:
This annoyed me infinitely too.

I still prefer Boost.Hana solution, which basically just add a target `make
headers` that you have to explicitly call to build all the headers.

PR #238, opened 2015-11-08
Disable the header-checking feature in CMakeLists.txt on Travis

CaseyCarter:
...so that we can get actual error messages.


CaseyCarter:
Maybe fixes #235.

PR #241, opened 2015-11-11
To container fix

### LINKED ISSUE #240: Cannot copy ForwardRange of variant into a container
I'm not sure if this is a bug or if it's a lack of my understanding but I've been stumped for a while now and can't see where I'm going wrong (and can't make sense of the error) so I figured it may be worth mentioning in case it is actually a bug..

Given a variant of two primitives (int and bool in this case), I can transform an arbitrary random access range into a range of these variants and then copy them into a `std::vector` using `to_` and it works, but if I try the same with a forward range I get errors.

```
#include <range/v3/all.hpp>

#include <boost/variant.hpp>

#include <string>
#include <vector>

int main()
{
  using my_variant = boost::variant<int, bool>;
  std::string is("abcd");

  // this works fine:
  is | ranges::view::transform([](auto) -> my_variant { return 1; })
     | ranges::to_<std::vector<my_variant>>();

  // so does this:
  is | ranges::view::transform([](auto) -> int { return 1; })
     | ranges::to_<std::vector<int>>();
  // and this:
  is | ranges::view::group_by([](auto, auto) { return true; })
     | ranges::view::transform([](auto) -> int { return 1; })
     | ranges::to_<std::vector<int>>();

  // but this does not:
  is | ranges::view::group_by([](auto, auto) { return true; })
     | ranges::view::transform([](auto) -> my_variant { return 1; })
     | ranges::to_<std::vector<my_variant>>();
}
```

Running with the clang in xcode 7.1 with the `std=c++1z` flag.

CaseyCarter:
Fixes #240 by not using uniform initialization in `to_container` when initializing a container from an iterator range.

### DIFF
--- a/include/range/v3/to_container.hpp
+++ b/include/range/v3/to_container.hpp
@@ -63,7 +63,7 @@ namespace ranges
                 Cont impl(Rng && rng, std::false_type) const
                 {
                     using I = range_common_iterator_t<Rng>;
-                    return Cont{I{begin(rng)}, I{end(rng)}};
+                    return Cont(I{begin(rng)}, I{end(rng)});
                 }
 
                 template<typename Rng,


ericniebler:
C++ sucks. That is all.

CaseyCarter:
I absolutely love uniform initialization. But I hate `std::initializer_list` with the white-hot burning passion of a thousand suns.

PR #242, opened 2015-11-11
Protect basic_mixin::get to avoid inadvertent access

CaseyCarter:
`basic_iterator` currently exposes its cursor though the public member function `get` in `basic_mixin`. Given that `get` is an incredibly common function name, there is a fair likelihood that someone will mistakenly use `i.get()` when they meant `i->get()` or `(*i).get()`. This patch makes `basic_mixin::get` a `protected` member, and officially blesses `get_cursor` as the approved way to break encapsulation and retrieve a cursor from a `basic_iterator`.


ericniebler:
Thanks

PR #245, opened 2015-11-12
Default partial sum

lichray:


ericniebler:
IIRC, plus<> is C++14. range-v3 should be portable to C++11.

lichray:
If you don't mind the C++11 and C++14 behaviors diverge a little bit, I updated the change.

ericniebler:
Thanks!

PR #250, opened 2015-11-25
Use an empty variant for visitation

### LINKED ISSUE #249: Using `concat` over a zipped range
```
    vector<int> iv1 = {1,2,3};
    vector<int> iv2 = {1,2,3};

    vector<float> fv1 = {4,5,6};
    vector<float> fv2 = {4,5,6};

    vector<double> dv1 = {7,8,9};
    vector<double> dv2 = {7,8,9};

    auto r1 = ranges::view::zip(ranges::view::all(iv1),ranges::view::all(fv1),ranges::view::all(dv1));
    auto r2 = ranges::view::zip(ranges::view::all(iv2),ranges::view::all(fv2),ranges::view::all(dv2));
    auto r3 = ranges::view::concat(r1, r2);
    for(auto i: r3){
      //..
    }
```
Would this be possible? 

I mean I could probably concat first and then zip but this would be much nicer to use for my current problem.

CaseyCarter:
Drive-by: shut up deprecation warnings in alg.random_shuffle.

Fixes #249.

### DIFF
--- a/include/range/v3/utility/variant.hpp
+++ b/include/range/v3/utility/variant.hpp
@@ -179,6 +179,8 @@ namespace ranges
                 }
             };
 
+            struct empty_variant_tag { };
+
             struct variant_core_access
             {
                 template<typename...Ts>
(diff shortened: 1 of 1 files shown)


CaseyCarter:
*sigh* pulling the random_shuffle warning suppression, cmake 2.8 doesn't support `target_compile_definitions`.

PR #257, opened 2015-11-29
Two features in one pull! 10% more awesome than competing brands.

CaseyCarter:
This set of commits provides two closely-related changes:
* Divorces "sizedness" from iterator category in `basic_iterator`. Difference is not linked to category in the iterator model; `basic_iterator` should be able to assist in the creation of sized iterator types that are not `RandomAccess`.
* Renames `default_sentinel` to `default_end_cursor` and adds a type alias `default_sentinel` for `basic_iterator<default_end_cursor>`. Also adds the few operators necessary for `default_sentinel` to conform to the TS.

`counted_iterator` uses the machinery added to `basic_iterator` removing the need for a special-purpose `counted_sentinel`. `default_sentinel` works with `counted_iterator` exactly as specified in the TS.

Small drive-by changes:
* Split `view_facade::size` into separate implementations for fixed-cardinality and variable-cardinality views so that fixed-cardinality views need not have `const` variants of `begin` and `end` for `size` to compile.
* Simplify concept `SizedIteratorRange` by refining `SizedIteratorRangeLike_`.
* `basic_mixin<T>` uses a `box<T>` for storage so that stateless iterators/sentinels are empty.


ericniebler:
You seem to be using the term "end cursor" and "sentinel cursor" interchangeably in the code. I don't care which, but this should be consistent. And there are many places in the code where these things are simply called "sentinels". Once we settle on a term we should clean this up.

CaseyCarter:
The naming of `SizedCursorSentinel` is because it is parallel to `SizedSentinel` in the TS design, it didn't occur to me that the name implies the end-cursor-thing is a sentinel. Would you agree that `SizedCursorRange` is more appropriate?

ericniebler:
Thanks. @Beman, this would be a good change to pick up if you're proposing `basic_iterator`.

PR #260, opened 2015-12-02
libstdc++<5.1 doesn't provide std::is_final

gnzlbg:
Commit 6bddf650d9b53197bb82a504e8038a160b56174c uses `std::is_final` if C++>11, which breaks e.g. when clang is used with libstdc++-4.9 in C++14 mode.

This commit also checks if the std library implementation is libstdc++, and in that case, if its version is >= 5.1, which implements std::is_final according to their docs (haven't tested this though).

### DIFF
--- a/include/range/v3/utility/box.hpp
+++ b/include/range/v3/utility/box.hpp
@@ -130,7 +130,7 @@ namespace ranges
 
         template<typename Element, typename Tag = Element,
             bool Empty = std::is_empty<Element>::value &&
-#ifdef RANGES_CXX_GREATER_THAN_11
+#ifdef __cpp_lib_is_final
                          !std::is_final<Element>::value>
 #else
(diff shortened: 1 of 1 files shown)


gnzlbg:
Yes that would be cleaner, I just thought that for this single point of usage this wasn't worth it. I will update the PR.

I also thought about suggesting the deprecation of libstdc++4.9 but I still need to use it in some systems... and this also reminded me of all the `std::is_trivially_x` traits that in libstdc++4-9 are called `std::has_trivial_x`...

gnzlbg:
I misunderstood your comment above, I thought you meant defining a macro to shorten this out. I've updated the PR.

ericniebler:
Squash please!

gnzlbg:
sorry for the delay, done!

PR #264, opened 2015-12-11
Update iterator concepts per ericniebler/stl2#153:

### LINKED ISSUE #231: Compilation failure because GCC _Safe_iterator meets SizedIteratorRange concept
I'm getting a compilation failure that I think stems from a (GCC `_GLIBCXX_DEBUG`) bidirectional iterator meeting the `SizedIteratorRange` concept and then failing when range-v3 actually uses an expression like `itr_a - itr_b`.

I'm guessing the troublesome subtlety may be: the iterator is actually a GCC `_Safe_iterator`, which *does* provide an `operator-()` but which just implements this by passing through to the wrapped `std::list` iterator that *doesn't* provide an `operator-()`. Hence I'm guessing the `concepts::same_type(s - i, i - s)` line in the `SizedIteratorRange` concept *is* checking that the iterator's `operator-()` exists and has a valid return type but *isn't* checking that a call to that expression is actually valid. Hence `SizedIteratorRange` isn't spotting that `itr_a - itr_b` isn't a valid expression and so doesn't get itself SFINAE-rejected as it should.

Some illustrative code:

~~~~~.cpp
#define _GLIBCXX_DEBUG
#define RANGES_NO_STD_FORWARD_DECLARATIONS

#include <iostream>
#include <list>
#include <type_traits>

#include <range/v3/all.hpp>

int main() {
	// Background context:
	// Under _GLIBCXX_DEBUG, list returns a list iterator wrapped in a _Safe_iterator
	// whose iterator_category is bidirectional_iterator_tag
	using my_range_type  = std::list< int >;
	using my_iter_type   = decltype( std::begin( std::declval< my_range_type >() ) );
	static_assert(
		std::is_same<
			std::iterator_traits<my_iter_type>::iterator_category,
			std::bidirectional_iterator_tag
		>::value,
		""
	);

	// Background context:
	// range-v3 identifies the _Safe_iterator<...> as SizedIteratorRange
	using rv3_iterator_t = ranges::range_iterator_t< my_range_type >;
	using rv3_sentinel_t = ranges::range_sentinel_t< my_range_type >;
	using rv3_sirc_t     = ranges::sized_iterator_range_concept<rv3_iterator_t,
	                                                            rv3_sentinel_t>;
	ranges::concepts::SizedIteratorRange * my_ptr = rv3_sirc_t();

	//  **** PROBLEM HERE *****
	// Fails to compile the '...::take( 3 );' line.
	// I think that's because range-v3 treats _Safe_iterator<...> as a
	// SizedIteratorRange iterator but then cannot subtract one from another
	// (because that's not supported by the underlying list iterator)
	const my_range_type my_values = { 1, 2, 3, 4 };
	const auto my_first_values = my_values | ranges::v3::view::take( 3 );
	std::cerr << *ranges::max_element( my_first_values ) << "\n";

	return 0;
}
~~~~~

Compile command:

    g++ -std=c++11 -isystem range-v3/include -ftemplate-backtrace-limit=0 rv3_safe_iter_prob.cpp

(also works with `-std=c++14`; `-ftemplate-backtrace-limit=0` avoids clipped messages)

CaseyCarter:
* Rename `MergeMovable` to `MoveMergeable`.
* Implement `WeaklyEqualityComparable`.
* Relax `IteratorRange` to agree with the Ranges TS `Sentinel` concept.
* `SinglePass` is now true for `OutputIterator`s.
* Coalesce:
  *  `{Input,Output,""}Iterator` with `Weak{Input,Output,""}Iterator`
  * `WeaklyComparable` with `Comparable`
  * `Mismatchable1` with `Mismatchable2`
  * `AsymmetricallyComparable` with `WeaklyAsymmetricallyComparable`

* Replace `is_sized_range`/`SizedRangeLike_` with `disable_sized_range`, which is a
  boolean UnaryTypeTrait instead of a variable template as in the TS.
* Implement `disable_sized_iterator_range<I, S>`, also a boolean UnaryTypeTrait,
  to serve the function of the TS's `disable_sized_sentinel<S, I>`. Closes #231.

* Update cursor concepts in `range_access`, and usages in `basic_iterator`.
* Remove relational operators from `basic_sentinel`, and the cross-type relational
  operators from `basic_iterator`.

* Bring `advance` and `distance` into spec, adjust `enumerate` to agree.
* Constrain `prev`.
* Don't use `sized_iterator_range_concept_t<I, I>` in `iter_enumerate`: `<I, I>` may
  not satisfy `IteratorRange`, let alone `SizedIteratorRange`.

* Implement `common_cursor::equal` overload for non-EC wrapped iterators.

* Remove `unreachable`'s same-type `==`/`!=` and cross-type `<`, etc.
* Remove all Iterator/Sentinel `common_type` specializations.

* `any_cursor` treats all non-equality-comparable (weak) iterators as equal.

### DIFF
--- a/include/range/v3/utility/iterator.hpp
+++ b/include/range/v3/utility/iterator.hpp
@@ -92,17 +92,22 @@ namespace ranges
                 RANGES_CXX14_CONSTEXPR
                 static D bounded_(I &it, D n, S bound, concepts::IteratorRange*,
                     concepts::InputIterator*);
+                template<typename I, typename D>
+                RANGES_CXX14_CONSTEXPR
+                static D bounded_(I &it, D n, I bound, concepts::IteratorRange*,
+                    concepts::BidirectionalIterator*);
                 template<typename I, typename D, typename S>
                 RANGES_CXX14_CONSTEXPR
-                static D bounded_(I &it, D n, S bound, concepts::IteratorRange*,
-                    concepts::BidirectionalIterator*);
-                template<typename I, typename D, typename S, typename Concept>
+                static D bounded_(I &it, D n, S bound, concepts::SizedIteratorRange*,
+                    concepts::InputIterator*);
+                template<typename I, typename D>
                 RANGES_CXX14_CONSTEXPR
-                static D bounded_(I &it, D n, S bound, concepts::SizedIteratorRange*, Concept);
+                static D bounded_(I &it, D n, I bound, concepts::SizedIteratorRange*,
+                    concepts::BidirectionalIterator*);
             public:
                 // Advance a certain number of steps:
                 template<typename I,
-                    CONCEPT_REQUIRES_(WeakIterator<I>())>
+                    CONCEPT_REQUIRES_(Iterator<I>())>
                 RANGES_CXX14_CONSTEXPR
                 void operator()(I &i, iterator_difference_t<I> n) const
                 {
@@ -111,15 +116,15 @@ namespace ranges
                     advance(i, n);
                 }
                 // Advance to a certain position:
-                template<typename I,
-                    CONCEPT_REQUIRES_(WeakIterator<I>())>
+                template<typename I, typename S,
+                    CONCEPT_REQUIRES_(IteratorRange<I, S>() && Assignable<I&, S&&>())>
                 RANGES_CXX14_CONSTEXPR
-                void operator()(I &i, I s) const
+                void operator()(I &i, S s) const
                 {
                     i = std::move(s);
                 }
                 template<typename I, typename S,
-                    CONCEPT_REQUIRES_(IteratorRange<I, S>())>
+                    CONCEPT_REQUIRES_(IteratorRange<I, S>() && !Assignable<I&, S&&>())>
                 RANGES_CXX14_CONSTEXPR
                 void operator()(I &i, S s) const
                 {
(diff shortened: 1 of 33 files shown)


include/range/v3/utility/iterator.hpp
```diff
@@ -182,14 +189,32 @@ namespace ranges
...
+                RANGES_ASSERT(0 <= d);
+                if (d <= n)
+                {
+                    ranges::advance(it, std::move(bound));
+                    return n - d;
+                }
+                ranges::advance(it, n);
+                return 0;
+            }
+            template<typename I, typename D>
```
ericniebler: I wonder why you had to add this overload.

include/range/v3/utility/iterator.hpp
```diff
@@ -389,8 +423,9 @@ namespace ranges
             RANGES_CXX14_CONSTEXPR
             iterator_size_t<I> operator()(I begin, S end) const
             {
-                RANGES_ASSERT(0 <= (end - begin));
-                return static_cast<iterator_size_t<I>>(end - begin);
+                auto n = static_cast<iterator_difference_t<I>>(end - begin);
```
ericniebler: Doesn't the expression `end - begin` already have type `iterator_difference_t<I>`?

ericniebler:
This is a big improvement, thanks.

include/range/v3/utility/iterator.hpp
CaseyCarter: This one and the next differed substantially in the first version of the patch, IIRC. The only real difference now is the type of `bound` and the `0 <= n` precondition on this overload. They could easily be merged into a single `I, D, S, SizedIteratorRange, Concept` overload with that precondition changed to `Same<I, S>() || 0 <= n`. That would also give me a chance to fix the space after `if` and the annoying fact that the first overload uses `d <= n` vs. the seconds `n >= d`.

include/range/v3/utility/iterator.hpp
CaseyCarter: Yes; it didn't a month ago. I'll clean this up as well.

PR #267, opened 2015-12-13
fix iterator type in iter_transform2_view::cursor constructor

tivek:
Fix iter_transform2_view::cursor constructor to take range_iterator_t<Rng1> and range_iterator_t<Rng2> instead of twice range_iterator_t<Rng2>.


ericniebler:
Thanks.

tivek:
Sure, thanks for the quick reply.

Issue #280, opened 2016-02-10
Lazy set algorithms?


tivek:
I need to compute set unions and intersections on ranges which may be infinite. To scratch that itch I used `view_facade` to write lazy versions of `set_intersection`, `set_difference`, `set_symmetric_difference`, and `set_union`.

A rough first version is [here](https://github.com/tivek/range-v3/blob/lazy-set-algorithm/include/range/v3/view/set_algorithm.hpp) (still need to write some tests...)

I would welcome a short look and comments to improve this code.

Here are some things I am mulling over:
- Any unobvious tricky situations to test for besides checking the common type/reference when doing union and symmetric difference?
- Resulting references/values of `set_intersection` and `set_difference` come from the first operand. `view_adaptor` would make for a simpler implementation.
- All these fat cursors are repetitive.

If lazy set operations sound useful to you I am willing to shape up the code and send a pull request.

ericniebler:
This looks amazing! Are there tests?

tivek:
In the following days I hope to have presentable tests for the most important properties: agreement with greedy algorithms, value and ref types, cardinalities.

tivek:
Tests are here: [test/view/set_*.cpp](https://github.com/tivek/range-v3/tree/lazy-set-algorithm/test/view).

There's some unexpected behavior in set_difference and set_symmetric_difference (both lazy and greedy algorithms are affected): if input ranges have identical elements, the first of the same elements does not end up in the result.

ericniebler:
Great, thanks. I am on a vacation this week with my family so I can't look right now.

tivek:
No rush, enjoy the vacation!

tonyelewis:
Just a quick note of support. For some time I've been thinking how great lazy set algorithms would be. I think this work could be really useful. Thanks!

tivek:
Thanks for the support. I originally stumbled upon range-v3 as a way to make efficient lazy inner joins between two or more ranges of different types. To get my feet wet with the library (and get into C++14!) I tried to implement lazy set algorithms first.

I still have a lot to learn and tinker with this code - Travis [complains](https://travis-ci.org/tivek/range-v3/jobs/111094066) with Clang 3.7 and GCC 4.9 even though GCC 5.3.0 is pretty happy with the view.set_* tests. So I'm grateful for any hints. And if the lazy set algorithms are useful to anyone else, well I see that as a bonus. :)

ericniebler:
[...]

PR #288, opened 2016-02-28
implement lazy set algorithms as range views

tivek:
Implement lazy set algorithms as range views, related to issue #280.

### DIFF
--- a/include/range/v3/view/set_algorithm.hpp
+++ b/include/range/v3/view/set_algorithm.hpp
@@ -1,0 +1,904 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2014
+//  Copyright Tomislav Ivek 2015-2016
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_SET_ALGORITHM_HPP
+#define RANGES_V3_VIEW_SET_ALGORITHM_HPP
+
+#include <utility>
+#include <iterator>
+#include <algorithm>
+#include <functional>
+#include <type_traits>
+#include <meta/meta.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/size.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/view_facade.hpp>
+#include <range/v3/utility/move.hpp>
+#include <range/v3/utility/semiregular.hpp>
+#include <range/v3/utility/functional.hpp>
+#include <range/v3/utility/static_const.hpp>
+#include <range/v3/view/view.hpp>
+#include <range/v3/view/all.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        /// \cond
+        namespace detail
+        {
+            // indirect_move is put here instead of in set_??_view::cursor to
+            // work around gcc friend name injection bug.
+            template<typename Cursor>
+            struct set_op_cursor_move
+            {
+                template<typename Sent>
+                friend auto indirect_move(basic_iterator<Cursor, Sent> const &it)
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    get_cursor(it).indirect_move_()
+                )
+            };
+            
+            template<typename Rng1, typename Rng2,
+                     typename C, typename P1, typename P2,
+                     template<bool, typename...> class Cursor, cardinality Cardinality>
+            struct set_algorithm_view
+            : view_facade<
+                    set_algorithm_view<Rng1, Rng2, C, P1, P2, Cursor, Cardinality>,
+                    Cardinality>
+            {
+            private:
+                friend range_access;
+                semiregular_t<function_type<C>> pred_;
+                semiregular_t<function_type<P1>> proj1_;
+                semiregular_t<function_type<P2>> proj2_;
+                Rng1 rng1_;
+                Rng2 rng2_;
+
+                template<bool IsConst>
+                using cursor = Cursor<IsConst, Rng1, Rng2, C, P1, P2>;
+
+                cursor<false> begin_cursor()
+                {
+                    return {pred_, proj1_, proj2_,
+                            ranges::begin(rng1_), ranges::end(rng1_),
+                            ranges::begin(rng2_), ranges::end(rng2_)};
+                }
+                CONCEPT_REQUIRES(Range<Rng1 const>() && Range<Rng2 const>())
+                cursor<true> begin_cursor() const
+                {
+                    return {pred_, proj1_, proj2_,
+                            ranges::begin(rng1_), ranges::end(rng1_),
+                            ranges::begin(rng2_), ranges::end(rng2_)};
+                }
+            public:
+                set_algorithm_view() = default;
+                set_algorithm_view(Rng1 rng1, Rng2 rng2,
+                                   C pred, P1 proj1, P2 proj2)
+                : pred_(as_function(std::move(pred)))
+                , proj1_(as_function(std::move(proj1)))
+                , proj2_(as_function(std::move(proj2)))
+                , rng1_(std::move(rng1))
+                , rng2_(std::move(rng2))
+                {}
+            };
+
+            
+            template<bool IsConst,
+                     typename Rng1, typename Rng2,
+                     typename C, typename P1, typename P2>
+            struct set_difference_cursor
+                : detail::set_op_cursor_move<set_difference_cursor<IsConst, Rng1, Rng2, C, P1, P2>>
+            {
+            private:
+                using pred_ref_ = semiregular_ref_or_val_t<function_type<C>, IsConst>;
+                using proj1_ref_ = semiregular_ref_or_val_t<function_type<P1>, IsConst>;
+                using proj2_ref_ = semiregular_ref_or_val_t<function_type<P2>, IsConst>;
+                pred_ref_ pred_;
+                proj1_ref_ proj1_;
+                proj2_ref_ proj2_;
+                
+                template<typename T>
+                using constify_if = meta::invoke<meta::add_const_if_c<IsConst>, T>;
+                
+                using R1 = constify_if<Rng1>;
+                using R2 = constify_if<Rng2>;
+                
+                range_iterator_t<R1> it1_;
+                range_sentinel_t<R1> end1_;
+                
+                range_iterator_t<R2> it2_;
+                range_sentinel_t<R2> end2_;
+            
+                void satisfy()
+                {
+                    while(it1_ != end1_)
+                    {
+                        if(it2_ == end2_)
+                            return;
+
+                        if(pred_(proj1_(*it1_), proj2_(*it2_)))
+                            return;
+                        
+                        if(!pred_(proj2_(*it2_), proj1_(*it1_)))
+                            ++it1_;
+                        
+                        ++it2_;
+                    }
+                }
+                
+            public:
+                using value_type = range_value_t<constify_if<Rng1>>;
+                using single_pass = meta::strict_or<SinglePass<range_iterator_t<R1>>,
+                                                    SinglePass<range_iterator_t<R2>>>;
+                
+                set_difference_cursor() = default;
+                set_difference_cursor(pred_ref_ pred, proj1_ref_ proj1, proj2_ref_ proj2,
+                                      range_iterator_t<R1> it1, range_sentinel_t<R1> end1,
+                                      range_iterator_t<R2> it2, range_sentinel_t<R2> end2)
+                  : pred_(std::move(pred)), proj1_(std::move(proj1)), proj2_(std::move(proj2)),
+                    it1_(std::move(it1)), end1_(std::move(end1)), it2_(std::move(it2)), end2_(std::move(end2))
+                {
+                    satisfy();
+                }
+                auto get() const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    *it1_
+                )
+                void next()
+                {
+                    ++it1_;
+                    satisfy();
+                }
+                bool equal(set_difference_cursor const &that) const
+                {
+                    return it1_ == that.it1_; // does not support comparing iterators from different ranges
+                }
+                bool done() const
+                {
+                    return it1_ == end1_;
+                }
+                auto indirect_move_() const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    iter_move(it1_)
+                )
+            };
+        
+            constexpr cardinality set_difference_cardinality(cardinality c1, cardinality c2)
+            {
+                return (c1 == unknown) ? unknown :
+                    (c1 >= 0) || (c1 == finite) ? finite : // else, c1 == infinite
+                        (c2 >= 0) || (c2 == finite) ? infinite : unknown;
+            }
+            
+        }
+        /// \endcond
+
+        
+        template<typename Rng1, typename Rng2,
+                 typename C, typename P1, typename P2>
+        using set_difference_view = detail::set_algorithm_view<Rng1, Rng2, C, P1, P2,
+                 detail::set_difference_cursor,
+                 detail::set_difference_cardinality(
+                    range_cardinality<Rng1>::value,
+                    range_cardinality<Rng2>::value)>;
+        
+        namespace view
+        {
+            struct set_difference_fn
+            {
+            public:
+                template<typename Rng1, typename Rng2,
+                         typename C, typename P1, typename P2,
+                         typename I1 = range_iterator_t<Rng1>,
+                         typename I2 = range_iterator_t<Rng2>>
+                using Concept = meta::and_<
+                    InputRange<Rng1>, InputRange<Rng2>,
+                    IndirectCallableRelation<C, Projected<I1, P1>, Projected<I2, P2>>
+                >;
+                template<typename Rng1, typename Rng2,
+                    typename C = ordered_less, typename P1 = ident, typename P2 = ident,
+                    CONCEPT_REQUIRES_(Concept<Rng1, Rng2, C, P1, P2>())>
+                set_difference_view<all_t<Rng1>, all_t<Rng2>, C, P1, P2>
+                operator()(Rng1 && rng1, Rng2 && rng2,
+                    C pred = C{}, P1 proj1 = P1{}, P2 proj2 = P2{}) const
+                {
+                    return {all(std::forward<Rng1>(rng1)),
+                            all(std::forward<Rng2>(rng2)),
+                            std::move(pred),
+                            std::move(proj1),
+                            std::move(proj2)};
+                }
+
+
+            #ifndef RANGES_DOXYGEN_INVOKED
+                template<typename Rng1, typename Rng2,
+                    typename C, typename P1, typename P2,
+                    typename I1 = range_iterator_t<Rng1>,
+                    typename I2 = range_iterator_t<Rng2>,
+                    CONCEPT_REQUIRES_(!Concept<Rng1, Rng2, C, P1, P2>())>
+                void operator()(Rng1 && rng1, Rng2 && rng2,
+                    C pred, P1 proj1, P2 proj2) const
+                {
+                    CONCEPT_ASSERT_MSG(InputRange<Rng1>(),
+                        "The first parameter of view::set_difference "
+                        "must be a model of the InputRange concept.");
+                    CONCEPT_ASSERT_MSG(InputRange<Rng2>(),
+                        "The second parameter of view::set_difference "
+                        "must be a model of the InputRange concept.");
+                    CONCEPT_ASSERT_MSG(
+                        IndirectCallableRelation<C, Projected<I1, P1>, Projected<I2, P2>>(),
+                        "The predicate function passed to view::set_difference "
+                        "must be callable with two arguments of the two "
+                        "input ranges' value types.");
+                    CONCEPT_ASSERT_MSG(
+                        Callable<P1, range_value_t<Rng1>>(),
+                        "The first projection function passed to view::set_difference "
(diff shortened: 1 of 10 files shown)


ericniebler:
Could you investigate the failures here: https://travis-ci.org/ericniebler/range-v3/builds/112324136

tivek:
So far I see three different issues. The one I can more or less pin down has to do with move_into(back_inserter(container of move-only type)). Will first install compiler versions that fail. then fix what I can and report the rest in separate issues.

CaseyCarter:
The gcc-4.9 compile errors - which do not manifest under 5.3 - are corrected by:

<pre>
diff --git a/include/range/v3/utility/common_iterator.hpp b/include/range/v3/utility/common_iterator.hpp
index aa2007f..1929102 100644
--- a/include/range/v3/utility/common_iterator.hpp
+++ b/include/range/v3/utility/common_iterator.hpp
@@ -91,7 +91,7 @@ namespace ranges
                     RANGES_ASSERT(is_sentinel());
                     return ranges::get<1>(data_);
                 }
-                CONCEPT_REQUIRES(SizedIteratorRange<I, S>() && SizedIteratorRange<I, I>())
+                CONCEPT_REQUIRES((bool) SizedIteratorRange<I, S>() && (bool) SizedIteratorRange<I, I>())
                 friend iterator_difference_t<I>
                 operator-(common_iterator<I, S> const &end, common_iterator<I, S> const &begin)
                 {
</pre>

Given that the pull doesn't touch common_iterator.hpp, and the error doesn't manifest in a newer gcc, I suspect the problem here is a bug in GCC 4.9.

tivek:
As far as the OSX clang/libc++ failure goes, a similar problem has been reported before: http://stackoverflow.com/questions/28528210/is-function-object-returned-by-stdmem-fn-required-to-have-a-const-overload

In short, old versions of libc++ __mem_fn lack a const overload of the call operator. The fix was made some time ago, https://github.com/llvm-mirror/libcxx/tree/a4c0d87a84b5d324726b334d10fe2f8c24215fad , but it didn't make it into Xcode 6.1 provided by Travis CI.

A short test demonstrates the failure:
```C++
#include <range/v3/utility/functional.hpp>
#include "../simple_test.hpp"
#include "../test_utils.hpp"


struct S
{
    int val;
};

int main()
{
    S s{5};
    
    ranges::v3::function_type<int S::*> f{&S::val};
    CHECK(f(s) == 5);

    const ranges::v3::function_type<int S::*> cf{&S::val};
    CHECK(cf(s) == 5);

    return test_result();
}
```
and its build log can be seen here: https://travis-ci.org/tivek/range-v3/builds/113831866

I do not have access to an OSX machine to check this with a more recent Xcode, but the code above and the related projection tests compile just fine with a recent libc++ on Linux.

Suggestions? Should I see how to fence off the projection tests for older versions of libc++?

CaseyCarter:
The Travis error compiling with clang and C++14 seems to be a `basic_iterator` design issue. Incidentally, it is reproducible with g++6; I have no idea why the bug does not manifest with either compiler when compiling for C++11.

`basic_iterator`'s constructor that accepts a cursor believes that it can construct a mixin from an rvalue cursor. This requirement is not enforced by the `Cursor` requirements, so we didn't notice that the range-v3 cursors for insert iterators don't satisfy it until compiling your particular test program in C++14.

Similar problems exist for the cursor-converting constructor that wants to construct a `mixin<T>` from an rvalue `mixin<ConvertibleTo<T>>`, and the cursor-converting assignment operator.

The simple fix is to support the required operations in the insert iterators. To prevent reccurrence, we need to add explicit requirements to `basic_iterator` and/or cursors. I'm inclined to require:
```c++
Constructible<range_access::mixin_base_t<Cursor>, Cursor>() &&
Constructible<range_access::mixin_base_t<Cursor>, const Cursor&>()
```
of all cursors, and constrain the converting `basic_iterator` constructor/assignment operator with
```c++
ConvertibleTo<Other, Cursor>() &&
Constructible<range_access::mixin_base_t<Cursor>, Other>() && 
Constructible<range_access::mixin_base_t<Cursor>, const Other&>()
```
However, constraining in this fashion requires that such constructors be *publicly* accessible. Are we comfortable with a design that makes it impossible to keep constructors that accept cursors out of the public interface of the resulting `basic_iterator`? Thoughts, @ericniebler, @beman?

ericniebler:
> Are we comfortable with a design that makes it impossible to keep constructors that accept cursors out of the public interface of the resulting `basic_iterator`?

Couldn't `basic_iterator` inherit the constructors from the mixin, then define *and delete* its own constructors that take the cursor type?

CaseyCarter:
> Couldn't basic_iterator inherit the constructors from the mixin, then define and delete its own constructors that take the cursor type?


What happens to e.g.:
```c++
        template<typename Derived, cardinality Cardinality>
        struct view_facade
          : view_interface<Derived, Cardinality>
        {
            // ...
            template<typename D = Derived, CONCEPT_REQUIRES_(Same<D, Derived>())>
            detail::facade_iterator_t<D> begin()
            {
                // Typically constructs basic_iterator from cursor
                return {range_access::begin_cursor(derived(), 42)};
            }
        };
```
and
```c++
struct forward_list {
  struct cursor { /* ... */ };
  using iterator = basic_iterator<cursor>;
  iterator begin() { return iterator{cursor{...}}; }
};
```

Sometimes we *want* to convert a cursor / cursor-like-sentinel-thingy to `basic_iterator`/`basic_sentinel`. Perhaps we should define `private` constructors in `basic_iterator` and expose them via `range_access`.

ericniebler:
Personally, I am OK with leaving the constructors from cursor publicly accessible. Most cursors are defined as private nested classes of range views. Those that aren't can be defined in detail namespaces. I say, keep it simple. I have a commit  coming soon that fixes this, much like Casey's.

tivek:
Thank you, it seems 8099ffb fixes the last remaining build failure. I am now going to rebase on top of master and squash the commits.

ericniebler:
There's still the missing const on std::mem_fn on older Xcode releases. I'm working around that right now.

test/view/set_difference.cpp
```diff
@@ -0,0 +1,317 @@
...
+    S s_finite[] = {S{-20}, S{-10}, S{1}, S{3}, S{3}, S{6}, S{8}, S{20}};
+    
+    {
+    //     auto res1 = view::set_difference(s_finite, view::ints(-2, 10),
+    //                                      ordered_less(),
+    //                                      &S::val,
+    //                                      ident()
+    //                                     );
+        // not using &S::val avoids a problem with older libc++ where mem_fn()
+        // does not provide an operator()const
```
ericniebler: That should be fixed now. I think I've worked around the issue.

test/view/set_difference.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+        std::vector<MoveOnlyString> expected;
+        move(res, back_inserter(expected));
+
+        ::check_equal(expected, {"a","b","c","x"});
+        ::check_equal(v1, {"b","x","y","z"});
+        //::check_equal(v0, {"","","b","","","x"});
+        
+        // FIXME: why is v0 not equal to {"","","b","","","x"}?
+        // the lazy view::set_difference mimics the greedy set_difference very closely!
```
ericniebler: This needs to be investigated.

test/view/set_difference.cpp
```diff
@@ -0,0 +1,317 @@
...
+        CONCEPT_ASSERT(Same<range_value_t<R>, MoveOnlyString>());
+        CONCEPT_ASSERT(Same<range_reference_t<R>, MoveOnlyString &>());
+        CONCEPT_ASSERT(Same<range_rvalue_reference_t<R>, MoveOnlyString &&>());
+    }
+    
+    // WARNING: set_difference between two infinite ranges can create infinite loops!
+    // {
+    //     auto empty_range = view::set_difference(view::ints, view::ints);
+    //     begin(empty_range); // infinite loop!
+    // }
```
ericniebler: The `set_difference` algorithm can statically assert the finite-ness of the ranges.

test/view/set_intersection.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+    bool operator!=(MoveOnlyString const &that) const
+    {
+        return !(*this == that);
+    }
+    friend std::ostream & operator<< (std::ostream &sout, MoveOnlyString const &str)
+    {
+        return sout << '"' << str.sz_ << '"';
+    }
+};
```
ericniebler: Probably better to move `MoveOnlyString` into ../test_utils.hpp. Copy-n-paste is bad.

test/view/set_intersection.cpp
```diff
@@ -0,0 +1,275 @@
...
+    S s_finite[] = {S{-20}, S{-10}, S{1}, S{3}, S{3}, S{6}, S{8}, S{20}};
+    
+    {
+//         auto res1 = view::set_intersection(s_finite, view::ints(-2, 10),
+//                                            ordered_less(),
+//                                            &S::val,
+//                                            ident()
+//                                           );
+        // not using &S::val avoids a problem with older libc++ where mem_fn()
+        // does not provide an operator()const
```
ericniebler: Here too.

test/view/set_symmetric_difference.cpp
```diff
@@ -0,0 +1,350 @@
...
+    
+    // projections
+    {        
+//         auto res1 = view::set_symmetric_difference(b_finite, d_finite,
+//                                                    ordered_less(),
+//                                                    &B::val,
+//                                                    &D::val
+//                                                   );
+        // not using &S::val avoids a problem with older libc++ where mem_fn()
+        // does not provide an operator()const
```
ericniebler: And here.

test/view/set_symmetric_difference.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+        std::vector<MoveOnlyString> expected;
+        move(res, back_inserter(expected));
+
+        ::check_equal(expected, {"a","b","c","x","y","z"});
+        ::check_equal(v1, {"b","x","",""});
+        //::check_equal(v0, {"","","b","","","x"});
+        
+        // FIXME: why is v0 not equal to {"","","b","","","x"}?
+        // the lazy view::set_symmetric_difference mimics the greedy set_symmetric_difference very closely!
```
ericniebler: This needs to be investigated.

test/view/set_symmetric_difference.cpp
```diff
@@ -0,0 +1,350 @@
...
+        CONCEPT_ASSERT(Same<range_reference_t<R>, MoveOnlyString &>());
+        CONCEPT_ASSERT(Same<range_rvalue_reference_t<R>, MoveOnlyString &&>());
+    }
+
+
+    // WARNING: set_symmetric_difference between two infinite ranges can create infinite loops!
+    // {
+    //     auto empty_range = view::set_symmetric_difference(view::ints, view::ints);
+    //     begin(empty_range); // infinite loop!
+    // }
```
ericniebler: Add a static assert.

test/view/set_union.cpp
```diff
@@ -0,0 +1,340 @@
...
+    
+    // projections
+    {        
+//         auto res1 = view::set_union(b_finite, d_finite,
+//                                     ordered_less(),
+//                                     &B::val,
+//                                     &D::val
+//                                    );
+        // not using &S::val avoids a problem with older libc++ where mem_fn()
+        // does not provide an operator()const
```
ericniebler: And here.

test/view/set_difference.cpp
CaseyCarter: Running the algorithm in my head - shouldn't `v0` be `{"", "b", "", "", "x", ""}`?

test/view/set_symmetric_difference.cpp
CaseyCarter: `v0` should be `{"", "b", "", "", "x", ""}` here as well.

test/view/set_difference.cpp
ericniebler: I don't have time to think this through right now. I would be willing to bet that if libstdc++ and libc++ agree on the result of `std::set_difference`, we should have the same behavior here. Is that not what's happening?

test/view/set_difference.cpp
tivek: It can. That however does not avoid possibly infinite loops. There are no compile-time guarantees that the difference is not empty, in fact the issue is similar to starting a possibly never-ending `find_if()` on an infinite range.

test/view/set_difference.cpp
tivek: Here I was expressing my surprise that `set_difference` moves out the last occurences of the repeating element, as opposed to the `set_intersection`.

In hindsight, `set_difference` must find the first mismatch and can't go backwards to move out some previous instance of a repeating element. So I take back my surprised comments.

As far as I can tell `std::set_difference` only defines the number of times an equivalent element is copied out instead of which ones exactly.

Finally, `view::set_difference` and `view::set_symmetric_difference` closely emulate their eager counterpart algorithms. The lazy and eager results are verifiably the same regardless of my wrong and misinformed hunches on what an algorithm should or should not do.

tivek:
I am away on a trip this week and might be slow to respond. I am going to shape up the code according to the comments as soon as I get back to my pc.

test/view/set_intersection.cpp
tivek: Agree fully. Here I just wanted to have the first batch of tests done. Will clean it up.

test/view/set_difference.cpp
CaseyCarter: > As far as I can tell `std::set_difference` only defines the number of times an equivalent element is copied out instead of which ones exactly.

[set.difference]/5:

> *Remarks:* If `[first1,last1)` contains *m* elements that are equivalent to each other and `[first2,
last2)` contains *n* elements that are equivalent to them, the last *max(m−n, 0)* elements from `[first1,
last1)` shall be copied to the output range.

tivek: Thanks Casey, it's clear now :)

test/view/set_difference.cpp
tivek: One thing that comes to mind... `view::set_difference` and view::set_symmetric_difference` could test if the two infinite input ranges are the same. In that case we can just return a default cursor for which `cursor::done()` always returns `true`. This does require at least giving the cursor an extra `bool` member and an extra check in `done()`. I wonder if it is worth it just to protect against (what I guess is) an unlikely case of taking a difference between two same infinite sets.

tivek:
I went through the comments and made the changes. What's left open are the static asserts on infinite ranges in `set_difference` and `set_symmetric_difference`.

ericniebler:
I'm not concerned about the asserts. If you squash the commits, I'll merge this.

tivek:
Squashed. Thank you and Casey for the quick replies and a great library.

include/range/v3/view/set_algorithm.hpp
```diff
@@ -0,0 +1,904 @@
...
+                void next()
+                {
+                    ++it1_;
+                    satisfy();
+                }
+                bool equal(set_difference_cursor const &that) const
+                {
+                    return (it1_ == that.it1_ && it2_ == that.it2_) ||
+                           (done() && that.done());
+                }
```
CaseyCarter: Why is `set_difference_cursor::equal` not simply `return it1_ == that.it1_;`? Are you trying to support comparison of iterators into differing ranges?

include/range/v3/view/set_algorithm.hpp
```diff
@@ -0,0 +1,904 @@
...
+            {
+            private:
+                friend range_access;
+                semiregular_t<function_type<C>> pred_;
+                semiregular_t<function_type<P1>> proj1_;
+                semiregular_t<function_type<P2>> proj2_;
+                Rng1 rng1_;
+                Rng2 rng2_;
+
+                struct sentinel;
```
CaseyCarter: This `sentinel` class is never defined.

include/range/v3/view/set_algorithm.hpp
CaseyCarter: This cursor needs to be single pass if either underlying iterator is single pass. i.e.,
```c++
using single_pass = meta::or_<SinglePass<range_iterator_t<R1>>, SinglePass<range_iterator_t<R2>>>;
```

include/range/v3/view/set_algorithm.hpp
```diff
@@ -0,0 +1,904 @@
...
+                void next()
+                {
+                    ++it1_;
+                    ++it2_;
+                    satisfy();        
+                }
+                bool equal(set_intersection_cursor const &that) const
+                {
+                    return (it1_ == that.it1_ && it2_ == that.it2_) ||
+                           (done() && that.done());
```
CaseyCarter: Similar argument here: comparing `it1_` and `it2_` is sufficient if you only need to compare cursors that point into the same range.

include/range/v3/view/set_algorithm.hpp
```diff
@@ -0,0 +1,904 @@
...
+                            
+                        case state_t::ONLY_SECOND:
+                            ++it2_;
+                            break;
+                    }
+                }
+                bool equal(set_union_cursor const &that) const
+                {
+                    return (it1_ == that.it1_ && it2_ == that.it2_) ||
+                           (done() && that.done());
```
CaseyCarter: Checking `done` is extraneous here unless you want to support comparison of past-the-end iterators of differing sequences.

include/range/v3/view/set_algorithm.hpp
```diff
@@ -0,0 +1,904 @@
...
+                            break;
+                        case state_t::ONLY_SECOND:
+                            ++it2_;
+                            break;
+                    }
+                }
+                bool equal(set_symmetric_difference_cursor const &that) const
+                {
+                    return (it1_ == that.it1_ || it2_ == that.it2_) ||
+                           (done() && that.done());
```
CaseyCarter: Same comment for checking `done` in `equal`.

include/range/v3/view/set_algorithm.hpp
tivek: ... And it is not needed, the default sentinel is supposed to be used. Will remove it.

include/range/v3/view/set_algorithm.hpp
tivek: > Why is set_difference_cursor::equal not simply return it1_ == that.it1_;? Are you trying to support comparison of iterators into differing ranges?

Yes, I wanted to prevent a false positive in case the two `it2_`s belong to different ranges. Maybe it that is too defensive.

tivek: > This cursor needs to be single pass if either underlying iterator is single pass. i.e.,
> 
> using single_pass = meta::or_<SinglePass<range_iterator_t<R1>>, SinglePass<range_iterator_t<R2>>>;
> 
> I think the same consideration applies to the other cursors.

Good catch. I remember overthinking the whole thing and in the end it seems I did not set the `single_pass` trait as it should be.

include/range/v3/view/set_algorithm.hpp
tivek: How much handholding do we want here? Can we trust the user not to compare cursors pointing into different ranges?

include/range/v3/view/set_algorithm.hpp
CaseyCarter: In your defense, at least one of the later cursors has a correct `single_pass`. I think you just missed the earlier two.

include/range/v3/view/set_algorithm.hpp
tivek: > Checking `done` is extraneous here unless you want to support comparison of past-the-end iterators of differing sequences.
Exactly. I put it here in case `that` points into a subrange of one of the input ranges of `this`. Suppose `that` is already done but `this` is still not. `this` and `that` are not equal, but only comparing their member iterators could in principle say otherwise. Checking `done` covers this case.

include/range/v3/view/set_algorithm.hpp
CaseyCarter: > Can we trust the user not to compare cursors pointing into different ranges?

Iterators into different ranges are not required to be in the domain of equality (by the standard or by range-v3). An iterator *may* provide stronger guarantees - pointers do, for example - but in general users who compare iterators into different ranges get the undefined behavior they deserve.

Since this is C++, I suggest that you make the common case fast and make any defensive code debug-only or not-at-all.

include/range/v3/view/set_algorithm.hpp
tivek: (And of course, checking `it2_ == that.it2_` is part of the same being-correct-with-subranges story).

include/range/v3/view/set_algorithm.hpp
tivek: I understand, but please see comments a bit below about iterators into range and its subrange. Do we dole out UB in that case too? In the end, calling `equal()` on fat iterators of lazy algorithms is not a common thing to do, as opposed to `done()` which is as fast as it gets.

include/range/v3/view/set_algorithm.hpp
CaseyCarter: Putting on my semanticist hat: This implementation does *not* support comparing cursors into distinct ranges, because the resulting iterators fail to meet the requirements of `EqualityComparable`. To satisfy `EqualityComparable`, `==` has to mean substitutable. As a rule of thumb, any state that contributes to the behavior of an `EqualityComparable` type must either be accounted for by its equality operator, or objects that differ in the value of that state must be excluded from the domain of equality. (For readers familiar with the recent changes to the `Sentinel` concept in the TS, this argument was the cause.)

For the cursors defined here, significant state includes predicates, projection functions, and the beginnings and endings of the underlying ranges. Since the equality operators ignore predicates, projections, and the ends of the underlying ranges, cursors that don't agree in those attributes cannot be in the domain of equality. In other words, these operator definitions only support comparison of iterators into distinct ranges when those distinct ranges have the same predicate, projection functions, and ends-of-underlying ranges. If that is the case, then `this->done() && that.done()` implies `this->it1_ == that.it1_ && this->it2_ == that.it2_`; the `done` check is redundant.

tivek: OK, so I was aiming for an `EqualityComparable` iterator but missed by a mile. Agreed the "complete&correct" `equal()` is not needed for most use cases. I really appreciate the explanation. Fancy hats are cool.

ericniebler:
Thanks, guys. This is an exciting addition to the range library.

PR #295, opened 2016-03-08
basic_sentinel's constructors must be public.

CaseyCarter:
Corrects a cause of major breakage when __GNUC__ is not defined, including DevDiv 191705. Changing the trigger for the GCC PR 60799 workaround from `#ifndef __GNUC__` to `#if !defined(__GNUC__) || defined(__clang__)` disables the workaround when compiling with clang, which should prevent this issue from reccurring.

### DIFF
--- a/include/range/v3/utility/basic_iterator.hpp
+++ b/include/range/v3/utility/basic_iterator.hpp
@@ -445,7 +445,7 @@ namespace ranges
         struct basic_sentinel : range_access::mixin_base_t<S>
         {
             // http://gcc.gnu.org/bugzilla/show_bug.cgi?id=60799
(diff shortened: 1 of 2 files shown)


CaseyCarter:
Travis is puking all over builds today, timing out while trying to install valgrind. I'll force push this change atop itself to restart the build when I determine Travis is working again.

ericniebler:
Looks good, thanks!

CaseyCarter:
>  DevDiv 191705

I suppose I should clarify for posterity that DevDiv 191705 is the MS internal bug that corresponds to the repro you submitted for comprehensions.cpp and Clang/C2.

Issue #206, opened 2015-07-28
AddressSanitizer has been disabled in travis for a while...


gnzlbg:
TL;DR: 
 - travis ASan builds are not testing anything, so we should disable them
 - ASan complains about memory errors in a couple of tests (`action::insert/push_front/push_back`) where valgrind doesn't complain.

Long story short, today ASan panicked in an app using range-v3 due to some issue with `action::push_back`. I looked into it and learned the following:
- LLVM ubuntu builds come with ASan disabled since January
   - LLVM deprecated the auto-tools build system in ~January in favor of CMake. ASan doesn't work with auto-tools anymore and all LLVM ubuntu package build-bots still use the deprecated auto-tools system. 

- I ran ASan locally in Debug mode with the following flags and the following test "broke":

  > cmake .. -DCMAKE_CXX_FLAGS="-fsanitize=address,integer,undefined,leak -fno-omit-frame-pointer" -DCMAKE_BUILD_TYPE=Debug

    - `action::insert`
    - `action::push_front`
    - `action::push_back`

All the error messages are very similar so it seems to be the same issue showing up in `ranges::insert`. 

Can anybody reproduce this? 

Here the output:

# action::insert

```bash
=================================================================
==12222==ERROR: AddressSanitizer: container-overflow on address 0x60300000db50 at pc 0x00010fd53414 bp 0x7fff4ff386c0 sp 0x7fff4ff386b8
WRITE of size 4 at 0x60300000db50 thread T0
[...]

PR #296, opened 2016-03-10
Travis updates:

CaseyCarter:
* Enable GCC5 C++11 ASAN in the test matrix. (GCC doesn't support
  -fsanitize=integer)
* Fix typo in .travis.yaml that was effectively disabling sanitzers
  for ASAN builds
* **Intentionally** disable ASAN for clang; it's broken on Travis pending resolution
  of https://llvm.org/bugs/show_bug.cgi?id=22757.

Relates to #206.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,15 +22,18 @@ script: cmake
 # - clang 3.5, 3.6, trunk
 #   - Note: trunk is tested with/without ASAN,
 #     the rest are only tested with ASAN=On.
-# - gcc 4.9, 5.0
+# - gcc 4.9, 5
 #
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


gnzlbg:
Looks good. ASAN was indeed disabled due to that issue. IIRC correctly when that first broke clang would actually compile and link fine even though the ASAN library was missing. The binaries would just run without Asan. Somebody complained about Asan triggering in their machines, which was weird since the Asan builds were fine, and that is when it was discovered in the logs that clang had been producing Asan builds without Asan for about 3 months, not only here, but in all projects that were using Travis back then. More than one year later we still don't have Clang available with Asan on Travis...

PR #298, opened 2016-03-11
size: grab array types before std::size sees them to avoid ambiguity.

CaseyCarter:
Avoids ambiguous lookup for `size(a)` when `std::size(T(&)[N])` is defined and `std` is an associated namespace of `T`.


ericniebler:
:thumbsup:

PR #311, opened 2016-03-31
fix equal() for set_union_cursor and set_symmetric_difference_cursor

tivek:
In case of lazy set union and symmetric difference views, comparing iterators for equality must involve subiterators of both operand ranges. Fixes bug where it1_ has already reached the end (and is therefore not incremented anymore) but it2_ still provides new elements.


ericniebler:
Needs tests.

tivek:
You shouldn't have to say that :) The relevant tests are now there.

ericniebler:
Great, thanks.

PR #312, opened 2016-04-01
Add clang 3.8 to the test matrix.

CaseyCarter:
...with working sanitizers!

Specifically: adds clang 3.8 configs with C++11/14, Debug/Release, ASAN=On only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
       os: osx
       compiler: clang
 
-    # ASAN disabled for clang builds pending resolution of
+    # ASAN disabled for clang < 3.8 builds per resolution of
     #  https://llvm.org/bugs/show_bug.cgi?id=22757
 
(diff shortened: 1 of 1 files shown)


ericniebler:
Looks like clang 3.8 release build are failing.

CaseyCarter:
Yes. I'm trying to replicate the Travis setup to reproduce.

CaseyCarter:
The errors appear to be an interaction between ASAN and valgrind. The ASAN setup code blows up and complains that something has been already been mapped - i conjecture by valgrind - into address space that it wants.

I managed to reproduce it, but an hour or so of investigation turned up no simple solution. I'm not motivated to invest more time, so for now I disabled the valgrind tests for ASAN compiles.

ericniebler:
Sounds reasonable. Thanks.

PR #315, opened 2016-04-07
intersperse_view::sentinel_adaptor::empty should be const

CaseyCarter:
...for obvious reasons, and so that the correctness of the view doesn't depend on `sentinel_adaptor` being empty & trivial to trigger the `compressed_pair` optimization.


ericniebler:
Thanks!

PR #316, opened 2016-04-09
MSFT ABI != _MSC_VER (and unrelated minor changes)

CaseyCarter:
* Detect Microsoft ABI with `_WIN32`, not `_MSC_VER` (Clang/C2 defining `_MSC_VER` is a transition mechanism that will go away soonish).
* Set -march and -mtune for debug builds as well (i.e., use same ISA for debug and release).
* concepts.hpp doesn't need `<iostream>` or `<range/v3/utility/associated_types.hpp>`
* The return type of `indirect_swap` should be "don't care" instead of "must be ``void`."


ericniebler:
Thanks!

pfultz2:
`_WIN32` != microsoft ABI either. Gcc supports EBO on windows, which `_WIN32` is of course defined.

CaseyCarter:
> `_WIN32` != microsoft ABI either. Gcc supports EBO on windows, which `_WIN32` is of course defined.

True. Fixed by #317.

PR #325, opened 2016-04-18
tagged<Base, Tags...> should be constructible from Base&& and Base const&

### LINKED ISSUE #324: libc++ Travis builds are broken
Travis builds use trunk libc++. libc++ maintainers applied a tuple constructor change on 2016-04-15 (r266461).

Hmmm...what constructor is used right now to convert a `Base` to a `tagged<Base, Tags...>`? I strongly suspect it is the inherited tuple-converting constructor, which is now constrained not to accept tuples of the same type (sensibly).

This is a design error: copy and move constructors are not inherited. If we want `tagged<Base, Tags...>` to be constructible from `Base&&` and/or `Base const&`, we need to add appropriate constructors to `tagged`.

Needs to be fixed in the TS, also.

CaseyCarter:
Fixes #324.

### DIFF
--- a/include/range/v3/utility/tagged_pair.hpp
+++ b/include/range/v3/utility/tagged_pair.hpp
@@ -74,10 +74,18 @@ namespace ranges
           : Base
           , detail::getters::collect<tagged<Base, Tags...>, Tags...>
         {
-            using Base::Base;
             tagged() = default;
             tagged(tagged &&) = default;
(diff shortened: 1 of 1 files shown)


CaseyCarter:
Yay transient Travis errors!

ericniebler:
Thanks!

Issue #160, opened 2015-05-29
Should the deprecated random_shuffle be removed?


rhalbersma:
Now that range-v3 itself does not use `random_shuffle` anywhere in its implementation and unit tests, it is possible to remove `algorithm::random_shuffle` from range-v3. 

One argument against this would be that `std::random_shuffle` is only deprecated from C++14 and beyond. On the other hand, the recommended migration route of `std::shuffle` is already available in C++11. This migration is also fully supported in range-v3 via both an algorithm and an action, whereas `random_shuffle` is only half-heartedly supported (no action, only an algorithm). 

So it seems that range-v3 could safely remove `random_shuffle` without too much pain on either its C++11 or it C++14 users.

ericniebler:
I actually don't understand why the committee chose to deprecate `random_shuffle` instead of requiring it to use a random number generator that doesn't suck, like `mt19937` seeded with `random_device`. It's a useful wrapper. I see no reason to remove it, but perhaps I'm missing something.

rhalbersma:
The committee not just deprecated `random_shuffle`, but, based on [N4190](https://isocpp.org/files/papers/N4190.txt), the current working paper actually *removes* it altogether, see [diff.cpp14.depr] in [N4296](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf): 

>The class templates auto_ptr, unary_function, and binary_function, **the function templates     random_shuffle**, and the function templates (and their return types) ptr_fun, mem_fun, mem_fun_ref, bind1st, and bind2nd **are not defined**.

A C++11/14/17 library should then follow suit and remove them as well.

But I agree that wrapping the generator is convenient. AFAICS, the wrapping `random_shuffle` overload that you define already uses a uniform random generator. So couldn't that be simply ported to a new wrapping `shuffle` overload? That could then be part of a future committee proposal or STL2.

CaseyCarter:
Thread-safety. It's too late to change `random_shuffle` to either (a) use a `mutex` and a global engine, or (b) use a `thread_local` engine. Either choice is a breaking change. As is changing the engine: it would break reproducability of results. [N4316](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4316.html) provides an overload of `shuffle` that uses a `thread_local` engine to provide the same interface that `random_shuffle` did.

rhalbersma:
[...]

PR #326, opened 2016-04-18
Disable deprecation warnings for test.alg.random_shuffle.

CaseyCarter:
...when cmake supports `target_compile_options`. While we're at it, let's get the cmake version recorded in the Travis build log as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -318,6 +318,7 @@ install:
   - if [ -n "$CLANG_VERSION" ]; then CXX_FLAGS="${CXX_FLAGS} -D__extern_always_inline=inline"; fi
   - if [ "$LIBCXX" == "On" ]; then CXX_FLAGS="${CXX_FLAGS} -stdlib=libc++ -I/usr/include/c++/v1/"; fi
   - if [ "$LIBCXX" == "On" ]; then CXX_LINKER_FLAGS="${CXX_FLAGS} -L/usr/lib/ -lc++"; fi
+  - cmake --version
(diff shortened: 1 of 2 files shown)


rhalbersma:
Why not follow C++17 and remove `random_shuffle` outright? Users making transition from old STL to range-v3/STL2 should not be enabled in their deprecated habits.

CaseyCarter:
> Why not follow C++17 and remove `random_shuffle` outright?

I have no strong feelings about removing the `random_shuffle` algorithm. I *do* have strong feelings about deprecation warnings when the test compiles.

CaseyCarter:
FYI: the Travis failures are due to #325.

CaseyCarter:
New push uses the more-widely-supported `target_compile_definitions` property (cmake >= 2.8.0) instead of `target_compile_options` (cmake >= 2.8.12).

capsocrates:
Removing random_shuffle from STL2 is a great idea.
http://en.cppreference.com/w/cpp/algorithm/random_shuffle implies that it's
been removed from the C++17 standard.

CaseyCarter:
> Removing `random_shuffle`

This pull request is about making `alg.random_shuffle` not fire deprecation warnings when it compiles, it has nothing to do with removing `random_shuffle`. As such, this thread is a silly place to discuss removing `random_shuffle` since that discussion will disappear when the pull request is closed.

Discussion about removing `random_shuffle` should take place in either #160, or a pull request that removes `random_shuffle`. We don't want valuable discussion to be...lost in the shuffle. (Sorry, could not help it.)

PR #328, opened 2016-04-19
remove random_shuffle

### LINKED ISSUE #160: Should the deprecated random_shuffle be removed?

rhalbersma:
std::random_shuffle will be removed in C++17. Closes #160


ericniebler:
We should also add an overload of `shuffle` that has the signature of `random_shuffle`.

CaseyCarter:
Pull request should note that this closes issue #160.

include/range/v3/algorithm/shuffle.hpp
```diff
@@ -76,6 +76,16 @@ namespace ranges
                 return orig;
             }
 
+            template<typename I, typename S,
+                CONCEPT_REQUIRES_(RandomAccessIterator<I>() && IteratorRange<I, S>() &&
+                    Permutable<I>())>
+            I operator()(I begin, S end_) const
+            {
+                std::random_device rng;
+                std::mt19937 urng(rng());
```
CaseyCarter: Do we want `urng` to be `thread_local` to avoid initialization cost on repeat calls in the same thread? I think yes. Do we want to seed with more than a single `unsigned int` of entropy? I think no. In both cases, my answers are chosen to most closely resemble `random_shuffle` (i.e., I'm "weakly for").

rhalbersma: OK, I'll add `thread_local` later tonight. Do you want to select `mt19937_64` when targetting 64-bit builds?

CaseyCarter: > OK, I'll add `thread_local` later tonight. Do you want to select `mt19937_64` when targetting 64-bit builds?

Should probably be `RANGES_STATIC_THREAD_LOCAL`; I don't know why the macro exists, but I assume there's a good reason.

In my (limited) experience `mt19937_64` is just slightly more efficient than `mt19937` on 64 bit implementations,
but I haven't seen hard numbers to recommend either choice - so put me down as "neutral."

ericniebler: Keep it simple, please. We can complexify it later if there is a demonstrated need.

PR #330, opened 2016-04-21
[RANGES_FOR] Use for directly in C++1z. Fixes #329.

### LINKED ISSUE #329: Update `RANGES_FOR` for C++1z
`RANGES_FOR` should use C++1z's range-based for-loop when available (e.g. in clang > 3.8).

CaseyCarter:


CaseyCarter:
Tested against GCC7, which does activate the change.

PR #331, opened 2016-04-21
split iota view header into ints view header and iota view header

krzysztof-jusiak:
Currently to include ints view one has to include iota which might be a bit confusing

This simple split allows to include ints view using ints.hpp (will still include iota.hpp) and ioata view without including ints view by using iota.hpp

It's just a change in file names, functionality hasn't changed, unless, someone was depending on ints being included by iota.


.travis.yml
```diff
@@ -319,7 +319,7 @@ install:
   - if [ "$LIBCXX" == "On" ]; then CXX_FLAGS="${CXX_FLAGS} -stdlib=libc++ -I/usr/include/c++/v1/"; fi
   - if [ "$LIBCXX" == "On" ]; then CXX_LINKER_FLAGS="${CXX_FLAGS} -L/usr/lib/ -lc++"; fi
   - cmake .. -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_EXE_LINKER_FLAGS="${CXX_LINKER_FLAGS}" -DRANGES_CXX_STD=$CPP -DRANGE_V3_NO_HEADER_CHECK=1
-  - make VERBOSE=1
+  - travis_wait make VERBOSE=1
```
ericniebler: What does this do?

ericniebler:
This change is well-motivated but would probably be pretty disruptive to downstream clients. I'm not certain the benefit outweighs the cost.

.travis.yml
krzysztof-jusiak: I was trying to fix the travis build as it seems to take a long time on gcc and sometimes there is no output for longer than 20 minutes which is a limit on travis after which the build is killed. 'travis_wait' is printing every few seconds to the output to prevent it. Nevertheless, it's not related to this PR, so sorry about it, just tried to make the PR checks to pass.

krzysztof-jusiak:
I'm not sure either whether the benefit outweighs the cost, it's your call in the end.
However, from my point of view, if you find this change useful, it's better to do it now when v1.0.0 is not released yet as afterwards it will be much harder. Having said that, I don't know how many users there are, so I might be wrong and it might be too late already.

krzysztof-jusiak:
Just an after thought. This PR would actually be irrelevant if compilation times would be good enough to simply include

```cpp
#include <range>
```

like in STL (as we don't have modules yet to import what we need), however, it's not really feasible in the current state of the library

All in all, I'm just trying to avoid writing comments about what is inluded...

```cpp
#include <range/v3/view/iota.hpp>  // view::ints
```

FloopCZ:
@ericniebler Now, range-v3 is properly versioned, so maybe this PR could be included in v0.2.0 or milestoned for the next minor release?

gnzlbg:
I think it would be better to just have `view/ints.hpp` include `view/iota.hpp` and be done with it. This way one does not affect downstream clients at all: those including `view/iota.hpp` still get ints, and those who by accident write `view/ints.hpp` get the correct behavior as well.

ericniebler:
I'm increasingly of the opinion that `view::ints` should be deprecated and replaced with `view::indices` (or `indexes`), per #592.

ericniebler:
Closing this since we are going a different direction on `view::ints`.

PR #334, opened 2016-04-24
Repurpose RANGES_CXX_NO_VARIABLE_TEMPLATES…

CaseyCarter:
…as `RANGES_USE_VARIABLE_TEMPLATES`, detect with `__cpp_variable_templates`. (i.e., use variable templates in GCC6+, not just in clang.)


ericniebler:
I prefer the negative feature check. `RANGES_NO_VARIABLE_TEMPLATES `

CaseyCarter:
> I prefer the negative feature check. `RANGES_NO_VARIABLE_TEMPLATES`

I'll re-push without the sense change; but you should know that every time someone reads `#if !RANGES_NO_XXX` a fairy dies.

ericniebler:
Ten year from now, are you going to want to tell people they need to #define RANGES_USE_VARIADIC_TEMPLATES to use range-v3?

CaseyCarter:
> Ten year from now, are you going to want to tell people they need to `#define RANGES_USE_VARIADIC_TEMPLATES` to use range-v3?

No - and that wasn't the usage model in the PR. You did make me realize that I didn't go far *enough*. Ten years from now, I'd like for feature detection to use the SD-6 facilities when they're available, and fallback to `__cplusplus` when they are not. Users should be able to override feature detection from the command line / build system.

So here's a rework of the PR with those features.

pfultz2:
> Users should be able to override feature detection from the command line / build system, and be able to select no variable templates, C++14 variable templates, or C++23 variable templates.

+1

Boost.Config is huge pain to configure the features. The way the macro is defined makes it easy for the user to configure when its incorrect(or some other reason).

ericniebler:
Nice!

PR #346, opened 2016-05-02
Deleted overloads don't need the proper return type. Fixes #345.

### LINKED ISSUE #345: Compilation errors with _GLIBCXX_DEBUG defined
Thanks very much for the work on #294.

Unfortunately, I'm now seeing a new compilation failure that I suspect arises from those changes. I'm now getting a failure under both GCC and Clang from just including `<range/v3/core.hpp>` when `_GLIBCXX_DEBUG` is defined.

GCC errors:

~~~~~
In file included from range-v3/include/range/v3/utility/iterator_traits.hpp:20:0,
                 from range-v3/include/range/v3/utility/iterator.hpp:25,
                 from range-v3/include/range/v3/begin_end.hpp:24,
                 from range-v3/include/range/v3/core.hpp:17,
                 from a.cpp:1:
range-v3/include/range/v3/utility/iterator_concepts.hpp:629:15: error: ‘iterator_difference_t’ in namespace ‘ranges’ does not name a template type
     ::ranges::iterator_difference_t<I1>
               ^
range-v3/include/range/v3/utility/iterator_concepts.hpp:634:15: error: ‘iterator_difference_t’ in namespace ‘ranges’ does not name a template type
     ::ranges::iterator_difference_t<I1>
               ^
~~~~~

Or am I just missing something that I should now be doing when using `_GLIBCXX_DEBUG`?

Thanks very much.

CaseyCarter:


include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -625,14 +625,14 @@ namespace ranges
 namespace __gnu_debug
 {
     template <class I1, class I2, class Seq,
-        CONCEPT_REQUIRES_(!::ranges::SizedIteratorRange<I1, I2>())>
-    ::ranges::iterator_difference_t<I1>
-        operator-(_Safe_iterator<I1, Seq> const &, _Safe_iterator<I2, Seq> const &) = delete;
+        CONCEPT_REQUIRES_(!::ranges::SizedIteratorRange<I2, I1>())>
```
CaseyCarter: `S{} - I{}` requires `SizedIteratorRange<I, S>()`/`SizedSentinel<S, I>()` and has type `iterator_difference_t<I>`/`difference_type_t<I>`.

ericniebler:
Sure, but why is `iterator_difference_t` not available at this point? Is there a real bug here? I don't have time to investigate ATM.

CaseyCarter:
>  Is there a real bug here?

Nothing fancy: `iterator_difference_t` is defined in `iterator_traits.hpp` which includes - and therefore is not included by - `iterator_concepts.hpp`.

We should probably define the associated type aliases in range_fwd.hpp so they're "global."

CaseyCarter:
> We should probably define the associated type aliases in range_fwd.hpp so they're "global."

And switch to the TS naming scheme for the iterator assoc. type aliases as well.

ericniebler:
> And switch to the TS naming scheme for the iterator assoc. type aliases as well.

That would be majorly disruptive. We should probably leave the old aliases around for folks using them.

gnzlbg:
> That would be majorly disruptive. We should probably leave the old aliases around for folks using them. 

Leaving the old aliases and tagging them with a deprecated attribute might be the right way to go but IIRC _nobody_ has complained on github issues about some of the biggest breakages of the recent past [*]... so I would say don't give this that much thought and go for it. 

I'd rather have the breakage sooner rather than later and if you really want to make me happy, give the culprit commit a title that _really_ sets it apart from other commits in the history so that I can easily pin point it out and update my code.  

[*] While this invalidates my point of nobody ever complaining,  there have been a couple of renames with different levels of breakage: `range` to `iterator_range`, the meta-rename, the `Iterable`-pocalypse... as well as things like the recent cursor refactorings.

PR #347, opened 2016-05-02
Disable GCC5 Release+ASAN config

CaseyCarter:
It often runs over the Travis-mandated 50 minute per-job time limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -256,9 +256,9 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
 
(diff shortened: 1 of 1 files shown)


ericniebler:
This is all kinds of broken on Travis. Can you take a look?

CaseyCarter:
I moved the "edit CMakeLists to workaround the valgrind bug" bit *after* the change from the source directory to the build directory, thus breaking all valgrind builds.

ericniebler:
> I need to go back to my pattern of waiting for changes to clear https://travis-ci.org/CaseyCarter/range-v3 before pushing so it looks like I never make mistakes.

It's not skin off my back if tests fail on your PRs. I'm not going to merge them until I see the tests are green.

PR #348, opened 2016-05-02
gcc 5+ has std::is_trivially_foo

CaseyCarter:


include/range/v3/range_fwd.hpp
```diff
@@ -392,8 +392,6 @@ namespace ranges
         template<typename I, typename S = I>
         struct sized_iterator_range;
 
-        struct make_range_fn;
```
CaseyCarter: This declaration is dead.

ericniebler:
Nice.

PR #350, opened 2016-05-03
[size] Update to TS spec:

CaseyCarter:
* Prefer member to non-member size
* Ensure that size always returns an `Integral`


CaseyCarter:
I notably did *NOT* alter `size` to accept its argument by `const&` as specified in the TS, since that breaks some of the views.

gnzlbg:
> I notably did NOT alter size to accept its argument by const& as specified in the TS, since that breaks some of the views.

Which views and why? Do they need to mutate something to compute their size?

CaseyCarter:
> Which views and why? Do they need to mutate something to compute their size?

Views that lazily compute `begin` or `end`. See #385 for a more complete rant.

PR #351, opened 2016-05-04
sort_patterns cleanup

CaseyCarter:
Various piddly cleanups in sort_patterns and counted_insertion_sort

EDIT: Removed the configuration-related changes from this PR.


CaseyCarter:
This leaves view/join as the only user of `RANGES_CXX_STD`; it's doing something weird that I haven't yet analyzed.

PR #355, opened 2016-05-09
fix libstdcxx4.9 support and test clang+libstdc++ on travis (C++11 only)

gnzlbg:
Instead of checking for clang this checks for libc++.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -215,6 +215,10 @@ matrix:
       os: linux
       addons: *clang38
 
+    - env: CLANG_VERSION=3.8 BUILD_TYPE=Release CPP=11 ASAN=Off LIBCXX=Off
+      os: linux
+      addons: *clang38
+
     # Test gcc-4.8: C++11, Build=Debug/Release, ASAN=Off
     - env: GCC_VERSION=4.8 BUILD_TYPE=Debug CPP=11 ASAN=Off LIBCXX=Off
       os: linux
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -254,23 +254,37 @@ namespace ranges
             void ignore_unused(Ts &&...)
             {}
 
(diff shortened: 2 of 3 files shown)


CaseyCarter:
This pessimizes clang+libstdc++ when libstdc++ >= 5. Why not simply implement the type traits under `ranges::v3::detail` for clang without libc++ and call it a day, e.g.:

```c++
#include <type_traits>
#if defined(__clang__) && !defined(_LIBCPP_VERSION)
    template<typename T>
    using is_trivially_copy_assignable =
        meta::bool_<__is_trivally_assignable(T &, T const&)>;

    template<typename T>
    using is_trivially_move_assignable =
        meta::bool_<__is_trivally_assignable(T &, T &&)>;

#elif defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5
    template<typename T>
    using is_trivially_copy_assignable = std::is_trivial<T>;
 
    template<typename T>
    using is_trivially_move_assignable = std::is_trivial<T>;

#else
using std::is_trivially_copy_assignable;
using std::is_trivially_move_assignable;
#endif
 ```
For bonus points. solve the `is_final` issue similarly ;)

gnzlbg:
>This pessimizes clang+libstdc++ when libstdc++ >= 5

Is this because clang defines `__GNUC__` to 4 even when using libstdc++>=5?

>  Why not simply implement the type traits under ranges::v3::detail for clang without libc++ and call it a day,

Will update the PR.

CaseyCarter:
> Is this because clang defines `__GNUC__` to `4` even when using libstdc++>=5?

Yip.

CaseyCarter:
@gnzlbg Oops - you lost the Travis change when you re-pushed.

gnzlbg:
Should be fixed now.

ericniebler:
Thanks guys!

PR #358, opened 2016-05-11
More config system work:

CaseyCarter:
* Add feature detection for return type deduction and generic lambdas
* Enable deprecation warnings again, silence the single resulting warning in test/utility/variant.hpp
* Bring diagnostic control/suppression into the config system
* Add Travis support for GCC 4.9 -std=c++14, and GCC 5 -std=c++14 with constexpr limited to C++11 (`-DRANGES_CXX_CONSTEXPR=RANGES_CXX_CONSTEXPR_11`)


include/range/v3/algorithm.hpp
```diff
@@ -98,6 +99,6 @@
 #include <range/v3/algorithm/aux_/sort_n_with_buffer.hpp>
 #include <range/v3/algorithm/aux_/upper_bound_n.hpp>
 
-#include <range/v3/detail/re_enable_warnings.hpp>
+RANGES_RE_ENABLE_WARNINGS
```
ericniebler: The spelling has always bothered me. How about `RANGE_V3_REENABLE_WARNINGS` ?

ericniebler: Ah, never mind.

CaseyCarter: How do you reen a warning, so as to determine which warnings are the reenable warnings? ;)

How about simply `ENABLE_WARNINGS`?

PR #359, opened 2016-05-11
Workaround GCC compile failures in calendar.cpp

CaseyCarter:


ericniebler:
No! This is the wrong fix. If GCC is correct, then range-v3's use of EBO for function object types is wrong and needs to be fixed everywhere.

PR #361, opened 2016-05-12
view::join rework

CaseyCarter:
* BUGFIX: Don't rely on value-initialized **input** iterators comparing equal.
* BUGFIX: Don't use `accumulate()` to calculate `size()` if the underlying range is not Forward.
* Use `default_sentinel`; store state in the range instead of the iterators.

### DIFF
--- a/include/range/v3/view/join.hpp
+++ b/include/range/v3/view/join.hpp
@@ -23,10 +23,11 @@
 #include <range/v3/begin_end.hpp>
 #include <range/v3/empty.hpp>
 #include <range/v3/range_traits.hpp>
-#include <range/v3/view_adaptor.hpp>
-#include <range/v3/view/transform.hpp>
 #include <range/v3/utility/functional.hpp>
 #include <range/v3/utility/static_const.hpp>
+#include <range/v3/utility/variant.hpp>
+#include <range/v3/view_facade.hpp>
+#include <range/v3/view/transform.hpp>
 #include <range/v3/view/all.hpp>
 #include <range/v3/view/view.hpp>
 #include <range/v3/view/single.hpp>
@@ -39,8 +40,8 @@ namespace ranges
         namespace detail
         {
             // Compute the cardinality of a joined range
-            template<typename Outer, typename Inner, typename Joiner = std::integral_constant<cardinality, static_cast<cardinality>(0)>>
-            using join_cardinality =
+            template<typename Outer, typename Inner, typename Joiner>
+            using join_cardinality_ =
                 std::integral_constant<cardinality,
                     Outer::value == infinite || Inner::value == infinite || (Joiner::value == infinite && Outer::value != 0 && Outer::value != 1) ?
                         infinite :
@@ -49,6 +50,12 @@ namespace ranges
                             Outer::value == finite || Inner::value == finite || (Joiner::value == finite && Outer::value != 0 && Outer::value != 1) ?
                                 finite :
                                 static_cast<cardinality>(Outer::value * Inner::value + (Outer::value == 0 ? 0 : (Outer::value - 1) * Joiner::value))>;
+            template<typename Range, typename JoinRange = void>
+            using join_cardinality =
+                join_cardinality_<range_cardinality<Range>, range_cardinality<range_reference_t<Range>>,
+                    meta::if_<std::is_same<void, JoinRange>,
+                        std::integral_constant<cardinality, static_cast<cardinality>(0)>,
+                        range_cardinality<JoinRange>>>;
         }
         /// \endcond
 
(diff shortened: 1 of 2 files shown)


include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,253 @@ namespace ranges
...
-            CONCEPT_ASSERT(Range<range_reference_t<Rng>>());
-            using size_t_ = common_type_t<range_size_t<Rng>, range_size_t<range_value_t<Rng>>>;
+            using size_type = common_type_t<range_size_t<Rng>, range_size_t<range_reference_t<Rng>>>;
 
+            join_view() = default;
+            explicit join_view(Rng rng)
+              : outer_{view::all(std::move(rng))}
+              // Delay initialization of the other members until begin() is first called,
+              // so that we can create a join_view over e.g. an infinite range of empty ranges,
+              // and avoid hanging if begin is never called.
```
ericniebler: I think we need to at least value-initialize the members. Otherwise, they might be singular, and copying/moving the `join_view` is UB, IIUC.

CaseyCarter: All members are value-initialized with default member initializers.

ericniebler:
Have you benchmarked this? `view::join` is the magic ingredient in range comprehensions (`view::for_each`). Please look at the [post about it](http://ericniebler.com/2014/04/27/range-comprehensions/). I make claims there about perf. Don't turn me into a liar, please! :-)

CaseyCarter:
> Have you benchmarked this?
Nope. And wow, is it slow. Moving state back into the iterator.

CaseyCarter:
> Nope. And wow, is it slow. Moving state back into the iterator.

Here's hoping no one ever benchmarks the cost if iterator copies ;)

CaseyCarter:
There's some nasty UB in here - closing the PR til it's fixed.

include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,240 @@ namespace ranges
...
             public:
                 using single_pass = std::true_type;
-                adaptor() = default;
-                adaptor(join_view &rng)
-                  : rng_(&rng), it_{}
-                {}
-                range_iterator_t<Rng> begin(join_view &)
+                cursor() = default;
+                cursor(join_view &rng)
+                  : cursor::box{ranges::end(rng.outer_)}
```
ericniebler: `cursor::box`? What kind of crazy syntax is that?

include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,240 @@ namespace ranges
...
+            {
+                return accumulate(view::transform(outer_, ranges::size), size_type{0});
+            }
+        private:
             friend range_access;
-            view::all_t<range_reference_t<Rng>> cur_;
+            Rng outer_{};
 
-            struct adaptor : adaptor_base
+            class cursor : box<range_sentinel_t<Rng>, Rng>
```
ericniebler: Using `Rng` as the tag type is likely to make symbols long for no good reason that I can see.

include/range/v3/view/join.hpp
CaseyCarter: Lazy syntax: cursor inherits box's injected name.

include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,240 @@ namespace ranges
...
-                    return ranges::indirect_move(it_);
-                }
-                void distance_to() = delete;
+                auto get() const
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    *inner_it_
+                )
+                auto indirect_move() const
+                RANGES_DECLTYPE_AUTO_RETURN
```
ericniebler: `RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT`

include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,240 @@ namespace ranges
...
+                range_cardinality<Rng>::value >= 0 && ForwardRange<Rng>() &&
+                SizedRange<range_reference_t<Rng>>())
+            size_type size() const
+            {
+                return accumulate(view::transform(outer_, ranges::size), size_type{0});
+            }
+        private:
             friend range_access;
-            view::all_t<range_reference_t<Rng>> cur_;
+            Rng outer_{};
```
ericniebler: `view::all_t<Rng> outer_{};`

include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,240 @@ namespace ranges
...
-            struct adaptor : adaptor_base
+            class cursor : box<range_sentinel_t<Rng>, Rng>
             {
             private:
-                join_view *rng_;
-                range_iterator_t<range_reference_t<Rng>> it_;
-                void satisfy(range_iterator_t<Rng> &it)
+                range_iterator_t<Rng> outer_it_{};
+                view::all_t<range_reference_t<Rng>> inner_{};
+                range_iterator_t<view::all_t<range_reference_t<Rng>>> inner_it_{};
```
ericniebler: By definition, `range_iterator_t<R>` is the same as `range_iterator_t<view::all_t<R>>`.

CaseyCarter:
The bug was a result of storing the inner range directly in the iterator. Since the inner-range iterator also points into/denotes that range, the resulting iterator doesn't satisfy `CopyConstructible`: the inner-range iterator in a copy of `foo` points into `foo`'s range.

Unfortunately, performance of the range comprehensions examples is more like 5s when the inner range is stored in the outer range than 2s as in the blog post when the inner range is stored in the iterator. The performance regression is not a result of this PR, master already had similar numbers before the PR.

PR #362, opened 2016-05-12
view::stride:

CaseyCarter:
* BUGFIX: `distance_to` works with non-Bidi iterators
* Minimize atomic operations


ericniebler:
BUGFIX, but no new test case == 😢

CaseyCarter:
> BUGFIX, but no new test case

Done. Took me longer than I care to admit to discover that `move_cursor::distance_to` was inappropriately commented out.

PR #365, opened 2016-05-15
view::join rework

CaseyCarter:
* BUGFIX: Don't rely on value-initialized **input** iterators comparing equal.
* BUGFIX: Don't use accumulate() to calculate size() if the underlying range is not Forward.
* Use default_sentinel, avoid view_adaptor.

### DIFF
--- a/include/range/v3/view/join.hpp
+++ b/include/range/v3/view/join.hpp
@@ -23,10 +23,11 @@
 #include <range/v3/begin_end.hpp>
 #include <range/v3/empty.hpp>
 #include <range/v3/range_traits.hpp>
-#include <range/v3/view_adaptor.hpp>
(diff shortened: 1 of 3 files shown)


include/range/v3/view/join.hpp
```diff
@@ -58,259 +65,235 @@ namespace ranges
...
-                auto get(range_iterator_t<Rng> const &) const ->
-                    common_reference_t<
-                        range_reference_t<range_reference_t<Rng>>,
-                        range_reference_t<ValRng>>
+                reference get() const
                 {
-                    if(toggl_)
-                        return *it_;
-                    return *val_it_;
+                    // return visit(cur_, [](auto& it) -> reference { return *it; });
```
ericniebler: Does visitation not work for this?

PR #366, opened 2016-05-15
Lambda ebo gcc6 snafu

ericniebler:


include/range/v3/utility/functional.hpp
```diff
@@ -295,44 +349,80 @@ namespace ranges
...
+        // Lambdas are *not*:
+        // - aggregates             ([expr.prim.lambda]/p3)
+        // - literal types          ([expr.prim.lambda]/p3)
+        // - default constructible  ([expr.prim.lambda]/p20)
+        // - copy assignable        ([expr.prim.lambda]/p20)
+        template<typename Fn>
+        using could_be_lambda_ =
+            meta::bool_<
+               // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71116
+               //!std::is_literal_type<Fn>::value &&
```
CaseyCarter: `constexpr` lambdas were voted into the WP in JAX, so [lambda closures may be literal in C++1z](https://github.com/cplusplus/draft/commit/7dcae6f1def23343816a4f99ba50f1c4a767749d#diff-64a673ab7a20068092a4cf86b36f4334L657).

ericniebler: Huh, cool. Maybe it's not a compiler bug.

CaseyCarter: > Maybe it's not a compiler bug.

Since the change wasn't a DR, closures should be literal in non-C++1z mode - that's a bug. I'm just pointing out that `is_literal` is not generally useful for detecting lambda-ness.

PR #370, opened 2016-05-17
Saner rules for common_type/common_reference of common_foo and std::foo.

### LINKED ISSUE #367: zip view returns range with non-Readable iterators
The following code fails to compile:
```c++
#include <range/v3/all.hpp>

namespace r = ranges;
namespace v = ranges::view;

int main() {
  auto zints = v::for_each(v::ints(0,10), [](int i) {
    return v::zip(v::repeat_n(i,2),v::ints(0,2));
  });
  static_assert(r::Readable<decltype(r::begin(zints))>{}, "Iterator type not readable");
}
```

I assume the common_reference_t code is failing in the Readable concept (it is SemiRegular and Movable).  Does it need a specialization to handle this case?  Or am I missing something?

Aside: what I'd really like is to take a list like [1 2 3] and output something with each element incremented and decremented, like [0 2 1 3 2 4].  I could do that by transforming zints (if zints were Readable).  Is there an easier way?  The requirement that yield_from take a not-container-like View prevents the simplest way (for solid technical reasons, I'm sure.)

Thanks for the great library, and the std.proposal!

CaseyCarter:
common_foo, common_foo => common_foo
common_foo, std::foo => common_foo if there are member references, otherwise std::foo

Fixes #367.

### DIFF
--- a/include/range/v3/utility/common_tuple.hpp
+++ b/include/range/v3/utility/common_tuple.hpp
@@ -415,42 +415,42 @@ namespace ranges
                     std::tuple<Ts...>>;
         }
 
-        // common_type for std::pairs
+        // common_type for pairs
         template<typename F1, typename S1, typename F2, typename S2>
         struct common_type<std::pair<F1, S1>, common_pair<F2, S2>>
           : detail::common_type_tuple_like<std::pair<F1, S1>, common_pair<F2, S2>,
-                meta::quote<std::pair>>
+                meta::quote<detail::make_common_pair>>
         {};
 
         template<typename F1, typename S1, typename F2, typename S2>
         struct common_type<common_pair<F1, S1>, std::pair<F2, S2>>
           : detail::common_type_tuple_like<common_pair<F1, S1>, std::pair<F2, S2>,
-                meta::quote<std::pair>>
+                meta::quote<detail::make_common_pair>>
         {};
 
         template<typename F1, typename S1, typename F2, typename S2>
         struct common_type<common_pair<F1, S1>, common_pair<F2, S2>>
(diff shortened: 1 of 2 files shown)


ericniebler:
I can almost remember my reason for doing that weirdness. But I can't. If the tests pass then I'm fine with this.

include/range/v3/utility/common_tuple.hpp
```diff
@@ -415,42 +415,42 @@ namespace ranges
                     std::tuple<Ts...>>;
         }
 
-        // common_type for std::pairs
+        // common_type for pairs
         template<typename F1, typename S1, typename F2, typename S2>
         struct common_type<std::pair<F1, S1>, common_pair<F2, S2>>
           : detail::common_type_tuple_like<std::pair<F1, S1>, common_pair<F2, S2>,
-                meta::quote<std::pair>>
+                meta::quote<detail::make_common_pair>>
```
ericniebler: Why not just `common_pair`?

CaseyCarter: `make_foo` encapsulates the "choose `foo` if there are reference members and `std::foo` otherwise" mechanic. I also vaguely recall there being a reason to favor the `std` flavors, so I'm trying to be conservative with the change.

CaseyCarter: >  I also vaguely recall there being a reason to favor the `std` flavors

It's possible it was simply an attempt to minimize the disruption of introducing the `common_foo` types. Anyway, the tests don't seem to care.

ericniebler:
Thanks!

PR #375, opened 2016-05-22
Update detail::as_const to N4582 spec.

CaseyCarter:


ericniebler:
Maybe we should promote this out of `detail::`.

CaseyCarter:
> Maybe we should promote this out of `detail::`.

Ironically enough, I implemented `as_const` in `ranges::v3` on a half-finished customization point object update branch before noticing the version that's already in `detail`. I have no strong feelings either way.

PR #376, opened 2016-05-23
Eliminate `view_adaptor_t`

CaseyCarter:
Both the alias template and the `view_adaptor` member of the same name.

While doing so, I noticed that `view_interface` and `view_facade` have public destructors which could be made private to reduce the danger of polymorphically deleting a class derived from either/both.

The compiler then complained about `drop_while_view`'s copy constructor failing to copy its `view_interface` base, causing me to notice that `drop_while_view` has mutually exclusive predicate and iterator members that should be stored in a variant.

### DIFF
--- a/include/range/v3/view/drop_while.hpp
+++ b/include/range/v3/view/drop_while.hpp
@@ -38,10 +38,9 @@ namespace ranges
         template<typename Rng, typename Pred>
         struct drop_while_view
           : view_interface<drop_while_view<Rng, Pred>, is_finite<Rng>::value ? finite : unknown>
-       {
+        {
         private:
             friend range_access;
-            using difference_type_ = range_difference_t<Rng>;
             Rng rng_;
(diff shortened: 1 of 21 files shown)


include/range/v3/view/drop_while.hpp
```diff
@@ -43,43 +43,22 @@ namespace ranges
...
-                begin_.reset();
-                return *this;
-            }
-            drop_while_view& operator=(drop_while_view const &that)
-            {
-                rng_ = that.rng_;
-                pred_ = that.pred_;
-                begin_.reset();
-                return *this;
-            }
```
ericniebler: The point of the copy/assign overload was that if we call `begin` on a `drop_while` range, we end up with an iterator in `begin_`. Then if we copy the range, the destination gets an iterator that points into the underlying source range, not the underlying destination range. So I think we need to keep the code as it was.

PR #387, opened 2016-06-01
Mutable istream_range / getlines_range

CaseyCarter:
`istream` and `getlines` are poster children for mutable input ranges: Why copy elements/lines from the input range when we can move them out?

### DIFF
--- a/include/range/v3/getlines.hpp
+++ b/include/range/v3/getlines.hpp
@@ -47,13 +47,17 @@ namespace ranges
                 {
                     rng_->next();
                 }
-                std::string const &get() const
+                std::string &get() const noexcept
                 {
                     return rng_->str_;
                 }
                 bool done() const
                 {
                     return !*rng_->sin_;
+                }
+                std::string && move() const noexcept
+                {
+                    return detail::move(rng_->str_);
                 }
             };
             void next()
@@ -71,7 +75,7 @@ namespace ranges
             {
                 this->next(); // prime the pump
             }
-            std::string & cached()
+            std::string & cached() noexcept
             {
                 return str_;
             }
--- a/include/range/v3/istream_range.hpp
+++ b/include/range/v3/istream_range.hpp
@@ -47,7 +47,7 @@ namespace ranges
                 {
                     rng_->next();
                 }
-                Val const &get() const
+                Val &get() const
                 {
                     return rng_->cached();
                 }
@@ -55,6 +55,11 @@ namespace ranges
                 {
                     return !*rng_->sin_;
                 }
+                Val && move() const noexcept
+                {
+                    return detail::move(rng_->cached());
+                }
+
             };
             void next()
             {
@@ -78,7 +83,7 @@ namespace ranges
             {
                 next(); // prime the pump
             }
-            Val & cached()
+            Val & cached() noexcept
             {
                 return obj_;
             }
--- a/include/range/v3/istream_range.hpp
+++ b/include/range/v3/istream_range.hpp
@@ -47,7 +47,7 @@ namespace ranges
                 {
                     rng_->next();
                 }
-                Val &get() const
+                Val &get() const noexcept
                 {
                     return rng_->cached();
                 }
--- a/test/getlines.cpp
+++ b/test/getlines.cpp
@@ -11,23 +11,28 @@
 
 #include <sstream>
 #include <range/v3/core.hpp>
+#include <range/v3/range_traits.hpp>
 #include "./simple_test.hpp"
 #include "./test_utils.hpp"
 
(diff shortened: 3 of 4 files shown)


ericniebler:
Well, of you move the strings out of getlines, you defeat the optimization whereby memory of reused across iterations. But your changes doesn't imply that, iiuc.

Care to have a look at the Travis failure?

ericniebler:
> Care to have a look at the Travis failure?

Disregard. It's an overloaded llvm apt repo.

CaseyCarter:
> Disregard. It's an overloaded llvm apt repo.

Yip. From twitter, it looks like it may be down indefinitely. I'll verify clang 3.8 locally.

> Well, if you move the strings out of getlines, you defeat the optimization

True - that motivation is soft for getlines. I can envision line-at-a-time text processing applications that  modify the string in-place before output.

The primary motivation is that there's no reason NOT to make input ranges that store an element(s) in the range object mutable.

ericniebler:
> The primary motivation is that there's no reason NOT to make input ranges that store an element(s) in the range object mutable.

I can buy this, so long as the range is single-pass and does not have `const` `begin`/`end`. You wouldn't want a hypothetical input `view::repeat` to return a mutable reference, for instance.

CaseyCarter:
> I can buy this, so long as the range is single-pass and does not have `const` `begin`/`end`.
I suppose I should clarify that I mean "InputRange && !ForwardRange && !View", which implies single-pass. And yes, `cbegin` on a non-View should return a constant iterator.

> You wouldn't want a hypothetical input `view::repeat` to return a mutable reference, for instance.

A range that repeats the last value written into it sounds like it could be useful - `latch`? - but such a thing is a non-`View`.

gnzlbg:
> input view::repeat

As in, "if nothing new, repeat the last value" ? Or what do you mean?


On Thu, Jun 2, 2016 at 9:35 PM, Casey Carter <notifications@github.com>
wrote:

> I can buy this, so long as the range is single-pass and does not have
> const begin/end.
> I suppose I should clarify that I mean "InputRange && !ForwardRange &&
> !View", which implies single-pass. And yes, cbegin on a non-View should
> return a constant iterator.
>
> You wouldn't want a hypothetical input view::repeat to return a mutable
> reference, for instance.
>
> A range that repeats the last value written into it sounds like it could
> be useful - latch? - but such a thing is a non-View.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/ericniebler/range-v3/pull/387#issuecomment-223398326>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AA3NpuIbTgZhYocP6rUomnslrgAabld7ks5qHzCBgaJpZM4Ir76T>
> .
>



-- 
Dipl.-Ing. Gonzalo Brito Gadeschi
Institute of Aerodynamics and Chair of Fluid Mechanics
RWTH Aachen University
Wuellnerstraße 5a
D-52062 Aachen
Germany
Phone: ++49-(0)241-80-94821
Fax: ++49-(0)241-80-92257
E-mail:  g.brito@aia.rwth-aachen.de
Internet: www.aia.rwth-aachen.de

CaseyCarter:
> As in, "if nothing new, repeat the last value" ? Or what do you mean?

`latch_range(foo)` would store `foo` inside the range object. Its iterator would behave like `view::repeat`'s iterator, except that it references the stored value instead of carrying a copy in the iterator.

include/range/v3/getlines.hpp
```diff
@@ -47,14 +47,18 @@ namespace ranges
                 {
                     rng_->next();
                 }
-                std::string const &get() const
+                std::string &get() const
```
ericniebler: `noexcept`?

include/range/v3/istream_range.hpp
```diff
@@ -47,14 +47,19 @@ namespace ranges
                 {
                     rng_->next();
                 }
-                Val const &get() const
+                Val &get() const
```
ericniebler: `noexcept`?

ericniebler:
> A range that repeats the last value written into it sounds like it could be useful - latch? - but such a thing is a non-`View`.

Why? Cyclic ranges can be views. Mutable ranges can be views. And its copy constructor would be O(1). What is it missing?

CaseyCarter:
> What is it missing?

Nothing given our current definition of the `View` concept; I have a hunch that `View`s should not participate in element ownership / lifetime in the same way that forward *iterators* should not (and do not, in C++14) participate in element ownership / lifetime. Or perhaps only forward `View`s, which would admit the input range `latch` as a `View`.

PR #388, opened 2016-06-04
Refactor compressed_pair and tagged:

### LINKED ISSUE #386: RFC: compressed_pair change
`compressed_pair` can only compress trivial types. That rules out a large class of empty types that *should* be compressed, most critically `default_sentinel`. I'd like to introduce a pure EBO mechanism "between" the "do nothing" and "use a single static" modes for `compressed_pair`, but doing so will require switching from `.first` notation to `.first()` notation everywhere that `compressed_pair` is used.

Comments?

CaseyCarter:
`compressed_pair`:
* Implement `compressed_pair`'s notion that multiple empty+trivial objects can be coalesced into a single object in `box`. Guard `box`'s EBO specialization against types that could be lambda closures when the compiler is GCC 6+.
* Implement `compressed_tuple` as a simple aggregation of `box`es.
* `compressed_pair` becomes essentially `tagged<compressed_tuple<tag::first(First), tag::second(Second)>`; all `.first` and `.second` member references become `.first()` and `.second()`.
* Replace `iterator_range`'s `tagged_pair` base with `tagged_compressed_tuple`.
* Replace `replacer_if_fn`'s `compressed_pair` member with a base.

`tagged` refactoring:
* Ensure that `tagged's conversions are only used for conversion.
* Construct by getter chain inheritance instead of fan inheritance, so that getters hide members of the same name in the base type.
* Use ADL to find `get` in tagged getters.
* Convert `tagged` to base before calling `get`.

Fixes #386.

### DIFF
--- a/include/range/v3/iterator_range.hpp
+++ b/include/range/v3/iterator_range.hpp
@@ -35,11 +35,11 @@ namespace ranges
         /// @{
         template<typename I, typename S /*= I*/>
         struct iterator_range
-          : tagged_pair<tag::begin(I), tag::end(S)>
+          : tagged_compressed_tuple<tag::begin(I), tag::end(S)>
           , view_interface<iterator_range<I, S>>
         {
         private:
-            using base_t_ = tagged_pair<tag::begin(I), tag::end(S)>;
+            using base_t = tagged_compressed_tuple<tag::begin(I), tag::end(S)>;
         public:
             using iterator = I;
             using sentinel = S;
@@ -47,33 +47,36 @@ namespace ranges
             using const_iterator = I; // Mostly to avoid spurious errors in Boost.Range
         #endif
 
+            using base_t::begin;
+            using base_t::end;
+
             iterator_range() = default;
             constexpr iterator_range(I begin, S end)
-              : base_t_{detail::move(begin), detail::move(end)}
+              : base_t{detail::move(begin), detail::move(end)}
             {}
             template<typename X, typename Y,
-                CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
+                CONCEPT_REQUIRES_(Constructible<I, X>() && Constructible<S, Y>())>
             constexpr iterator_range(iterator_range<X, Y> rng)
-              : base_t_{detail::move(rng.first), detail::move(rng.second)}
+              : base_t{detail::move(rng.begin()), detail::move(rng.end())}
             {}
             template<typename X, typename Y,
-                CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
-            constexpr iterator_range(std::pair<X, Y> rng)
-              : base_t_{detail::move(rng.first), detail::move(rng.second)}
+                CONCEPT_REQUIRES_(Constructible<I, X>() && Constructible<S, Y>())>
+            explicit constexpr iterator_range(std::pair<X, Y> rng)
+              : base_t{detail::move(rng.first), detail::move(rng.second)}
             {}
             template<typename X, typename Y,
-                CONCEPT_REQUIRES_(Assignable<I &, X &&>() && Assignable<S &, Y &&>())>
+                CONCEPT_REQUIRES_(Assignable<I &, X>() && Assignable<S &, Y>())>
             iterator_range &operator=(iterator_range<X, Y> rng)
             {
-                this->first = detail::move(rng).first;
-                this->second = detail::move(rng).second;
+                begin() = detail::move(rng).begin();
+                end() = detail::move(rng).end();
                 return *this;
             }
             template<typename X, typename Y,
                 CONCEPT_REQUIRES_(ConvertibleTo<I, X>() && ConvertibleTo<S, Y>())>
             constexpr operator std::pair<X, Y>() const
             {
-                return {this->first, this->second};
+                return {begin(), end()};
             }
         };
 
@@ -116,12 +119,12 @@ namespace ranges
             template<typename X, typename Y,
                 CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
             RANGES_NDEBUG_CONSTEXPR sized_iterator_range(iterator_range<X, Y> rng, iterator_size_t<I> size)
-              : sized_iterator_range{detail::move(rng).first, detail::move(rng).second, size}
+              : sized_iterator_range{detail::move(rng).first(), detail::move(rng).second, size}
             {}
             template<typename X, typename Y,
                 CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
             RANGES_NDEBUG_CONSTEXPR sized_iterator_range(sized_iterator_range<X, Y> rng)
-              : sized_iterator_range{rng.begin(), rng.end(), rng.size_}
+              : sized_iterator_range{detail::move(rng).rng_.first(), detail::move(rng).rng_.second, rng.size_}
             {}
             template<typename X, typename Y,
                 CONCEPT_REQUIRES_(Assignable<I &, X &&>() && Assignable<S &, Y &&>())>
@@ -133,11 +136,11 @@ namespace ranges
             }
             I begin() const
             {
-                return rng_.first;
+                return rng_.begin();
             }
             S end() const
             {
-                return rng_.second;
+                return rng_.end();
             }
             iterator_size_t<I> size() const
             {
@@ -188,30 +191,6 @@ namespace ranges
         {
             constexpr auto&& make_iterator_range = static_const<make_iterator_range_fn>::value;
         }
-
-        /// Tuple-like access for `iterator_range`
-        template<std::size_t N, typename I, typename S>
-        constexpr auto get(iterator_range<I, S> & p)
-        RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
-        (
-            ranges::get<N>(static_cast<tagged_pair<tag::begin(I), tag::end(S)> &>(p))
-        )
-
-        /// \overload
-        template<std::size_t N, typename I, typename S>
-        constexpr auto get(iterator_range<I, S> const & p)
-        RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
-        (
-            ranges::get<N>(static_cast<tagged_pair<tag::begin(I), tag::end(S)> &>(p))
-        )
-
-        /// \overload
-        template<std::size_t N, typename I, typename S>
-        constexpr auto get(iterator_range<I, S> && p)
-        RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
-        (
-            ranges::get<N>(static_cast<tagged_pair<tag::begin(I), tag::end(S)> &&>(p))
-        )
 
         /// Tuple-like access for `sized_iterator_range`
         template<std::size_t N, typename I, typename S,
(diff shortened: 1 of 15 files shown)


CaseyCarter:
Verified this with clang 3.7 & 3.8 c++11 & 14 using libc++. There's a problem I still need to run down with clang + libstdc++, but what's here is close enough for code review.

CaseyCarter:
> clang + libstdc++

Issue is that clang 3.7 can't grok the libstdc++ 6.1 headers. Everything is kosher for clang3.7 with libstdc++ 4.9.3 & 5.3.1. That's good enough for me.

include/range/v3/iterator_range.hpp
```diff
@@ -35,45 +35,48 @@ namespace ranges
...
+            using base_t::end;
+
             iterator_range() = default;
             constexpr iterator_range(I begin, S end)
-              : base_t_{detail::move(begin), detail::move(end)}
+              : base_t{detail::move(begin), detail::move(end)}
             {}
             template<typename X, typename Y,
-                CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
+                CONCEPT_REQUIRES_(Constructible<I, X>() && Constructible<S, Y>())>
```
ericniebler: Are you sure?

include/range/v3/iterator_range.hpp
```diff
@@ -189,30 +192,6 @@ namespace ranges
...
-        )
-
-        /// \overload
-        template<std::size_t N, typename I, typename S>
-        constexpr auto get(iterator_range<I, S> && p)
-        RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
-        (
-            ranges::get<N>(static_cast<tagged_pair<tag::begin(I), tag::end(S)> &&>(p))
-        )
-
```
ericniebler: I feel like there was a reason why I added these instead of relying on the base-to-derived conversion. :-P

include/range/v3/utility/box.hpp
```diff
@@ -129,21 +130,73 @@ namespace ranges
...
 
-        template<typename Element, typename Tag = Element,
-            bool Empty = std::is_empty<Element>::value && !detail::is_final<Element>::value>
-        struct box
+        /// \cond
+        namespace detail
         {
-        private:
-            Element value;
+            enum class box_compress { none, ebo, coalesce };
```
ericniebler: A comment here about what these mean would be helpful.

include/range/v3/utility/box.hpp
```diff
@@ -191,45 +249,73 @@ namespace ranges
             }
         };
 
-        template<typename Element, typename Tag = Element, bool Inherit = true>
-        using box_if =
-            box<Element, Tag, Inherit && std::is_empty<Element>::value &&
-                                         !detail::is_final<Element>::value>;
+        template<typename Element, typename Tag>
+        class box<Element, Tag, detail::box_compress::coalesce>
```
ericniebler: If we're using EBO, when do we benefit from this implementation?

include/range/v3/utility/compressed_pair.hpp
```diff
@@ -14,110 +15,118 @@
...
-                T first;
-                first_base() = default;
-                template<typename U,
-                    meta::if_<std::is_constructible<T, U &&>, int> = 0>
-                constexpr explicit first_base(U && u)
-                  : first((U &&) u)
-                {}
-            };
+                static_assert(Same<meta::index_sequence<Is...>,
+                    meta::make_index_sequence<sizeof...(Is)>>(), "What madness is this?!?");
```
ericniebler: Nice.

include/range/v3/utility/compressed_pair.hpp
```diff
@@ -14,110 +15,118 @@
...
-                static T first;
-                first_base() = default;
-                template<typename U,
-                    meta::if_<std::is_constructible<T, U &&>, int> = 0>
-                constexpr explicit first_base(U &&)
+                compressed_tuple_() = default;
+
+                // TODO: EXPLICIT
+                template<typename... Args,
+                    meta::if_<meta::and_<std::is_constructible<Ts, Args>...>, int> = 0>
```
ericniebler: Prefer `strict_and` where you don't need short-circuiting.

include/range/v3/utility/compressed_pair.hpp
```diff
@@ -14,110 +15,118 @@
...
+            tagged<compressed_tuple<detail::tag_elem<Ts>...>, detail::tag_spec<Ts>...>;
+
+        RANGES_DEFINE_TAG_SPECIFIER(first)
+        RANGES_DEFINE_TAG_SPECIFIER(second)
 
         template<typename First, typename Second>
         struct compressed_pair
-          : private detail::first_base<First>
-          , private detail::second_base<Second>
+            : tagged_compressed_tuple<tag::first(First), tag::second(Second)>
```
ericniebler: 👍

include/range/v3/utility/functional.hpp
```diff
@@ -458,30 +440,36 @@ namespace ranges
...
-            (
-                detail::as_const(overloaded::compressed_pair::second)(
-                    detail::forward<Args>(args)...)
-            )
+            RANGES_DECLTYPE_NOEXCEPT(
+                std::declval<overloaded<Rest...> const &>()(
+                    detail::forward<Args>(args)...))
+            {
+                return this->second()(detail::forward<Args>(args)...);
+            }
```
ericniebler: Is this ugliness to avoid an MSVC bug?

include/range/v3/utility/tagged_pair.hpp
```diff
@@ -14,175 +15,196 @@
...
+                constexpr                                                            \
+                meta::_t<std::tuple_element<I, Untagged>> const &NAME() const &      \
+                    noexcept(noexcept(                                               \
+                        detail::adl_get<I>(std::declval<Untagged const &>())))       \
+                {                                                                    \
+                    return detail::adl_get<I>(static_cast<Untagged const &>(*this)); \
+                }                                                                    \
+            };                                                                       \
+        };                                                                           \
+    }                                                                                \
```
ericniebler: Is the purpose of this extensive `tagged` rewrite to take better advantage of EBO? What was the old implementation getting wrong? Should we update the Ranges TS?

include/range/v3/view_adaptor.hpp
```diff
@@ -316,24 +331,29 @@ namespace ranges
...
-            template<typename I, typename IA, typename A = Adapt,
-                typename R = decltype(std::declval<A>().empty(std::declval<I>(), first))>
+            template<typename I, typename IA, typename A = Adapt, typename = decltype(
+                std::declval<A const &>().empty(
+                    std::declval<I const &>(),
+                    std::declval<BaseSent const &>()))>
             constexpr bool equal_(adaptor_cursor<I, IA> const &that, long) const
             {
-                return second.empty(that.first, first);
+                return second().empty(that.first(), first());
```
ericniebler: I remember being nauseous when I wrote this code and having every intention to come back to it and make it suck less. Now it sucks more (which is the nature of sucky code under maintenance). I welcome any thoughts you might have. `view_adaptor` in general has always been a problematic design issue.

test/utility/basic_iterator.cpp
```diff
@@ -359,6 +357,23 @@ namespace test_forward_sized
...
+void test_box()
+{
+    struct A : ranges::box<int> {};
+    CHECK(sizeof(A) == sizeof(int));
+    struct empty {};
+    struct B : ranges::box<empty> { int i; };
+    CHECK(sizeof(B) == sizeof(int));
+    B b1, b2;
+    if (ranges::detail::box_compression<empty>() == ranges::detail::box_compress::coalesce)
+        CHECK((&b1.get() == &b2.get()));
```
ericniebler: I'm guessing this test can't pass on MSVC.

include/range/v3/iterator_range.hpp
CaseyCarter: `std::is_constructible<T, Args...>` and `ranges::Constructible<T, Args...>` are defined in terms of `std::declval<Args>()`,  which is an lvalue when `Args` is an lvalue reference type and an xvalue otherwise. The `&&` has no effect.

The rule holds for `std::experimental::ranges::Constructible` as well, albeit with a slightly different proof.

include/range/v3/iterator_range.hpp
CaseyCarter: Maybe the base was once private? I commented these out with the expectation of fixing them when a test broke, but no test ever did. If it makes you nervous I can put them back.

include/range/v3/utility/box.hpp
CaseyCarter: When `Element` is final, or in the - I imagine unlikely in real-world code - eventuality that some class with `n` bytes of non-empty representation multiply inherits the same empty base more than `n` times.

I think the marginal utility is low.

include/range/v3/utility/box.hpp
CaseyCarter: I'll add one if we decide not to axe "coalesce."

include/range/v3/utility/compressed_pair.hpp
CaseyCarter: Fixed here and in `compressed_pair`.

include/range/v3/utility/functional.hpp
CaseyCarter: GCC doesn't like `this` - and consequently member function accesses - in `noexcept` clauses [PR52869](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52869). 4.8 (and maybe 4.9?) don't like them in trailing return types, either.

Not sure why I used `this->member()` syntax here instead of `using base_type::member;`. It's probably fewer overall keystrokes, but inconsistent.

include/range/v3/utility/tagged_pair.hpp
CaseyCarter: Primarily:
* Use ADL for `get` so that non-standard types are taggable
* Inject the tagged getters between the top-level tagged type and the base type, instead of using sibling inheritance, so that tagged getters hide members of the same name inherited from the base instead of being ambiguous. (This wreaked all kinds of havoc with recursive composition of tagged types, which became fairly common with the `compressed_pair` change.)

CaseyCarter: > Should we update the Ranges TS?

Yes, although I think we will need a better design for `get` as a customization point.

include/range/v3/view_adaptor.hpp
CaseyCarter: I share your general distaste for `view_adaptor`, but haven't invested in thought into how to improve it. Rest assured I will speak up if and when I do.

include/range/v3/utility/tagged_pair.hpp
CaseyCarter: Updated ericniebler/stl2#172.

CaseyCarter:
Let me know if you want to keep/remove the coalescing variant of `box`, and I'll squash the commits.

include/range/v3/utility/box.hpp
ericniebler: > When `Element` is final

Eh. Keep it.

ericniebler:
> I'll squash the commits

Don't squash the Travis fix since that will cause a merge conflict. OK to squash the others.

CaseyCarter:
> Don't squash the Travis fix since that will cause a merge conflict. OK to squash the others.

I'll squash the *outstanding* commits.

PR #389, opened 2016-06-06
Tighten up warnings:

CaseyCarter:
* Remove several warning suppression flags from CMakeLists.txt

* Add three new targeted warning suppression macros in config.hpp:
  * `RANGES_DIAGNOSTIC_IGNORE_GLOBAL_CONSTRUCTORS`
  * `RANGES_DIAGNOSTIC_IGNORE_SIGN_CONVERSION`
  * `RANGES_DIAGNOSTIC_IGNORE_UNNEEDED_INTERNAL`
  These are used to suppress targeted warnings in the tests.

* We want to (a) use the test suite to catch questionable usage in library headers, (b) expose diagnostics to users who use the library in questionable ways. Disabling a large class of warnings in the library headers impeded both of those goals. Consequently, `RANGES_DISABLE_WARNINGS` and `RANGES_RE_ENABLE_WARNINGS` are now no-ops by default. This is a breaking change for code compiling with warnings-as-errors that may now diagnose errors that were suppressed before. Users who just don't care can define `RANGES_FEWER_WARNINGS` to get the old behavior.

* Suppress a couple of dozen sign-conversion and 64-to-32 truncation warnings. Many remain as exercises for the reader.

* In functional.hpp, implement class template `convert_to<T>`, a function object that returns its argument converted to `T`.

* **Relocate the code of some of the tests into an unnamed namespace since it does not need external linkage. This is mostly a style preference.** (This was a poor choice: it's *extremely* noisy, hiding the real changes.)

EDIT:
I should point out that I've validated this change builds equivalently with clang-3.5/6/7/8 Debug/Release C++14.


ericniebler:
> Don't force calendar.cpp to compile with -std=gnu++14. This is an annoyance when intentionally compiling with -std=c++1y, -std=c++1z, or a non-GCC-flavored compiler.

How are you making sure it doesn't build with -std=c++11, which is the default (is it not)?

ericniebler:
> In functional.hpp, implement class template `convert_to<T>`, a function object that returns its argument converted to `T`.

That's what `coerce` is for. Does it not meet your needs?

ericniebler:
> Relocate the code of some of the tests into an unnamed namespace since it does not need external linkage. This is mostly a style preference. (This was a poor choice: it's extremely noisy, hiding the real changes.)

Right, if I were more anal, I'd ask for those changes in a separate pull, but I don't want to waste your time, and it's not a big enough deal. In the future, though.

CaseyCarter:
> How are you making sure it doesn't build with -std=c++11, which is the default (is it not)?
calendar & sort_patterns are both protected by

ericniebler:
> calendar & sort_patterns are both guarded by

The current behavior is: I can build everything with the defaults, and I can get tests built in C++11 and calendar built in C++14. How do I get that behavior with this patch?

CaseyCarter:
> That's what `coerce` is for. Does it not meet your needs?

`convert_to<T>(foo)` is roughly `DECAY_COPY(coerce<T>(foo))`, except that it doesn't interfere with copy elision and it will copy an rvalue `T const`.

CaseyCarter:
> The current behavior is: I can build everything with the defaults, and I can get tests built in C++11 and calendar built in C++14. How do I get that behavior with this patch?

You won't get calendar at all with the defaults, you need to define `RANGES_BUILD_CALENDAR_EXAMPLE` ;) Serious answer: this isn't possible without editing CMakeLists.txt. Without the change, it's similarly impossible to build calendar with C++1z. I find it more convenient to have separate builds set up for distinct configurations, so I would achieve your desired behavior with "make && make -C ../build-c++14 calendar".

ericniebler:
> Without the change, it's similarly impossible to build calendar with C++1z

Calendar doesn't _need_ C++1z, so I don't care about that. It _does_ need C++14, and the rest of the library should be tested in C++11 mode. That's the behavior I want.

CaseyCarter:
> That's the behavior I want.

Done.

ericniebler:
Great, thanks!

PR #390, opened 2016-06-06
A random-access sized range can be bounded in O(1)

CaseyCarter:


include/range/v3/view/bounded.hpp
```diff
@@ -89,12 +95,20 @@ namespace ranges
...
                     return bounded_view<all_t<Rng>>{all(std::forward<Rng>(rng))};
                 }
                 template<typename Rng,
+                    CONCEPT_REQUIRES_(Range<Rng>() && !BoundedRange<Rng>() && RandomAccessRange<Rng>() && SizedRange<Rng>())>
+                iterator_range<range_iterator_t<Rng>> operator()(Rng && rng) const
+                {
+                    auto first = ranges::begin(rng);
+                    auto last = first + ranges::size(rng);
+                    return iterator_range<range_iterator_t<Rng>>{detail::move(first), detail::move(last)};
+                }
```
ericniebler: This returns dangling iterators if `Rng` is an rvalue.

include/range/v3/view/bounded.hpp
```diff
@@ -89,12 +95,24 @@ namespace ranges
...
+                    CONCEPT_REQUIRES_(Range<Rng>() && !BoundedRange<Rng>() &&
+                        RandomAccessRange<Rng>() && SizedRange<Rng>())>
+                iterator_range<range_iterator_t<all_t<Rng>>> operator()(Rng && rng) const
+                {
+                    auto first = ranges::begin(rng);
+                    auto last = first + ranges::distance(rng);
+                    return iterator_range<range_iterator_t<all_t<Rng>>>{
+                        detail::move(first), detail::move(last)
+                    };
+                }
```
ericniebler: This still returns dangling iterators.

CaseyCarter: This overload does not participate in resolution when `Rng` is an rvalue non-`View`, since `all_t<Rng>` is ill-formed for rvalue non-`View` ranges.

CaseyCarter: Oh, I'm being thick. `View`s can have dangling iterators, too.

PR #397, opened 2016-06-08
[concat] iterator is EqualityComparable only if the base range iterators are.

### LINKED ISSUE #395: Combining concat, generate, and take_while results in a non-range
See the following code in which the begin and end iterators do not satisfy `WeaklyEqualityComparable`:
```cpp
int i = 0;
auto rng = ranges::view::concat(ranges::view::generate([&] { return i++; }))
    | ranges::view::take_while([](int i2) { return i2 < 30; });
auto failsToCompile = ranges::begin(rng) != ranges::end(rng); // fails to compile
```

CaseyCarter:
Fixes #395.

### DIFF
--- a/include/range/v3/view/concat.hpp
+++ b/include/range/v3/view/concat.hpp
@@ -236,6 +236,7 @@ namespace ranges
                 {
                     its_.visit_i(next_fun{this});
                 }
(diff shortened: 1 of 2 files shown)


ericniebler:
Awesome, thanks!

PR #399, opened 2016-06-12
Implement a few random things.

CaseyCarter:
* Relocate URNG machinery to utility/random.hpp
* Steal some unpredictable URNG seeding code from the P0347 authors.
* Implement algorithm `sample` from the C++17 WP
* Implement `sample_view`

`sample` has two required implementations: a stable variant for `Forward` source ranges, and an unstable variant for `Input` source ranges with `RandomAccess` result ranges. The stable variant only requires the source range to be `Forward` because it needs to calculate the size. Range-v3 `Input` ranges sometimes know their own size, so range-v3 `sample` uses the stable implementation when the source range is `Forward` or `Sized`.

### DIFF
--- a/include/range/v3/algorithm/sample.hpp
+++ b/include/range/v3/algorithm/sample.hpp
@@ -21,11 +21,11 @@
 #include <range/v3/range_traits.hpp>
 #include <range/v3/algorithm/min.hpp>
 #include <range/v3/utility/random.hpp>
-#include <range/v3/utility/swap.hpp>
 #include <range/v3/utility/iterator.hpp>
 #include <range/v3/utility/iterator_concepts.hpp>
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/static_const.hpp>
+#include <range/v3/utility/tagged_pair.hpp>
 
 namespace ranges
 {
@@ -45,7 +45,8 @@ namespace ranges
             template<typename I, typename S, typename O,
                 typename Gen = detail::default_random_engine&,
                 CONCEPT_REQUIRES_(Constraint<I, S, O, Gen>())>
-            static O sized_impl(I first, S last, iterator_difference_t<I> pop_size,
+            static tagged_pair<tag::in(I), tag::out(O)>
+            sized_impl(I first, S last, iterator_difference_t<I> pop_size,
                 O out, iterator_difference_t<I> n, Gen && gen)
             {
                 std::uniform_int_distribution<iterator_difference_t<I>> dist;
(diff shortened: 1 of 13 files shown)


CaseyCarter:
Fixing merge conflicts...

include/range/v3/algorithm/sample.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+            template<typename I, typename S, typename O,
+                typename Gen = detail::default_random_engine&,
+                CONCEPT_REQUIRES_(Constraint<I, S, O, Gen>())>
+            static O sized_impl(I first, S last, iterator_difference_t<I> pop_size,
+                O out, iterator_difference_t<I> n, Gen && gen)
+            {
+                std::uniform_int_distribution<iterator_difference_t<I>> dist;
+                using param_t = typename decltype(dist)::param_type;
+                n = std::min(pop_size, n);
```
ericniebler: Hrm, we don't have `ranges::min`? You  should at least #include <algorithm>.

CaseyCarter: Yes, we have `ranges::min`. I wrote it ;)

ericniebler:
I haven't studied this algorithm in detail, but it seems that we're losing useful information by only returning the output iterator. Would the last position of the input iterator or the number of elements sampled be useful information?

CaseyCarter:
I was returning source range end position, but decided it wasn't useful since the source range isn't always fully traversed. However, the algorithm does take input iterators, it should return the end position so the user can choose to find end-of-range. I'll switch back.

I strongly suspect the user will know the sample size in nearly every use case, passing either a count or an output range with a known size. I therefore believe the sample size would not be useful. Note that I have (I, S, I, n), (I, S, R), (R, I, n) and (R, R) signatures - I didn't bother with (I, S, I, S) or (R, I, S) because I can't grok a use case "give me a sample of unknown size of this data set." If I'm wrong, we can always add it later. An enterprising user could easily write a "counting" output range and recover the sample size.

CaseyCarter:
Also, FWIW, I know just enough about random number generation to be aware that it's incredibly easy to screw up. I've run the auto seed sequence against SmallCrush from [TESTU01](http://www.iro.umontreal.ca/~simardr/testu01/tu01.html), using the auto seed sequence to repeatedly initialize a block of 624 uint32s and returning them one-at-a-time, and it passes all tests even with the `random_device` replaced with a zero filling function.

erikerlandson:
I think these might be accelerated with "gap sampling":
http://erikerlandson.github.io/blog/2014/09/11/faster-random-samples-with-gap-sampling/

PR #401, opened 2016-06-12
Replace non-test usage of std::min/max/minmax with ranges::min/max/minmax

CaseyCarter:


include/range/v3/view/transform.hpp
```diff
@@ -287,13 +288,17 @@ namespace ranges
                 SizedRange<Rng1 const>() && SizedRange<Rng2 const>())
             constexpr size_type_ size() const
             {
-                return std::min<size_type_>(ranges::size(rng1_), ranges::size(rng2_));
+                return ranges::on(ranges::min,
+                    ranges::compose(convert_to<size_type_>{}, ranges::size))(
+                        rng1_, rng2_);
```
ericniebler: Fun, but it would probably be simpler without `on` and `compose`.

CaseyCarter: Would you prefer:
```c++
auto f = compose(convert_to<size_type_>{}, ranges::size);
return ranges::min(f(rng1_), f(rng2_));
```
or
```c++
return ranges::min(
    static_cast<size_type_>(ranges::size(rng1_)),
    static_cast<size_type_>(ranges::size(rng2_)));
```
It's a shame that we can't have both `ranges::min(x, y)` and `ranges::min<T>(x, y)` be valid syntax when `ranges::min` is an object. I wonder if anyone has poked EWG about making:
```c++
template <class> class some_class_template;
template<class T = int>
some_class_template<T> foo{};

auto* x = &foo;
```
well-formed.

gnzlbg: > It's a shame that we can't have both ranges::min(x, y) and ranges::min<T>(x,
y) be valid syntax when ranges::min is an object

Indeed, having to write ranges::min.operator()<T>(x, y) is just painful. It
would also be nice if operator() could be static ... I never understood
this restriction.

ericniebler: This:

```
return ranges::min(
    static_cast<size_type_>(ranges::size(rng1_)),
    static_cast<size_type_>(ranges::size(rng2_)));
```

> It's a shame that we can't have both `ranges::min(x, y)` and `ranges::min<T>(x, y)` be valid syntax when `ranges::min` is an object.

Agreed. :-( Also a shame that we have to type `std::equal_to<>{}` instead of simply `std::equal_to{}`. I wonder if the goofy work on class-template-parameter-deduction-from-constructor-arguments helps here any.

include/range/v3/view/transform.hpp
```diff
@@ -16,15 +16,16 @@
...
 #include <type_traits>
 #include <meta/meta.hpp>
 #include <range/v3/range_fwd.hpp>
 #include <range/v3/size.hpp>
 #include <range/v3/begin_end.hpp>
 #include <range/v3/range_traits.hpp>
 #include <range/v3/view_adaptor.hpp>
+#include <range/v3/algorithm/max.hpp>
+#include <range/v3/algorithm/min.hpp>
+#include <range/v3/utility/functional.hpp>
```
ericniebler: Do we still need to include #include functional.hpp?

CaseyCarter: I'm sure we don't need to include it _twice_.

CaseyCarter: `function_type`, `Callable`, ... yes we need to include functional.hpp.

include/range/v3/view/transform.hpp
CaseyCarter: > Agreed. :-( Also a shame that we have to type std::equal_to<>{} instead of simply std::equal_to{}. I wonder if the goofy work on class-template-parameter-deduction-from-constructor-arguments helps here any.

[P0091](http://wg21.link/p0091) is purely focused on deduction for class template constructors. The wording is minimal, but it looks to me like `std::equal_to{}` will deduce `void`. It won't help with our `ranges::min<foo>` problem.

PR #405, opened 2016-06-14
Shuffle: single-pass algorithm

CaseyCarter:
Fairly self-explanatory. Performance is equivalent to the "old" implementation when `is_same<S, I>()`, the "old" implementation's best case.


ericniebler:
Nice, thanks.

PR #408, opened 2016-06-15
single-pass implementations of the partition_point family

CaseyCarter:
* New algorithm: `aux::partition_point_n`
* `partition_point` exponentially probes for either end-of-range or a value that is past the partition point (doesn't satisfy the predicate), then defers to `partition_point_n` with the narrowed range.
* `lower_bound{_n}` and `upper_bound{_n}` are implemented in terms of `partition_point{_n}`.
* `equal_range` exponentially probes for:
  * end-of range or an iterator that is past-the-end of the target range (denotes an element greater than the desired value), at which point it defers to `equal_range_n`.
  * an iterator inside the target range (denotes an element equal to the desired value), at which point it delegates to `lower_bound_n(stuff-before-the-iterator)` and `upper_bound(stuff-after)`.
* Fix projection bug in `equal_range_n`; add test coverage.

### DIFF
--- a/include/range/v3/algorithm/equal_range.hpp
+++ b/include/range/v3/algorithm/equal_range.hpp
@@ -2,6 +2,7 @@
 // Range v3 library
 //
 //  Copyright Eric Niebler 2014
+//  Copyright Casey Carter 2016
 //
 //  Use, modification and distribution is subject to the
 //  Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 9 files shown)


include/range/v3/algorithm/equal_range.hpp
```diff
@@ -31,24 +35,86 @@ namespace ranges
...
             iterator_range<I>
-            operator()(I begin, S end, V const & val, C pred = C{}, P proj = P{}) const
+            operator()(I begin, S end, V const &val, C pred_ = C{}, P proj_ = P{}) const
+            {
+                // Probe exponentially for either end-of-range, an iterator that
+                // is past the equal range (i.e., denotes an element greater
+                // than val), or is in the equal range (denotes an element equal
+                // to val).
+                auto pred = as_function(pred_);
+                auto proj = as_function(proj_);
```
ericniebler: These should be `auto&&`. Don't force a copy if we don't have to.

CaseyCarter: Not sure how I missed that. Also in partition_point

PR #410, opened 2016-06-15
Kill default_random_engine on IPhone.

### LINKED ISSUE #407: Thread-local storage is not supported on iOS (__IPHONE_OS_VERSION_MIN_REQUIRED <= 70100)
This error comes from `random.hpp` when `__IPHONE_OS_VERSION_MIN_REQUIRED <= 70100`:
```cpp
error: thread-local storage is not supported for the current target
                static __thread bool initialized = false;
                       ^
error: thread-local storage is not supported for the current target
                static __thread meta::_t<std::aligned_storage<
                       ^
```
`boost::thread_specific_ptr` works on this platform, but it might be a lot of work to port it. Since my company has no use for the default random engine, just having a way to opt out of TLS would be adequate.

CaseyCarter:
I don't have a development environment to test a proper fix, so someone else can fix this. Or complain that it's broken.
Closes #407.

### DIFF
--- a/include/range/v3/utility/random.hpp
+++ b/include/range/v3/utility/random.hpp
@@ -51,6 +51,7 @@
 #include <chrono>
 #include <functional>  // for std::hash
 #include <initializer_list>
+#include <mutex>
 #include <new>
 #include <random>
 #include <thread>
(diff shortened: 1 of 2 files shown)


include/range/v3/utility/random.hpp
```diff
@@ -495,13 +495,31 @@ namespace ranges
...
+            default_random_engine& get_random_engine()
             {
+#ifdef TARGET_OS_IPHONE
+                static_assert(meta::invoke<meta::id<std::false_type>, T>::value,
+                    "IPhone did not support C++11 thread_local when I wrote this. "
+                    "If it does now, or you want to contribute a workourand, then "
+                    "by all means contribute a patch / complain in the range-v3 "
+                    "forum. Otherwise, you should avoid calling functions that "
+                    "accept an URNG without explicitly specifying an URNG.");
+                std::terminate();
```
ericniebler: My preference in this case would be to have a single global engine and synchronize access to it. Unceremoniously dumping core is not a nice thing to do.

CaseyCarter: There's no core dumping; the `static_assert` prevents instantiation. The `terminate` call is there only to silence the "no return in a function that needs one" warning.

Silently serializing multithreaded code is not nice, either. I'll throw in a `#pragma message` with some words of warning.

CaseyCarter:
There's a new pull with a mutex-wrapped URNG for systems with borked thread_local. I'm not terribly proud of the solution, but I don't think there's demand to justify anything more complicated.

ericniebler:
Works for me. Thanks.

PR #414, opened 2016-06-20
Implement drop a la take, instead of take_exactly.

### LINKED ISSUE #413: crashes at runtime while constructing a vector from a view with nested transforms
The following code crashes at runtime in `skips` while constructing a `vector`:
```cpp
#include <iostream>
#include <range/v3/all.hpp>

using namespace ranges::view;

std::vector<std::vector<int>> skips(std::vector<int> xs) {
    return ints(0, (int)xs.size())
        | transform([&](int n) {
            return xs | chunk(n + 1)
                      | transform(drop(n))
                      | join;
        });
}

int main() {
    auto skipped = skips({1,2,3,4,5,6,7,8});
    return 0;
}
```

If you change `transform(drop(n))` to just `drop(n)`, it compiles and runs.

Compiled with `Apple LLVM version 7.3.0 (clang-703.0.31)`

[coliru](http://coliru.stacked-crooked.com/a/e3aae50976021de0)

CaseyCarter:
i.e., `drop(rng, n)` is an empty range if `rng` has less than `n` elements instead of a precondition violation.

Fixes #413.


ericniebler:
Please also add drop_exactly, which will be faster. You'll also need to update the docs, such as they are.

CaseyCarter:
Done. `action::drop` already has "at most n" semantics, so I've left it untouched.

Also, I have no idea what the procedure is for regenerating the docs. I updated index.md, and I can run doxygen to generate everything, but I'm not certain how the magic happens with the gh-pages branch.

PR #423, opened 2016-07-05
Use inline variables when available

gnzlbg:
- adds RANGES_INLINE_VARIABLE(type, name); macro to declare inline variables
- deprecates static_const utility on all C++ versions
  - clients might be using it in their own code => deprecate for a while
  - adds a detail::inline_variable_ utility equivalent to static_const that
    is used internally in range-v3 and that clients are not supposed to use

### DIFF
--- a/include/range/v3/detail/config.hpp
+++ b/include/range/v3/detail/config.hpp
@@ -282,14 +282,38 @@
 #define RANGES_NDEBUG_CONSTEXPR inline
 #endif
 
-// Workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70552
-#if defined(__GNUC__) && !defined(__clang__) && \
-    ((__GNUC__ == 4 && __GNUC_MINOR__ == 9 && __GNUC_PATCHLEVEL__ >= 4) || \
-     (__GNUC__ == 5 && __GNUC_MINOR__ >= 3))
-#define RANGES_GCC_BROKEN_CUSTPOINT inline
+#define RANGES_CXX_INLINE_VARIABLES_11 0
+#define RANGES_CXX_INLINE_VARIABLES_14 0
+#define RANGES_CXX_INLINE_VARIABLES_17 201606
+#ifndef RANGES_CXX_INLINE_VARIABLES
+
+#ifdef __cpp_inline_variables // TODO: fix this if SD-6 picks another name
+#define RANGES_CXX_INLINE_VARIABLES __cpp_inline_variables
+#elif defined(__clang__) && \
+    (__clang__major__ > 3 || __clang_major__ == 3 && __clang_minor__ == 9) && \
+    __cplusplus > 201402L
+// TODO: remove once clang defines __cpp_inline_variables (or equivalent)
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_INLINE_VARIABLES_17
 #else
-#define RANGES_GCC_BROKEN_CUSTPOINT
-#endif
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_FEATURE(INLINE_VARIABLES)
+#endif  // __cpp_inline_variables
+
+#endif  // RANGES_CXX_INLINE_VARIABLES
+
+#if RANGES_CXX_INLINE_VARIABLES < RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE(type, name)                  \
+    inline namespace                   \
+    {                                                       \
+        constexpr auto& name = static_const<type>::value;   \
+    }                                                     
+
+#else  // RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE(type, name) \
+    inline namespace function_objects             \
+    {                                      \
+        inline constexpr type name{};      \
+    }
+#endif // RANGES_CXX_INLINE_VARIABLES
 
 #ifdef RANGES_FEWER_WARNINGS
 #define RANGES_DISABLE_WARNINGS                 \
--- a/include/range/v3/utility/static_const.hpp
+++ b/include/range/v3/utility/static_const.hpp
@@ -20,6 +20,7 @@ namespace ranges
     inline namespace v3
     {
         /// \ingroup group-utility
+        
         template<typename T>
         struct static_const
         {
--- a/include/range/v3/action/action.hpp
+++ b/include/range/v3/action/action.hpp
@@ -55,10 +55,7 @@ namespace ranges
 
             /// \ingroup group-actions
             /// \relates make_action_fn
-            namespace
-            {
-                constexpr auto&& make_action = static_const<make_action_fn>::value;
-            }
+            RANGES_INLINE_VARIABLE(make_action_fn, make_action)
 
             template<typename Action>
             struct action : pipeable<action<Action>>
--- a/include/range/v3/action/drop.hpp
+++ b/include/range/v3/action/drop.hpp
@@ -88,10 +88,7 @@ namespace ranges
             /// \ingroup group-actions
             /// \relates drop_fn
             /// \sa action
-            namespace
-            {
-                constexpr auto&& drop = static_const<action<drop_fn>>::value;
-            }
+            RANGES_INLINE_VARIABLE(action<drop_fn>, drop)
         }
         /// @}
     }
--- a/include/range/v3/action/drop_while.hpp
+++ b/include/range/v3/action/drop_while.hpp
@@ -92,10 +92,7 @@ namespace ranges
             /// \ingroup group-actions
             /// \relates drop_while_fn
             /// \sa action
-            namespace
-            {
-                constexpr auto&& drop_while = static_const<action<drop_while_fn>>::value;
-            }
+            RANGES_INLINE_VARIABLE(action<drop_while_fn>, drop_while)
         }
         /// @}
     }
--- a/include/range/v3/action/erase.hpp
+++ b/include/range/v3/action/erase.hpp
@@ -67,10 +67,7 @@ namespace ranges
         /// \endcond
 
         /// \ingroup group-actions
-        namespace
-        {
-            constexpr auto&& erase = static_const<adl_erase_detail::erase_fn>::value;
-        }
+        RANGES_INLINE_VARIABLE(adl_erase_detail::erase_fn, erase)
 
         namespace action
         {
--- a/include/range/v3/action/insert.hpp
+++ b/include/range/v3/action/insert.hpp
@@ -230,10 +230,7 @@ namespace ranges
         /// \endcond
 
         /// \ingroup group-actions
-        namespace
-        {
-            constexpr auto&& insert = static_const<adl_insert_detail::insert_fn>::value;
-        }
+        RANGES_INLINE_VARIABLE(adl_insert_detail::insert_fn, insert)
 
         namespace action
         {
--- a/include/range/v3/action/join.hpp
+++ b/include/range/v3/action/join.hpp
@@ -79,10 +79,7 @@ namespace ranges
             /// \ingroup group-actions
             /// \relates join_fn
             /// \sa action
-            namespace
-            {
-                constexpr auto&& join = static_const<action<join_fn>>::value;
-            }
+            RANGES_INLINE_VARIABLE(action<join_fn>, join)
         }
         /// @}
     }
--- a/include/range/v3/action/push_back.hpp
+++ b/include/range/v3/action/push_back.hpp
@@ -104,10 +104,8 @@ namespace ranges
         {
             /// \ingroup group-actions
             /// \sa with_braced_init_args
-            namespace
-            {
-                constexpr auto&& push_back = static_const<with_braced_init_args<action<adl_push_back_detail::push_back_fn>>>::value;
-            }
+            RANGES_INLINE_VARIABLE(with_braced_init_args<action<adl_push_back_detail::push_back_fn>>,
+                                   push_back)
         }
 
         using action::push_back;
--- a/include/range/v3/action/push_front.hpp
+++ b/include/range/v3/action/push_front.hpp
@@ -104,10 +104,8 @@ namespace ranges
         {
             /// \ingroup group-actions
             /// \sa with_braced_init_args
-            namespace
-            {
-                constexpr auto&& push_front = static_const<with_braced_init_args<action<adl_push_front_detail::push_front_fn>>>::value;
-            }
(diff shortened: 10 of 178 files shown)


gnzlbg:
I still need to figure out what is going with `swap`, see [here](https://github.com/ericniebler/range-v3/pull/423/files#diff-aa7db2aba919d70d7304855522c7d270R260). 

Basically the whole library compiles fine if I put the inline variable in the `v3` namespace, except for the header `/utility/any.hpp`, where `any` defines a friend function `swap` which collides with the inline variable if they are in the same namespace (because one is a function, and the other one is a variable).

include/range/v3/utility/inline_variable.hpp
```diff
@@ -0,0 +1,38 @@
...
+            };
+
+            template<typename T>
+            constexpr T inline_variable_<T>::value;
+        }
+    }
+}
+#endif  // RANGES_CXX_INLINE_VARIABLES
+
+#endif
```
ericniebler: This is a utility for building static const objects, not inline variables. The name was right before.

ericniebler:
Sounds like a clang bug. IIUC, inline variables shouldn't be changing name lookup. Have you reported it?

include/range/v3/utility/static_const.hpp
```diff
@@ -14,14 +14,18 @@
...
 {
     inline namespace v3
     {
         /// \ingroup group-utility
+        
         template<typename T>
-        struct static_const
+        struct
+        RANGES_DEPRECATED("Use RANGES_CXX_INLINE_VARIABLE instead which supports C++17 inline variables")
+        static_const
```
ericniebler: Suggest undeprecating this and removing `inline_variable_`.

gnzlbg:
> Sounds like a clang bug. IIUC, inline variables shouldn't be changing name lookup. Have you reported it?

I don't think it is a bug. It actually makes sense to me that the following fails to compile:

```c++
void foo();
constexpr int foo = 0; // ERROR: foo re-declared as a different kind of symbol
```

The following doesn't fail, but I kind of think it should

```c++
void foo();
namespace {
  constexpr int foo = 0;  // OK: foo's are "technically" in different namespaces.
}
// which foo do you mean here?
```

include/range/v3/utility/static_const.hpp
gnzlbg: Sorry about that, I didn't knew if `static_const` should be deprecated or not.  Thought it might be good to alert users that they could be using the macro instead, but `static_const` works just fine.

ericniebler:
> It actually makes sense to me that the following fails to compile:
> 
> ```
> void foo();
> constexpr int foo = 0; // ERROR: foo re-declared as a different kind of symbol
> ```

But that's not the situation. We have:

```lang=c++
struct S {
  friend void foo();
};
constexpr int foo = 0;
```

There should be nothing wrong with that.

gnzlbg:
> There should be nothing wrong with that.

But `friend void foo();` just puts `foo` in `S` namespace, so that collides as well (this already happens without inline variables, `static int foo` is enough to make that fail to compile).

I removed the `inline_variable` thing, `static_const` is back to normal.

ericniebler:
> But `friend void foo();` just puts `foo` in `S` namespace

But this program:

```lang=c++
struct S {
    friend void foo() {}
};

int main() {
    foo();
}
```

Gives me: 

```
error: 'foo' was not declared in this scope
     foo();
         ^
```

I thought friend name injection stopped being a thing a long time ago?

CaseyCarter:
`hidden` friends are not found by non-ADL lookups, but the entities they name still exist, and C++ forbids two different entities with the same name unless they are both functions and/or function templates ([basic.scope.declarative]). This program, e.g.,:
```c++
struct S {
    friend void foo() {}
};

int foo;

int main() {}
```
Will blow up any compiler you point it at because `::foo` can't be both a (visible) object and a (hidden) function. 

[basic.scope.declarative]/1 seems to imply the opposite:

> Every name is introduced in some portion of program text called a *declarative region*, which is the largest part of the program in which that name is *valid*, that is, in which that name may be **used as an unqualified name** [emphasis added] to refer to the same entity. In general, each particular name is valid only within some possibly discontiguous portion of program text called its *scope*. To determine the scope of a declaration, it is sometimes convenient to refer to the *potential scope* of a declaration. The scope of a declaration is the same as its potential scope unless the potential scope contains another declaration of the same name. In that case, the potential scope of the declaration in the inner (contained) declarative region is excluded from the scope of the declaration in the outer (containing) declarative region.

Since I cannot refer to the function `foo` at the point in my program where I declare the integer `foo`, it would seem the name is not valid, and therefore that point is not part of its potential scope since no other declaration is present. I think this is a wording error, and that the intent is that the *potential scope* of so-called "hidden friends" is that portion of the program in which a matching declaration actually refers to the same entity. I.e., if the program redeclared `foo` in the global scope:
```c++
struct S {
    friend void foo() {}
};

void foo();

int main() { foo(); }
```
the formerly hidden friend becomes visible, and is in fact the same function.

The later text in [basic.scope.declarative] hints at this. Para 3:
> The names declared by a declaration are introduced into the scope in which the declaration occurs, except that the presence of a `friend` specifier (11.3), certain uses of the *elaborated-type-specifier* (7.1.6.3), and *using-directives* (7.3.4) alter this general behavior.

and the note after para 4:
> [ Note: These restrictions apply to the declarative region into which a name is introduced, which is not necessarily the same as the region in which the declaration occurs. In particular, *elaborated-type-specifiers* (7.1.6.3) and friend declarations (11.3) may introduce a (possibly not visible) name into an enclosing namespace; these restrictions apply to that region. ...

Our customization point design "cheats" around this rule by putting the object declaration inside a different namespace, accessed via a using directive (such as an unnamed namespace):
```c++
struct S {
    friend void foo() {}
};

namespace bob { int foo; }
using namespace bob;

int x = foo;

int main() {}
```
`::foo` and `bob::foo` are distinct entities yet all is right with the world. We are taking advantage of the fact that lookup for use of a name has differing properties from lookup to determine if a name is already defined. We should be able to achieve the same effect with an inline variable in a non-unnamed namespace. Clang trunk seems to like:
```c++
namespace ranges { inline namespace v3 {
struct S {
    friend void swap(S&, S&) {}
};

namespace CPOs { inline constexpr int swap = 42; }
using namespace CPOs;
}}

int main() { return ranges::swap; }
```

CaseyCarter:
I suggest something like:
```c++
namespace ranges
{
    inline namespace v3
    {
        namespace function_objects {}
        using namespace function_objects;
    }
}

#define RANGES_INLINE_VARIABLE(name, type) \
    namespace function_objects { inline constexpr type name{}; }
```

pfultz2:
> We should be able to achieve the same effect with an inline variable in a non-unnamed namespace. Clang trunk seems to like:

It seems like `swap` wouldn't have the same address across translation units due to the anonymous namespace.

CaseyCarter:
> It seems like `swap` wouldn't have the same address across translation units due to the anonymous namespace.

Apologies for using the ridiculously confusing phrase "non-unnamed namespace" - I meant a normal, named namespace like `CPOs` in the example.

pfultz2:
Also, you could try defining it like this:

```cpp
#if RANGES_CXX_INLINE_VARIABLES < RANGES_CXX_INLINE_VARIABLES_17
#   if defined(__GNUC__) || defined (__clang__)
#       define RANGES_INLINE_VAR extern __attribute__((weak))
#   elif defined(_MSC_VER)
#       define RANGES_INLINE_VAR extern __declspec(selectany)
#   else // Your compiler is hopeless
#       define RANGES_INLINE_VAR static
#   endif
#else
#    define RANGES_INLINE_VAR inline
#endif
```

Then just do:

```cpp
namespace function_objects { RANGES_INLINE_VAR constexpr adl_swap_detail::swap_fn swap{}; }
```

On gcc and clang this should keep ABI compatibility between C++17 and earlier versions, if that were to matter. Of course, `__declspec(selectany)` doesn't work with `constexpr` yet. However, it may not matter, since range-v3 doesn't work with MSVC yet either.

gnzlbg:
The last commit:

- adds `function_objects` namespaces and puts the inline variables in them per @CaseyCarter suggestion,
- the variables are marked `inline` since, as @CaseyCarter mentions in [clang bug 28385](https://llvm.org/bugs/show_bug.cgi?id=28385#c5), the approved "inline variables paper" is P0386R2 (not P0386R0 which is the only one available), and there `constexpr` doesn't always imply inline (as opposed to in P0386R0): it only implies inline for static data members (and not for variables in namespace scope).

I've tested this PR with the changes of PR #426 and everything works fine on clang-trunk in C++11/14/1z.

Still, I'd block merging this PR  till something like #426 gets merged and we start testing clang-trunk on travis again.

Question: I don't like this:

```c++
        namespace function_objects {}
        using namespace function_objects;

        namespace aux
        {
            namespace function_objects {}
            using namespace function_objects;
        }

        namespace view
        {
          namespace function_objects {}
          using namespace function_objects;
        }

        namespace action
        {
          namespace function_objects {}
          using namespace function_objects;
        }

        namespace detail
        {
          namespace function_objects {}
          using namespace function_objects;
        }
```

but, does anybody have a better idea?

ericniebler:
> I think I'd rather use an anonymous namespace in the macro

Does that work? If not, then pls squash this and I'll commit. If so, I think that would be an improvement.

CaseyCarter:
>  I think I'd rather use an anonymous namespace in the macro

If you put the definitions in unnamed namespaces, they define distinct objects in each TU, in which case there's no reason to declare them `inline`.

gnzlbg:
@ericniebler squashed, let's wait till it passes the tests under clang-trunk.

@CaseyCarter Yes, forgot about that. That was a pointless suggestion. We really do need some named namespace.

CaseyCarter:
I admit I'm not a fan of the name `function_objects` despite it being my suggestion. The name isn't significant in itself, it's just something that will appear in error messages where `::<anonymous>::` used to be. We can trivially change it whenever we decide on a better color for the bikeshed.

include/range/v3/detail/config.hpp
```diff
@@ -291,6 +291,42 @@
...
+// TODO: remove once clang defines __cpp_inline_variables (or equivalent)
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_INLINE_VARIABLES_17
+#else
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_FEATURE(INLINE_VARIABLES)
+#endif  // __cpp_inline_variables
+
+#endif  // RANGES_CXX_INLINE_VARIABLES
+
+#if RANGES_CXX_INLINE_VARIABLES < RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE_(type, name, inline_)      \
```
viboes: inline_ not used here

include/range/v3/detail/config.hpp
```diff
@@ -291,6 +291,42 @@
...
+    inline_ namespace                                     \
+    {                                                     \
+        constexpr auto& name = static_const<type>::value; \
+    }                                                     
+
+#else  // RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE_(type, name, inline_) \
```
viboes: neither here

include/range/v3/detail/config.hpp
```diff
@@ -291,6 +291,42 @@
...
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_INLINE_VARIABLES_17
+#else
+#define RANGES_CXX_INLINE_VARIABLES RANGES_CXX_FEATURE(INLINE_VARIABLES)
+#endif  // __cpp_inline_variables
+
+#endif  // RANGES_CXX_INLINE_VARIABLES
+
+#if RANGES_CXX_INLINE_VARIABLES < RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE_(type, name, inline_)      \
+    inline_ namespace                                     \
```
CaseyCarter: ^^ `inline_` is used here.

include/range/v3/detail/config.hpp
```diff
@@ -291,6 +291,42 @@
...
+    {                                                     \
+        constexpr auto& name = static_const<type>::value; \
+    }                                                     
+
+#else  // RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
+#define RANGES_INLINE_VARIABLE_(type, name, inline_) \
+    namespace function_objects                       \
```
CaseyCarter: And intentionally not used here. The point of `inline_` is to insert `RANGES_GCC_BROKEN_CUSTPOINT` into the definition of `RANGES_INLINE_VARIABLE_` only when inline variables are not supported. (I'm not sure why @gnzlbg passes it in through a parameter instead of directly putting `RANGES_GCC_BROKEN_CUSTPOINT` in the definition of `RANGES_INLINE_VARIABLE`, but it does not seem worth the trouble to change.)

gnzlbg: Nice catch, I think this was a leftover from the original implementation that tried to support variable templates. 

@CaseyCarter I think that instead of:

```c++
namespace function_objects {}
using namespace function_objects;
```

we can just make the function objects namespace `inline`, so that we get rind of `RANGES_GCC_BROKEN_CUSTPOINT`:

```c++
inline namespace function_objects {}
```

Do you see any drawback to this approach?

CaseyCarter: My vague recollection is that some supported versions of GCC only worked *with* `inline`, and some supported versions only worked *without* `inline`, so that we could not simply tack `inline` onto all of the unnamed namespaces. All supported variations have Travis coverage, so feel free to make the change and see if it makes it through Travis.

gnzlbg: I've just removed it I think, we'll see if anything breaks.

wilhelmtell:
I don't know if I'm beating a dead horse here, but..

> I think I'd rather use an anonymous namespace in the macro

>> If you put the definitions in unnamed namespaces, they define distinct objects in each TU, in which case there's no reason to declare them inline.
>>
>> EDIT: And formally, we get ODR violations.

@CaseyCarter How can stuff in an unnamed namespace violate ODR?

But never mind that.. @gnzlbg What good is it to put stuff that's in a header in an unnamed namespace? Wouldn't that bloat the TUs' object code, with no benefit? Isn't this what a nested `detail` namespace is for?

CaseyCarter:
> @CaseyCarter How can stuff in an unnamed namespace violate ODR?

When external linkage entities with definitions in multiple TUs refer to different entities with internal linkage using the same name:
```c++
constexpr int foo = 42;
inline int& f() { return foo; }
```

PR #431, opened 2016-07-11
Minor

CaseyCarter:
Several small, independent changes not worth individual pull requests.


include/range/v3/detail/optional.hpp
```diff
@@ -37,7 +37,7 @@ namespace ranges
             optional(T t)
               : data_(emplaced_index<1>, std::move(t))
             {}
-            template <typename...Args, CONCEPT_REQUIRES_(Constructible<T, Args &&...>())>
+            template <typename...Args, CONCEPT_REQUIRES_(Constructible<T, Args...>())>
```
ericniebler: Why?

CaseyCarter: Equivalent but shorter.

ericniebler: Are you not forgetting that `Concept<T>()` in range-v3 means `Concept<T&>()`?

CaseyCarter: Not this time ;) `Constructible` have been cured of that madness and use `std::declval`. I REALLY need to find some time to transition the rest of the library.

ericniebler: >  I REALLY need to find some time to transition the rest of the library

I would love such a change, but until then, I prefer the consistency. Please stick to T&& for now, as annoying as it is.

CaseyCarter: Fair enough.

CaseyCarter:
Squashed without the `optional` change.

PR #435, opened 2016-07-21
GCC 6.2 will fix PR 71117

CaseyCarter:
...and won't need the `could_be_lambda` workaround.

### DIFF
--- a/include/range/v3/utility/box.hpp
+++ b/include/range/v3/utility/box.hpp
@@ -157,9 +157,9 @@ namespace ranges
             }
             template<typename T, typename = meta::if_<
                 meta::strict_and<std::is_empty<T>, meta::not_<detail::is_final<T>>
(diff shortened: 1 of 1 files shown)


ericniebler:
Should be `__GNUC__ == 6 && __GNUC_MINOR__ < 2`, right?

CaseyCarter:
You mean you aren't concerned that the 7.1 release will bring the bug back?

Yes, I'm a moron.

gnzlbg:
On Thu, Jul 21, 2016 at 7:47 PM, Casey Carter <notifications@github.com>
wrote:

> You mean you aren't concerned that the 7.1 release will bring the bug back?


After surviving the GCC5pocalypse I wouldn't discard anything :D

CaseyCarter:
`constexpr` in GCC 5 is like `concept` in GCC 6. Sure, it's there, but using it can

PR #436, opened 2016-08-03
fix typo in error string

wilhelmtell:


wilhelmtell:
Thank you for the detailed error text! My god we live in a new era! o,O

:D

ericniebler:
Thanks!

PR #437, opened 2016-08-04
rm virtual keyword in override

wilhelmtell:
what good is it stating an override is virtual? the base class,
std::bad_cast, already says the what() member function is virtual.

if the base class' signature is virtual, then the keyword is a no-op. if
the base class' signature were not virtual, then the virtual override
can bring confusion down the line.

EDIT: I hear now that some compilers complain if the derived class doesn't repeat the virtual keyword. Is that so? I think such a warning is unhelpful. I think that repeating the virtual keyword, when it's doing nothing, is at best of no benefit. But it's also fragile..

### DIFF
--- a/include/range/v3/utility/any.hpp
+++ b/include/range/v3/utility/any.hpp
@@ -28,7 +28,7 @@ namespace ranges
         struct bad_any_cast
           : std::bad_cast
         {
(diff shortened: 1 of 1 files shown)


gnzlbg:
Spurious travis failure, tests seem to pass.

PR #438, opened 2016-08-04
copy_n + istream_iterator: fix an excess read

wilhelmtell:
calling copy_n() with a single-pass input iterator, like an
istream_iterator, triggers an excess read from the iterator:

    std::istringstream iss{"1 2 3 4 5"};
    int output[] = { 0, 0, 0, 0 };
    auto const p = copy_n(std::istream_iterator<int>{iss}, 2, &output[0]);
    copy_n(std::istream_iterator<int>{iss}, 2, p.second);

this fills the output array with

    { 1, 2, 4, 5 }

rather than

    { 1, 2, 3, 4 }

### DIFF
--- a/include/range/v3/algorithm/copy_n.hpp
+++ b/include/range/v3/algorithm/copy_n.hpp
@@ -48,8 +48,16 @@ namespace ranges
                 RANGES_ASSERT(0 <= n);
                 auto norig = n;
                 auto b = uncounted(begin);
-                for(; n != 0; ++b, ++out, --n)
+                if(n > 0) {
+                  while(true) {
                     *out = *b;
+                    ++out;
+                    if(--n > 0)
+                      ++b;
+                    else
+                      break;
+                  }
+                }
                 return {recounted(begin, b, norig), out};
             }
         };
--- a/test/algorithm/CMakeLists.txt
+++ b/test/algorithm/CMakeLists.txt
@@ -19,6 +19,9 @@ add_test(test.alg.copy, alg.copy)
 add_executable(alg.copy_backward copy_backward.cpp)
 add_test(test.alg.copy_backward, alg.copy_backward)
 
+add_executable(alg.copy_n copy_n.cpp)
+add_test(test.alg.copy_n, alg.copy_n)
+
 add_executable(alg.count count.cpp)
 add_test(test.alg.count, alg.count)
 
--- a/test/algorithm/copy_n.cpp
+++ b/test/algorithm/copy_n.cpp
@@ -1,0 +1,37 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+
+#include <iterator>
+#include <sstream>
+#include <range/v3/core.hpp>
+#include <range/v3/algorithm/copy_n.hpp>
+#include "../simple_test.hpp"
+
+int main()
+{
+    using ranges::begin;
+    using ranges::end;
+    using ranges::size;
+
+    {
+      std::istringstream iss{ "1 2 3 4 5" };
+      int output[] = { 0, 0, 0, 0 };
+      auto const p = ranges::copy_n(std::istream_iterator<int>{iss}, 2, &output[0]);
+      ranges::copy_n(std::istream_iterator<int>{iss}, 2, p.second);
+      int const expected[] = { 1, 2, 3, 4 };
+      CHECK(output[0] == expected[0]);
+      CHECK(output[1] == expected[1]);
+      CHECK(output[2] == expected[2]);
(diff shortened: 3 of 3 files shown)


wilhelmtell:
LLVM and GCC had a similar issue a number of years ago.

CaseyCarter:
`copy_n` increments its argument iterator exactly `n` times, just as is the case for every `_n` algorithm. The error in the example code is the mistaken belief that you can create a new `istream_iterator` that is equivalent to an existing `istream_iterator` for the same stream. All of the major STL implementations now have a broken `copy_n` because someone convinced their maintainers that it was more important for this particular usage with a single iterator type to meet their expectations than for `copy_n` to have semantics consistent with *every other algorithm in the standard library*.

Range-v3's `copy_n` avoids this problem by properly returning a pair consisting of the final values of the input iterator and output iterator; you want:
```c++
std::istringstream iss{"1 2 3 4 5"};
int output[] = { 0, 0, 0, 0 };
auto const p = copy_n(std::istream_iterator<int>{iss}, 2, &output[0]);
copy_n(p.in(), 2, p.out());
```

gnzlbg:
From N4382 25.3.1:

> ```c++
template<WeakInputIterator I, WeaklyIncrementable O>
requires IndirectlyCopyable<I, O>
pair<I, O>
copy_n(I first, iterator_distance_t<I> n, O result);
```
> 5 Effects: For each non-negative integer `i < n`, performs `*(result + i) = *(first + i)`.
> 6 Returns: `result + n{first + n, result + n}`.
> 7 Complexity: Exactly `n` assignments.

it looks to me that both `first` and `result` must be incremented `n` times.

**Comment by CaseyCarter** [comment]
For completeness, I should point out this is [LWG2471](http://wg21.link/lwg2471) "`copy_n`'s number of `InputIterator` increments unspecified". In discussion of this issue in Chicago this week there was no strong consensus for any resolution of the issue: the group was split between those that wanted a `copy_n` that is not broken and implementors who already broke their `copy_n` and refuse to change it because backwards compatibility. LWG decided to pass the buck on 2471 to LEWG.

**Comment by gnzlbg** [comment]
Is there an easy way to not loose an element of the stream when using `std::copy_n` with input iterators?

The only thing I can think of is something like:

```c++
std::istringstream iss{"1 2 3 4 5"};
int output[] = { 0, 0, 0, 0 };
auto it = std::istream_iterator<int>{iss};
auto const p = copy_n(iterator_reference(it), 2, &output[0]);
copy_n(iterator_reference(it), 2, p.second);
```

But that would require something like an `iterator_reference` to at least make it work.

Incrementing input iterators `n-1` times seems the lesser evil.

wilhelmtell:
I think clang and gcc also do tag dispatching to let random access iterators bypass this workaround.

wilhelmtell:
I know this changes a perfectly simple, clear, and consistent algorithm, into something that is neither nor. And I see @CaseyCarter 's code works. But I didn't come up with it right away, nor I think would any of the other average users out there. I think that code just shifts the inconsistency from the library code to the user code.

ericniebler:
> The error in the example code is the mistaken belief that you can create a new `istream_iterator` that is equivalent to an existing `istream_iterator` for the same stream.

This. Thanks, Casey.

CaseyCarter:
> I think clang and gcc also do tag dispatching to let random access iterators bypass this workaround.

@mclow Posted some investigation to the Library reflector:

I just tried this with libc++:
calling copy_n(iter, 4, back_inserter<blah>(blah));

This is what I got:
  Input Iterator count = 3
  Forward Iterator count = 3
  Bidi Iterator count = 3
  RA Iterator count = 4

FWIW, I got the same answer with libstc++ and gcc 6.1.

For VS (using webcompiler.cloudapp.net), I get:
   Input Iterator count = 3
   Forward Iterator count = 4
   Bidi Iterator count = 4
   RA Iterator count = 4

There is implementation divergence around what is the *best* way to break `copy_n` ;)

CaseyCarter:
> Incrementing input iterators n-1 times seems to me to be the lesser evil, even if it is inconsistent.

I refuse to accept a design in which `copy_n(i, n, o)` has different semantics from `copy(counted_iterator{i, n}, default_sentinel{}, o)`.

wilhelmtell:
But, I wish there was a way to let the user know at compile time that there's an issue with

    p = copy_n(std::istream_iterator<int>{std::cin}, 2, &output[0]);
    copy_n(std::istream_iterator<int>{std::cin}, 2, p);

wilhelmtell:
Maybe `std::istream_iterator<>` should not exist.

wilhelmtell:
@CaseyCarter 

> implementors who already broke their copy_n and refuse to change it because backwards compatibility

If they broke `copy_n()` there is nothing to fear in fixing it. standard compliance triumphs over compatibility with standard in-compliance.

gnzlbg:
> I refuse to accept a design in which copy_n(i, n, o) and copy(counted_iterator{i, n}, default_sentinel{}, o) have different effects.

@CaseyCarter Just to clarify, I was just talking about `std::` and not `std2::`. What do you suggest that we do in `std::`? There are many options, but I don't know of any without drawbacks. I can understand why doing nothing seems attractive.

PR #440, opened 2016-08-10
zip_with's underlying iterators do not always model Sentinel

### LINKED ISSUE #439: view::zip with view::for_each fails to compile
Here's a minimal example:

    #include <vector>
    
    #include <range/v3/core.hpp>
    #include <range/v3/to_container.hpp>
    #include <range/v3/view/for_each.hpp>
    #include <range/v3/view/zip.hpp>
    
    
    using namespace ranges;
    
    int main()
    {
        const auto v1 = std::vector<int>{1, 2, 3};
        const auto v2 = std::vector<char>{'a', 'b', 'c'};
        view::zip(v1 | view::for_each([](const auto& el) { return yield(el); }), v2) | to_vector;
        // expected std::vector<std::pair<int,char>>{{1,'a'},{2,'b'},{3,'c'}}
    }

And here's the compilation error: http://susepaste.org/3a5b89e2

I'm using gcc 6.1.1

CaseyCarter:
Fixes #439.

### DIFF
--- a/include/range/v3/view/zip_with.hpp
+++ b/include/range/v3/view/zip_with.hpp
@@ -167,6 +167,7 @@ namespace ranges
                 {
                     tuple_for_each(its_, detail::inc);
                 }
+                CONCEPT_REQUIRES(meta::and_c<(bool) Sentinel<range_iterator_t<Rngs>, range_iterator_t<Rngs>>()...>::value)
                 bool equal(cursor const &that) const
                 {
                     // By returning true if *any* of the iterators are equal, we allow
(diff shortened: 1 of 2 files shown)


ericniebler:
Would be awesome if this change came with a test. :-)  Not that I'm not appreciative.

ericniebler:
The test seems to have turned up a problem.

ericniebler:
I think if you rebase this on the current master, the problem goes away.

CaseyCarter:
Great job with the `iterator_traits` fix. test.view.zip compiles locally with GCC-4.8 and 4.9, I suspect you are correct.

Issue #441, opened 2016-08-11
view::zip with view::for_each fails to compile (another case)


etam:
Here's the minimal example:

```
#include <vector>

#include <range/v3/core.hpp>
#include <range/v3/view/for_each.hpp>
#include <range/v3/view/zip.hpp>

int main()
{
    using namespace ranges;

    const auto v1 = std::vector<int>{1, 2, 3};
    const auto v2 = std::vector<char>{'a', 'b', 'c'};

    view::zip(v1, v2) | view::for_each([](const auto& value) { return yield(value); });
}
```

output from gcc 6.1.1: http://susepaste.org/533f6b44
output from clang 3.8.0: http://susepaste.org/c7b2e98f

ericniebler:
Looking at CaseyCarter/range-v3@3b6544cb7b6d64446e4706d7ac433081a23177ed, is the problem that the `pair` created by the `zip` view not copy constructible?

CaseyCarter:
Yes, the basic problem is that `view::single` requires a `SemiRegular` type, and the reference type of a zip view is (generally) not `SemiRegular` because it has reference members. This is sort-of fixed by pushing the error up to `yield`.

On a higher level, the root cause is that proxy references don't decay to their value types as do core language references when passed to a function template that accepts parameters by value. Fixing this - indeed deciding if it needs to *be* fixed - is research.

aldanor:
What would the minimal suggested change be for the example above then?

FloopCZ:
Hi, the minimal possible change might be to explicitly specify the lambda argument type to something copy constructible:
```
view::zip(v1, v2) | view::for_each([](std::pair<int, char> value) { return yield(value); });
```
@ericniebler Maybe zip could return std::reference_wrapper instead of int& or char&?

FloopCZ:
So with PR #559, one can yield a pair of references:
```
view::zip(v1, v2) 
| view::for_each([](std::pair<const int&, const char&> value) { return yield(value); });
```
However, it is not possible to use automatic type deduction for the lambda argument, as requested by the issuer:
```
view::zip(v1, v2)
| view::for_each([](const auto& value) { return yield(value); });
```
Any ideas why the lambda argument type cannot be deduced automatically?

FloopCZ:
ok, I dug into the issue and the problem seems to be with the IndirectInvocable concept. If I comment out the concept check in view/for_each.hpp:63
```
IndirectInvocable<F, range_iterator_t<Rng>>
[...]

PR #445, opened 2016-08-17
Improve error message for misuse of yield().

CaseyCarter:
Partially addresses #441.


ericniebler:
I'm inclined to wait for the proper fix for this problem: relax the requirements of `view::single` to make the code work.

ericniebler:
OK, I'll merge this for now and implement the proper fix when I get a chance. Thanks.

PR #463, opened 2016-09-01
Make gh-pages.clean actually clean

timsong-cpp:
Right now at least on my system it's seemingly trying to remove a file literally named `*.meow` rather than all files with the extension `meow`.


ericniebler:
Nice, thanks.

PR #476, opened 2016-09-29
Fix Clang and libc++ Travis build

EricWF:
This patch fixes the Clang and libcxx travis builders. The main changes are:

1. Upgrade to Clang 4.0 since that's what the llvm-nightly repository contains.
2. Rewrite the install_libcxx.sh script so it works better, and supports building with sanitizers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -174,7 +174,7 @@ matrix:
     #   os: linux
     #   addons: *clang37
 
-    # Test clang-3.9: C++11/C++14, Build=Debug/Release, ASAN=On/Off
+    # Test clang-4.0: C++11/C++14/1z, LIBCXX=On/Off, Build=Debug/Release, ASAN=On/Off
     - env: CLANG_VERSION=4.0 BUILD_TYPE=Debug CPP=11 ASAN=On LIBCXX=On
       os: linux
       addons: &clang40
--- a/.travis.yml
+++ b/.travis.yml
@@ -175,61 +175,61 @@ matrix:
     #   addons: *clang37
 
     # Test clang-3.9: C++11/C++14, Build=Debug/Release, ASAN=On/Off
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Debug CPP=11 ASAN=On LIBCXX=On
+    - env: CLANG_VERSION=4.0 BUILD_TYPE=Debug CPP=11 ASAN=On LIBCXX=On
       os: linux
-      addons: &clang39
+      addons: &clang40
         apt:
           packages:
             - util-linux
-            - clang-3.9
+            - clang-4.0
             - valgrind
           sources:
             - ubuntu-toolchain-r-test
             - llvm-toolchain-precise
 
     # Release + ASAN + C++11/14/1z
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Release CPP=11 ASAN=On LIBCXX=On
+    - env: CLANG_VERSION=4.0 BUILD_TYPE=Release CPP=11 ASAN=On LIBCXX=On
       os: linux
-      addons: *clang39
+      addons: *clang40
 
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Release CPP=14 ASAN=On LIBCXX=On
+    - env: CLANG_VERSION=4.0 BUILD_TYPE=Release CPP=14 ASAN=On LIBCXX=On
       os: linux
-      addons: *clang39
+      addons: *clang40
 
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Release CPP=1z ASAN=On LIBCXX=On
+    - env: CLANG_VERSION=4.0 BUILD_TYPE=Release CPP=1z ASAN=On LIBCXX=On
       os: linux
-      addons: *clang39
+      addons: *clang40
 
     # Release + Valgrind + C++11/14/1z
-    - env: CLANG_VERSION=3.9 BUILD_TYPE=Release CPP=11 ASAN=Off LIBCXX=On
+    - env: CLANG_VERSION=4.0 BUILD_TYPE=Release CPP=11 ASAN=Off LIBCXX=On
       os: linux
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -175,61 +175,61 @@ matrix:
     #   addons: *clang37
 
     # Test clang-3.9: C++11/C++14, Build=Debug/Release, ASAN=On/Off
```
CaseyCarter: Lies! This does not test clang-3.9.

EricWF: Ack. Fixed.

EricWF:
> +1 from me. I'd rather have working 4.0 than broken 3.9 support.

I'll look into re-adding Clang 3.9 back, but I just wanted to get everything, including libc++, working first.

EricWF:
Looks like there is one failure caused by a timeout, but that's unrelated to this fix.

CaseyCarter:
> Looks like there is one failure caused by a timeout, but that's unrelated to this fix.

I had to disable g++ ASAN builds because of timeouts, it looks like we may have to do the same for clang builds and have no ASAN on Travis.

...or split the test suite into multiple pieces and have 10,000 jobs.

EricWF:
>  I had to disable g++ ASAN builds because of timeouts, it looks like we may have to do the same for clang builds and have no ASAN on Travis.
> 
> ...or split the test suite into multiple pieces and have 10,000 jobs.

The bot is compiling with -Ofast, maybe turning that down to -O1 will help save some time.

CaseyCarter:
> The bot is compiling with -Ofast, maybe turning that down to -O1 will help save some time.

It sounds reasonable that the delta between -O1 and -O2 compile time may be larger than the delta between -O1 and -O2 tests. Worth investigating, anyway.

EricWF:
@CaseyCarter: I don't have commit access so please merge this when appropriate

CaseyCarter:
@EricNiebler, give me commit access so I can commit this. (He's getting married tomorrow, don't hold your breath.)

install_libcxx.sh
```diff
@@ -1,15 +1,22 @@
...
-(cd libcxx/build && ../../cmake/bin/cmake .. -DLIBCXX_CXX_ABI=libstdc++ -DLIBCXX_CXX_ABI_INCLUDE_PATHS="/usr/include/c++/4.6;/usr/include/c++/4.6/x86_64-linux-gnu")
-make -C libcxx/build cxx -j2
-sudo cp libcxx/build/lib/libc++.so.1.0 /usr/lib/
-sudo cp -r libcxx/build/include/c++/v1 /usr/include/c++/v1/
-sudo ln -sf /usr/lib/libc++.so.1.0 /usr/lib/libc++.so
-sudo ln -sf /usr/lib/libc++.so.1.0 /usr/lib/libc++.so.1
+#!/usr/bin/env bash
+
+# Install a newer CMake version
+curl -sSL https://cmake.org/files/v3.6/cmake-3.6.1-Linux-x86_64.sh -o install-cmake.sh
```
gnzlbg: nitpick: I kind of prefer having the CMake URL in some place that is easy to change.

gnzlbg:
Looks good to me. 

I've always wondered why the new "sudo-less" travis builds do allow `sudo` in the build scripts and whether we are doing the right thing there but... it always worked just fine. 

@ericniebler congratulations!

install_libcxx.sh
EricWF: Just let me know where to put it.

Although I should note that a newer CMake is only required when compiling LLVM/libc++ from source, so that's why it's local to this file.

gnzlbg: I would just put it at the top of the file like `CMAKE_URL=...  # Update this when LLVM requirements change`.

ericniebler:
Thanks for all the work guys. Should this still be failing on Travis, though?

CaseyCarter:
>  /usr/bin/ld: cannot find -lc++

Gobsmacked as to how this was working before, but now is not.

CaseyCarter:
> ./install_libcxx.sh: line 5: CMAKE_URL: command not found
> curl: no URL specified!
> curl: try 'curl --help' or 'curl --manual' for more information

Ok, less gobsmacked now.

install_libcxx.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
 # Install a newer CMake version
-curl -sSL https://cmake.org/files/v3.6/cmake-3.6.1-Linux-x86_64.sh -o install-cmake.sh
+# Update this when LLVM requirements change
+CMAKE_URL = https://cmake.org/files/v3.6/cmake-3.6.1-Linux-x86_64.sh
```
CaseyCarter: CMAKE_URL=https://cmake.org/files/v3.6/cmake-3.6.1-Linux-x86_64.sh

install_libcxx.sh
```diff
@@ -1,15 +1,24 @@
...
-git clone --depth=1 https://github.com/llvm-mirror/libcxx.git
-mkdir libcxx/build
-(cd libcxx/build && ../../cmake/bin/cmake .. -DLIBCXX_CXX_ABI=libstdc++ -DLIBCXX_CXX_ABI_INCLUDE_PATHS="/usr/include/c++/4.6;/usr/include/c++/4.6/x86_64-linux-gnu")
-make -C libcxx/build cxx -j2
-sudo cp libcxx/build/lib/libc++.so.1.0 /usr/lib/
-sudo cp -r libcxx/build/include/c++/v1 /usr/include/c++/v1/
-sudo ln -sf /usr/lib/libc++.so.1.0 /usr/lib/libc++.so
-sudo ln -sf /usr/lib/libc++.so.1.0 /usr/lib/libc++.so.1
+#!/usr/bin/env bash
+
```
CaseyCarter: Let's insert a `set -e` line near the top here so the script will fail if any of the commands fails.

EricWF:
OK. Now it's no longer failing. Sorry about that.

PR #478, opened 2016-10-03
Rename not_ to not_fn, implement as in the C++17 WP:

CaseyCarter:
* call operators forward value category
* call operators SFINAE on return type of the wrapped function
* extension: call operators propagate `noexcept`
* Keep the name `not_` around with a deprecation notice


CaseyCarter:
Curse you, GCC 4.8.

CaseyCarter:
Travis failure is a timeout on Clang4 C++1z ASAN

PR #480, opened 2016-10-09
Use a reusable CMake module to perform the header checks

ldionne:
Also, the header checks are now implemented as normal targets, which means they are performed whenever a header changes and at build-time, instead of at CMake configure-time.

If the current behavior is preferred, feel free to discard this PR. I simply wrote the `HeaderTests` module for Hana and thought it might be useful elsewhere.


ericniebler:
Fabulous! Thanks, Louis!

gnzlbg:
Thanks! The original implementation was also using build targets but for some reason I cannot recall was switched to use `check_include_file_cxx` which... I have been hating since (it took a _very long_ time to run it in my system).

PR #482, opened 2016-10-20
Add an optional stepsize paramter to view::chunk

tobimpub:
In other words, adds support for sliding-window like behaviour.
ex: view::ints | view::take(3) | view::chunk(2,1)
    returns [[0,1],[1,2],[2]]

### DIFF
--- a/include/range/v3/view/chunk.hpp
+++ b/include/range/v3/view/chunk.hpp
@@ -51,24 +51,26 @@ namespace ranges
                     range_difference_t<Rng>,
                     constant<range_difference_t<Rng>, 0>>;
             range_difference_t<Rng> n_;
+            range_difference_t<Rng> s_;
             friend range_access;
             struct adaptor;
             adaptor begin_adaptor() const
             {
-                return adaptor{n_, ranges::end(this->base())};
+                return adaptor{n_, s_, ranges::end(this->base())};
             }
         public:
             chunk_view() = default;
-            chunk_view(Rng rng, range_difference_t<Rng> n)
-              : chunk_view::view_adaptor(std::move(rng)), n_(n)
+            chunk_view(Rng rng, range_difference_t<Rng> n, range_difference_t<Rng> s)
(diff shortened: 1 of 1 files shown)


ericniebler:
Travis says that the chunk test is now failing:

```
151: Test command: /home/travis/build/ericniebler/range-v3/build/test/view/view.chunk
151: Test timeout computed to be: 1500
151: > ERROR: CHECK failed "sizeof(rng1.begin()) == sizeof(v.begin()) * 2 + sizeof(std::ptrdiff_t) * 2"
151: > 	/home/travis/build/ericniebler/range-v3/test/view/chunk.cpp(41)
151: > 	EXPECTED: 32
151: > 	ACTUAL: 40
151: > ERROR: CHECK failed "sizeof(rng2.begin()) == sizeof(l.begin()) * 2 + sizeof(std::ptrdiff_t)"
151: > 	/home/travis/build/ericniebler/range-v3/test/view/chunk.cpp(55)
151: > 	EXPECTED: 24
151: > 	ACTUAL: 32
151/190 Test #151: test.view.chunk, ......................***Failed    0.02 sec
```

ericniebler:
Also, please add new tests for the new functionality.

ericniebler:
And I'm still on the fence about whether `view::chunk` is the right place for sliding window functionality. It feels a bit off to me.

CaseyCarter:
> And I'm still on the fence about whether `view::chunk` is the right place for sliding window functionality. It feels a bit off to me.

Seconded. The failing test is looking at the size of the iterators; the change increases iterator size but does not alter the test to agree. I think that increase in iterator size is enough to trigger my "don't pay for what you don't use" detector. Why should this be an optional feature of `view::chunk` instead of a distinct view?

tobimpub:
I agree a new view would probably be a better way to do this, I just wanted to gather your input on the idea first.
So how about view::sliding?

ericniebler:
> So how about `view::sliding`?

Sounds reasonable.

PR #488, opened 2016-10-24
Add a sliding view

tobimpub:
Example: view::ints(0,3) | view::sliding(2) gives [[0,1],[1,2],[2]]

### DIFF
--- a/include/range/v3/view/sliding.hpp
+++ b/include/range/v3/view/sliding.hpp
@@ -1,0 +1,183 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2014
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_SLIDING_HPP
+#define RANGES_V3_VIEW_SLIDING_HPP
+
+#include <utility>
+#include <functional>
+#include <meta/meta.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/iterator_range.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/view_adaptor.hpp>
+#include <range/v3/view/all.hpp>
+#include <range/v3/view/view.hpp>
+#include <range/v3/view/take.hpp>
+#include <range/v3/utility/box.hpp>
+#include <range/v3/utility/static_const.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        /// \addtogroup group-views
+        /// @{
+        template<typename Rng>
+        struct sliding_view
+          : view_adaptor<
+                sliding_view<Rng>,
+                Rng,
+                is_finite<Rng>::value ? finite : range_cardinality<Rng>::value>
+        {
+        private:
+            CONCEPT_ASSERT(ForwardRange<Rng>());
+            using offset_t =
+                meta::if_<
+                    BidirectionalRange<Rng>,
+                    range_difference_t<Rng>,
+                    constant<range_difference_t<Rng>, 0>>;
+            range_difference_t<Rng> n_;
+            friend range_access;
+            struct adaptor;
+            adaptor begin_adaptor() const
+            {
+                return adaptor{n_, ranges::end(this->base())};
+            }
+        public:
+            sliding_view() = default;
+            sliding_view(Rng rng, range_difference_t<Rng> n)
+              : sliding_view::view_adaptor(std::move(rng)), n_(n)
+            {
+                RANGES_ASSERT(0 < n_);
+            }
+            CONCEPT_REQUIRES(SizedRange<Rng>())
+            range_size_t<Rng> size() const
+            {
+                auto sz = ranges::distance(this->base());
+                return static_cast<range_size_t<Rng>>(sz);
+            }
+        };
+
+        template<typename Rng>
+        struct sliding_view<Rng>::adaptor
+          : adaptor_base, private box<offset_t>
+        {
+        private:
+            range_difference_t<Rng> n_;
+            range_sentinel_t<Rng> end_;
+            offset_t & offset() {return this->box<offset_t>::get();}
+            offset_t const & offset() const {return this->box<offset_t>::get();}
+        public:
+            adaptor() = default;
+            adaptor(range_difference_t<Rng> n, range_sentinel_t<Rng> end)
+              : box<offset_t>{0}, n_(n), end_(end)
+            {}
+            auto get(range_iterator_t<Rng> it) const ->
+                decltype(view::take(make_iterator_range(std::move(it), end_), n_))
+            {
+                RANGES_ASSERT(it != end_);
+                RANGES_ASSERT(0 == offset());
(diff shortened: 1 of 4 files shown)


tobimpub:
maybe view::slide would be more consistent with the other view names...

tobimpub:
updated

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,182 @@
...
+            range_size_t<Rng> size() const
+            {
+                auto sz = ranges::distance(this->base());
+                return static_cast<range_size_t<Rng>>(sz);
+            }
+        };
+
+        template<typename Rng>
+        struct sliding_view<Rng>::adaptor
+          : adaptor_base, private box<offset_t>
```
CaseyCarter: `view::chunk(n)`'s iterators normally wrap base iterators that are `n` steps apart. They need to store `offset` so that the last-but-one iterator can be less than `n` steps away from the `end` iterator when the size of the underlying range is not a multiple of `n`. This view's iterators correspond exactly to the iterators of the underlying view, so `offset` is not necessary.

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,182 @@
...
+        {
+        private:
+            range_difference_t<Rng> n_;
+            range_sentinel_t<Rng> end_;
+            offset_t & offset() {return this->box<offset_t>::get();}
+            offset_t const & offset() const {return this->box<offset_t>::get();}
```
CaseyCarter: So these can go away.

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,182 @@
...
+            {
+                RANGES_ASSERT(it != end_);
+                RANGES_ASSERT(0 == offset());
+                offset() = ranges::advance(it, 1, end_);
+            }
+            CONCEPT_REQUIRES(BidirectionalRange<Rng>())
+            void prev(range_iterator_t<Rng> &it)
+            {
+                ranges::advance(it, -1 + offset());
+                offset() = 0;
```
CaseyCarter: `--it;`

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,182 @@
...
+            }
+            CONCEPT_REQUIRES(
+                SizedSentinel<range_iterator_t<Rng>, range_iterator_t<Rng>>())
+            range_difference_t<Rng> distance_to(range_iterator_t<Rng> const &here,
+                range_iterator_t<Rng> const &there, adaptor const &that) const
+            {
+                // This assertion is true for all range types except cyclic ranges:
+                //RANGES_ASSERT(0 == ((there - here) + that.offset());
+                return ((there - here) + that.offset() - offset());
```
CaseyCarter: `return there - here;`

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,182 @@
...
+            }
+            CONCEPT_REQUIRES(RandomAccessRange<Rng>())
+            void advance(range_iterator_t<Rng> &it, range_difference_t<Rng> n)
+            {
+                if(0 < n)
+                    offset() = ranges::advance(it, n + offset(), end_);
+                else if(0 > n)
+                    offset() = (ranges::advance(it, n + offset()), 0);
+            }
```
CaseyCarter:
```c++
RANGES_ASSERT(end_ - it <= n);
it += n;
```

test/view/sliding.cpp
```diff
@@ -0,0 +1,152 @@
...
+        ::check_equal(*it++, {2,3,4});
+        ::check_equal(*it++, {3,4,5});
+        ::check_equal(*it++, {4,5,6});
+        ::check_equal(*it++, {5,6});
+        ::check_equal(*it++, {6});
+        CHECK(it == ranges::end(rng));
+        ::check_equal(*ranges::next(it, -3), {4,5,6});
+        CHECK(size(rng), 7u);
+        if (!ranges::v3::detail::broken_ebo)
+            CHECK(sizeof(rng.begin()) == sizeof(v.begin()) * 2 + sizeof(std::ptrdiff_t) * 2);
```
CaseyCarter: ..which should shrink the iterators by a `ptrdiff_t`.

include/range/v3/view/sliding.hpp
tobimpub: well it is kinda obvious in retrospect...
thanks for the explanation!

include/range/v3/view/sliding.hpp
```diff
@@ -73,53 +68,45 @@ namespace ranges
...
+                return there - here;
             }
             CONCEPT_REQUIRES(RandomAccessRange<Rng>())
             void advance(range_iterator_t<Rng> &it, range_difference_t<Rng> n)
             {
-                if(0 < n)
-                    offset() = ranges::advance(it, n + offset(), end_);
-                else if(0 > n)
-                    offset() = (ranges::advance(it, n + offset()), 0);
+                //RANGES_ASSERT(end_ - it <= n);
```
CaseyCarter: Wooops. We require `RandomAccessRange`, but not `SizedRange`. I guess l lied about this assertion being a good idea.

CaseyCarter:
> maybe `view::slide` would be more consistent with the other view names...

`sliding_window`? It's a bit long, but very discoverable.

ericniebler:
> > maybe `view::slide` would be more consistent with the other view names...
>
> `sliding_window`? It's a bit long, but very discoverable.

`view::sliding`? Not likely to be confusing to those looking for it.

test/view/sliding.cpp
```diff
@@ -0,0 +1,152 @@
...
+        ::models<concepts::RandomAccessRange>(rng);
+        ::models<concepts::SizedRange>(rng);
+        auto it = ranges::begin(rng);
+        ::check_equal(*it++, {0,1,2});
+        ::check_equal(*it++, {1,2,3});
+        ::check_equal(*it++, {2,3,4});
+        ::check_equal(*it++, {3,4,5});
+        ::check_equal(*it++, {4,5,6});
+        ::check_equal(*it++, {5,6});
+        ::check_equal(*it++, {6});
```
ericniebler: For `view::sliding(N)`, there will be N-1 elements at the end that have fewer than N elements. But not at the start. Why are these the correct semantics?

tobimpub: The behaviour is simply a result of my use case, there is nothing inherently correct about it.
Making it symmetrical is probably a good idea. The user can still drop the first N-1 to get the current semantics. (open variant)
I did some research and found that most definitions actually exclude elements that have fewer than N elements. (closed variant)
I could change the PR to one of those variants, or provide both?

ericniebler: I favor following existing practice (closed variant), which also agrees with my intuition, FWIW. That would seem to either (a) require BidirectionalIterators, or (b) do an O(M) probe for the "last-N" position, cacheing the result, from `begin`. Option (b) would be consistent with what the other views do.

tobimpub: Ok, i think i need a little extra guidance here.
I was able to come up with a working implementation of option (a), and also an extended version that works for ForwardRanges (This just caches a trailing iterator that gets pulled along with every `next()`, `prev()` and `advance()`). Unfortunately, I could not figure out how to partially spezialize the adaptor so that one or the other would be used depending on the Concept mechanism.
I'm also struggling with option (b). How can I use the cached iterator? Using that as the return from `end()` breaks the tests.

ericniebler: Sorry for not getting back sooner. I'll have more time after the post meeting mailing in 2 weeks.

ericniebler:
This is superseded by #522. @tobimpub, pls comment there if you disagree. I'll close this one once #522 gets merged.

tobimpub:
@ericniebler i'm fine with that.  https://github.com/ericniebler/range-v3/pull/522 looks good.

Issue #486, opened 2016-10-21
Make range-v3 views work with Boost


tonyelewis:
I think it would be helpful if range-v3 played nicely with Boost. If I try this:

``` .cpp
#include <range/v3/all.hpp>

#include <boost/algorithm/string/join.hpp>

#include <iostream>

int main() {
    const std::vector<std::string> a = { "underground",
                                         "overground" };

    const std::vector<std::string> b = { "wombling",
                                         "free" };

    std::cerr << boost::algorithm::join( ranges::view::concat( a, b ), " " ) << "\n";
}
```

I get the following compile errors, which AFAICT are due to the relevant range-v3 types not having the typedefs (eg `iterator`, `const_iterator`, `value_type`) that Boost expect ranges to have.

```
In file included from range-v3_boost_interop.cpp:3:
In file included from /opt/include/boost/algorithm/string/join.hpp:16:
/opt/include/boost/range/value_type.hpp:26:70: error: no type named 'type' in 'boost::range_iterator<ranges::v3::concat_view<ranges::v3::iterator_range<std::__1::__wrap_iter<const std::__1::basic_string<char> *>, std::__1::__wrap_iter<const std::__1::basic_string<char> *> >, ranges::v3::iterator_range<std::__1::__wrap_iter<const std::__1::basic_string<char> *>, std::__1::__wrap_iter<const std::__1::basic_string<char> *> > >, void>'
    struct range_value : iterator_value< typename range_iterator<T>::type >
                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
/opt/include/boost/algorithm/string/join.hpp:45:25: note: in instantiation of template class 'boost::range_value<ranges::v3::concat_view<ranges::v3::iterator_range<std::__1::__wrap_iter<const std::__1::basic_string<char> *>, std::__1::__wrap_iter<const std::__1::basic_string<char> *> >, ranges::v3::iterator_range<std::__1::__wrap_iter<const std::__1::basic_string<char> *>, std::__1::__wrap_iter<const std::__1::basic_string<char> *> > > >' requested here
        inline typename range_value<SequenceSequenceT>::type 
                        ^
[...]

PR #495, opened 2016-10-31
Specialise Boost Range metafns for range-v3 views

tonyelewis:
This allows range-v3 views to be treated as ranges by the Boost Range
library and other related Boost libraries.

This is an attempt to tackle ericniebler/range-v3#486.

### DIFF
--- a/include/range/v3/detail/boost_range_hook.hpp
+++ b/include/range/v3/detail/boost_range_hook.hpp
@@ -1,0 +1,48 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014
+//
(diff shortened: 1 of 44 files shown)


ericniebler:
Looks good in general. My only quibble would be that all range-v3 macros should start with `RANGES_`, so I suggest `RANGES_SATISFY_BOOST_RANGE` instead of `BOOST_RANGE_HOOK`.

tonyelewis:
Yes. That's better. I've pushed that change.

ericniebler:
Thanks!

tonyelewis:
You're welcome. Thanks for engaging with the proposal. :)

PR #496, opened 2016-10-31
difference_type to spec + ericniebler/stl2#243

CaseyCarter:


include/range/v3/utility/associated_types.hpp
```diff
@@ -29,52 +29,41 @@ namespace ranges
...
         ////////////////////////////////////////////////////////////////////////////////////////
+        /// \cond
+        namespace detail
+        {
+            template<typename T, typename Enable = void>
+            struct difference_type2
+            {};
+
+            template<typename T>
+            struct difference_type2<T, meta::void_<decltype(std::declval<const T>() - std::declval<const T>())>>
```
ericniebler: What if `a-b` is not `Integral`?

ericniebler: So, just because something can be subtracted, `difference_type` should work for that type? That is not what we have currently specified. And `make_signed` is unlikely to so anything sane with a type that is not `Integral`.

ericniebler: I prefer `struct difference_type<T, meta::if_<std::is_integral<decltype(std::declval<const T>() - std::declval<const T>()) >>>`

CaseyCarter: > That is not what we have currently specified. And `make_signed` is unlikely to do anything sane with a type that is not `Integral`.

*sigh* I'm really on the ball today. Fixing.

CaseyCarter: > `struct difference_type<T, meta::if_<`

I hope you mean `struct difference_type2`: I would hate to manually disambiguate this specialization with all of the `difference_type` specializations (and trigger https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78173 in the process).

PR #508, opened 2016-11-18
some toolchains (e.g. Android NDK) might not support options -march=n…

SSE4:
…ative or -mtune=native, so check for them


SSE4:
#507

gnzlbg:
See this comment: 

https://github.com/ericniebler/range-v3/issues/507#issuecomment-261488391

I'd rather add a macro and use that instead for all flags.  Something like this should do:

```cmake
include(CheckCXXCompilerFlag)
macro(range_v3_append_flag testname flag)
    check_cxx_compiler_flag(${flag} ${testname})
    if (${testname})
        add_compile_options(${flag})
    endif()
endmacro()
range_v3_append_flag(RANGE_V3_HAS_MARCH_NATIVE "-march=native")
range_v3_append_flag(RANGE_V3_HAS_MTUNE_NATIVE "-mtune=native")
```

This is predestined to happen again.

SSE4:
@gnzlbg thanks for the feedback - I've updated PR according to your suggestions

ericniebler:
Looks like this broke the build.

SSE4:
@ericniebler yeah, seems like CMake version on CI is older and doesn't support add_compile_options (which was added in 2.8.12). I'll replace this with settings CMAKE_CXX_FLAGS.

CaseyCarter:
If you're mucking about in CMakeLists.txt, you should simply remove the `-march=native` flag: it is implied by `-march=native`.

SSE4:
@CaseyCarter which one - mtune or march?

CaseyCarter:
> @CaseyCarter which one - mtune or march?

sorry typo: `-march=foo` implies `-mtune=foo`.

CaseyCarter:
> `-march=foo` implies `-mtune=foo`.

Some quick research in the GCC manual demonstrates this is true only for x86; on other architectures it does not hold. Ignore me and leave `-mtune` there after all.

SSE4:
@CaseyCarter sure
@ericniebler I've updated PR replacing add_compile_options via set(CMAKE_CXX_FLAGS ...)

ericniebler:
Thanks all!

PR #509, opened 2016-11-18
Adjust concepts as in P0441

CaseyCarter:


include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -198,7 +198,6 @@ namespace ranges
                 template<typename I, typename O>
                 auto requires_(I&&, O&&) -> decltype(
                     concepts::valid_expr(
-                        concepts::model_of<IndirectlyMovable, I, O>(),
```
ericniebler: Huh, I wonder how that snuck in.

PR #511, opened 2016-11-19
properly constrain uses of iota_minus_ in view/iota.hpp; fixes #506

### LINKED ISSUE #506: Underconstrained functions in iota view
The iota views take two different types, `From` and `To`, and call `iota_minus(to_, from_)` multiple times, e.g., [here](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/iota.hpp#L261). 

This requires `Same<From, To>` since otherwise `iota_minus` cannot deduce `Val` but this constraint is not expressed.

We have lots of options:

- constrain all the functions that call `iota_minus(to_, from_)` on `Same<From, To>`, 
- allow `iota_minus` to work with two different types,
- or force a conversion to one of the types, e.g.,  by using`detail::iota_minus<From>(to_, from_)`. 

No idea what the right way to fix this is. I would say that `iota_minus` ought to work with different types, or otherwise it doesn't really make much sense to have all the iota views take two different types. 

EDIT: FYI there are no tests for the iota views taking two different types.

ericniebler:


CaseyCarter:
`closed_iota_view::equal` needs to check `done_` as well as `from_`:

```c++
CONCEPT_REQUIRES(Incrementable<From>())
bool equal(closed_iota_view const &that) const
{
    return that.from_ == from_ && that.done_ == done_;
}
```

PR #517, opened 2016-12-02
Disambiguate value_type for cv-qualified types with member value_type…

CaseyCarter:
… or element_type

### DIFF
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -162,7 +162,7 @@ namespace ranges
         template<typename T, typename = void>
         struct difference_type;
 
-        template<typename T, typename = void>
+        template<typename T>
         struct value_type;
 
         template<typename T, typename = void>
--- a/include/range/v3/utility/associated_types.hpp
+++ b/include/range/v3/utility/associated_types.hpp
@@ -1,7 +1,7 @@
 /// \file
 // Range v3 library
 //
-//  Copyright Eric Niebler 2013-2014
+//  Copyright Eric Niebler 2013-2014, 2016
 //
 //  Use, modification and distribution is subject to the
 //  Boost Software License, Version 1.0. (See accompanying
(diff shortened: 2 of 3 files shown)


ericniebler:
Most definitely fixes a bug (thanks!), but this is heavy at compile-time. This seems to work (local tests still running), and is probably lighter weight:

```c++
/// \cond
namespace detail
{
    template <typename T, std::size_t N>
    std::remove_cv<T> value_type_helper(T (*)[N]);

    template <typename T>
    std::remove_cv<T> value_type_helper(T (*)[]);

    template <typename T>
    meta::if_<std::is_object<T>, std::remove_cv<T>> value_type_helper(T **);

    template <typename T>
    std::remove_reference<typename T::value_type &> value_type_helper(T *);

    template <typename T>
    std::remove_reference<typename T::element_type &> value_type_helper(T *);

    template <typename T>
    meta::if_<std::is_base_of<std::ios_base, T>, meta::id<typename T::char_type>>
    value_type_helper(T *);

    template <typename T>
    using value_type_ = meta::_t<decltype(detail::value_type_helper((T *)nullptr))>;
}
/// \endcond

////////////////////////////////////////////////////////////////////////////////////////
template<typename T>
struct value_type
  : meta::defer<detail::value_type_, T>
{};

template<typename T>
struct value_type<T const>
  : meta::defer<detail::value_type_, T>
{};
```

Here's some quick tests I wrote. Probably needs more.

```c++
struct S { using value_type = int; };
CONCEPT_ASSERT(Same<int, ranges::value_type<S>::type>());
CONCEPT_ASSERT(Same<int, ranges::value_type<S const>::type>());
CONCEPT_ASSERT(Same<int, ranges::value_type<int*>::type>());
CONCEPT_ASSERT(Same<int, ranges::value_type<int*const>::type>());
CONCEPT_ASSERT(Same<int, ranges::value_type<int[]>::type>());
CONCEPT_ASSERT(Same<int, ranges::value_type<std::shared_ptr<int> const>::type>());
CONCEPT_ASSERT(Same<char, ranges::value_type<std::ostream>::type>());
static_assert(!meta::is_trait<ranges::value_type<int*&>>::value, "");
static_assert(!meta::is_trait<ranges::value_type<std::shared_ptr<void>>>::value, "");
```

CaseyCarter:
```c++
template <typename T>
std::remove_cv<T> value_type_helper(T (*)[]);
```
Function parameters of type "pointer/reference to array of unknown bound" were forbidden by C++ before [CWG393](http://www.open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#393). This declaration is rejected by all released versions of GCC.

CaseyCarter:
Why *do* we support `value_type` for arrays of unknown bound?

ericniebler:
I dunno. Arrays are not move constructible, so they don't satisfy Readable as it's currently defined.

CaseyCarter:
I can imagine some kind of containerish concept using `value_type_t` that could be satisfied by complete array types, but certainly not arrays of unknown bound.

Also, `value_type<T const>` needs to respect specializations of `value_type<T>`:
```diff
 template<typename T>
 struct value_type<T const>
-  : meta::defer<detail::value_type_, T>
+  : value_type<T>
 {};
```

ericniebler:
Ok. Do you want to update this pull request or should I just fix it?

CaseyCarter:
> Do you want to update this pull request or should I just fix it?

I'll update it once I'm sure we've found everything that needs fixing. For example:
```c++
CONCEPT_ASSERT(Same<int const, ranges::value_type<std::shared_ptr<int const>>::type>());
```
`element_type` in `pointer_traits` and in the smart pointers is the type of the object the pointer denotes, *not* the value type of that object (i.e., it can be a cv-qualified type). We should strip cv-qualifiers from `typename T::element_type`:
```diff
 template <typename T>
-std::remove_reference<typename T::element_type &> value_type_helper(T *);
+std::remove_cv<std::remove_reference_t<typename T::element_type &>> value_type_helper(T *);
```

CaseyCarter:
We're *still* not quite to spec: the spec requires `::value_type` and `::element_type` to be object types, and `remove_reference<typename T::value_type &>` fails to do so for object reference types. In fact it incorrectly coerces them into object types. Methinks we should give them the same treatment as the pointer case:
```c++
template<typename T>
using object_remove_cv = meta::if_<std::is_object<T>, std::remove_cv<T>>;

template<typename T>
object_remove_cv<T> value_type_helper(T **);

template<typename T>
object_remove_cv<typename T::value_type> value_type_helper(T *);

template<typename T>
object_remove_cv<typename T::element_type> value_type_helper(T *);
```

CaseyCarter:
..although I'm on the fence about whether we should strip cv-qualifiers from `T::value_type` or hard error via `static_assert` with a YOUR VALUE TYPE IS BAD AND YOU SHOULD FEEL BAD message if `T::value_type` is cv-qualified.

ericniebler:
> ..although I'm on the fence about whether we should strip cv-qualifiers from `T::value_type` or hard error via `static_assert` with a YOUR VALUE TYPE IS BAD AND YOU SHOULD FEEL BAD message if `T::value_type` is cv-qualified.

The latter, or leave it and folks get what they deserve.

PR #520, opened 2016-12-03
attempt to fix libc++ build

EricWF:
Submitting a PR so that Travis will run. Not sure if this is a valid fix yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ script: cmake
 #
 matrix:
   include:
-    - env: BUILD_TYPE=Release CPP=11 ASAN=Off LIBCXX=On
+    - env: BUILD_TYPE=Release CPP=11 ASAN=Off SYSTEM_LIBCXX=On
       os: osx
       compiler: clang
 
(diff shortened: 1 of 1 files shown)


EricWF:
@CaseyCarter Looks like this patch fixes the issue. The problem was that the C++ STL include directory was out-of-order compared to other system header search paths. The fix is to use `-cxx-isystem` instead of `-I` to add the header path.

Additionally `-Wno-unused-command-line-arguments` is needed because `-nostdinc++` causes `-stdlib=libc++` to be "unused" during preprocessing/compilation.

EricWF:
Boo ya! Travis is green! In summary the fixes in this patch are:

1. Don't enable `LIBCXX=On` on OS X.  This configuration now tests the system libc++ as opposed to ToT libc++.

2. Add the custom libc++ include path using `-cxx-isystem <path>` instead of `-I<path>`. Libc++ intercepts the C standard library headers by providing it's own `<foo.h>` wrapper for each `<cfoo>` header. The `foo.h` header then forwards tho the appropriate header using `#include_next <foo.h>`.  When the libc++ header path is added using `-I` the expression `#include_next <stddef.h>` fails to find the compiler builitin <stddef.h> since builtin header paths are searched *previous* to path's specified with `-I` .

NOTE: When attempting to override STL headers (for either GCC or Clang) you must specify `-nostdinc++` to omit the default include paths.

The finicky libc++ headers  are really a shame. On the plus side they work great with Clang Modules! (I would expect an 80% speedup for the ranges-v3 test-suite with modules enabled. (that's what I see with libc++)).

ericniebler:
Thank you, Eric!

gnzlbg:
> (I would expect up to an 80% speedup for the ranges-v3 test-suite with modules enabled. (that's what I see with libc++)).

@EricWF How do you enable modules for libc++? Is that in libc++ trunk already? I am trying to find the bits in the CMake config that enable them without any luck.

EricWF:
> @EricWF How do you enable modules for libc++?

The main work in setting up modules is writing the [`module.modulemap`](http://clang.llvm.org/docs/Modules.html#module-maps). After you've done that all it takes is adding `-fmodules` when building code (Also add `-Xclang -fmodules-local-submodule-visibility` on Linux but not OS X). Clang will then automatically build and cache the modules the first time it encounters them.

The [libc++ module.modulemap](https://github.com/llvm-mirror/libcxx/blob/master/include/module.modulemap) is in-tree.

PR #521, opened 2016-12-05
Implement non_propagating_cache:

CaseyCarter:
A rule-of-five wrapper for an optional iterator that points into a contained range. Does not copy/move construct/assign the iterator, but *does* invalidate the cached value when assigned, or when used as the source of a move (i.e., invalidates the cache when the range into which it points has presumably been assigned to / moved from.)

Use to simplify the "caching" views:
* adjacent_remove_if
* cycle
* drop
* drop_exactly
* drop_while
* remove_if
* reverse
* slice

### DIFF
--- a/include/range/v3/detail/optional.hpp
+++ b/include/range/v3/detail/optional.hpp
@@ -70,6 +70,46 @@ namespace ranges
                 ranges::emplace<0>(data_);
             }
         };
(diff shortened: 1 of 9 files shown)


ericniebler:
Why didn't I think of that?

PR #522, opened 2016-12-05
Sliding window

CaseyCarter:
A bit of polish on Tobias's sliding view from #488, rebased onto my `non_propagating_cache` change from #521.

### DIFF
--- a/include/range/v3/view/sliding.hpp
+++ b/include/range/v3/view/sliding.hpp
@@ -1,0 +1,183 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2014
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_SLIDING_HPP
+#define RANGES_V3_VIEW_SLIDING_HPP
+
+#include <utility>
+#include <functional>
+#include <meta/meta.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/iterator_range.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/view_adaptor.hpp>
+#include <range/v3/view/all.hpp>
+#include <range/v3/view/view.hpp>
+#include <range/v3/view/take.hpp>
+#include <range/v3/utility/box.hpp>
+#include <range/v3/utility/static_const.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        /// \addtogroup group-views
+        /// @{
+        template<typename Rng>
+        struct sliding_view
+          : view_adaptor<
+                sliding_view<Rng>,
+                Rng,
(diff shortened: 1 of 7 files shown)


CaseyCarter:
Poking @tobimpub for review: shout if you see something I've broken.

ericniebler:
Valgrind is unhappy. ??? Doesn't look like it could possibly be your fault, though. :-/

include/range/v3/view/sliding.hpp
```diff
@@ -0,0 +1,269 @@
...
+                sliding_view<Rng>,
+                !(RandomAccessRange<Rng>() && SizedRange<Rng>())>
+        {
+        private:
+            CONCEPT_ASSERT(ForwardRange<Rng>());
+            friend range_access;
+            struct adaptor;
+
+            static constexpr bool cache_enabled =
+                !(RandomAccessRange<Rng>() && SizedRange<Rng>());
```
ericniebler: I have it from Andrew that class static members are slow to instantiate. An alias or even a static member function is lighter weight.

CaseyCarter:
I saw the valgrind failure, I'll take a look at it. I also realized in the shower this morning that the view doesn't properly handle the end iterator of bidi bounded ranges.

gnzlbg:
@CaseyCarter 

> I also realized in the shower this morning

That mystical source of wisdom... Lucky to know I am not the only one that is productive in the shower.

tobimpub:
@CaseyCarter sorry for the silence. I took a look and this seems to be a working version of what i was trying to do, so thank you!
Regarding the Bidi end iterator: i guess resetting trailing on call to adaptor::end() is not an option?

CaseyCarter:
> Regarding the Bidi end iterator: i guess resetting trailing on call to adaptor::end() is not an option?

The view is basically complete [on this branch in my fork](https://github.com/CaseyCarter/range-v3/blob/sliding_window_fix/include/range/v3/view/sliding.hpp) which I haven't pushed into the PR yet while I'm running down the valgrind error. (It seems to be a compiler bug; only triggers with GCC, only with version 6, only with -O2. Goes away if I don't use `optional` to cache the "`begin() + window_size - 1" iterator.)

There are three different specializations of `sliding_view`:

* The default implementation that requires forward iterators. Caches "begin(base()) + n_ - 1", known as the "first" iterator, in the view. `begin` returns an iterator that adapts this "first" iterator, and stores a "trailing" iterator initialized to the underlying range's `begin`. On dereference, the view iterator returns a counted view of `n_` elements beginning at its trailing iterator. If the underlying iterator is random access, the trailing iterator is not stored - it can be calculated in O(1) by subtracting `n_ - 1` from the current position.

* For bidirectional bounded (`end()` returns an iterator) ranges, I flip everything from left to right: cache `end(base()) - n_ + 1` in the view (known as the "last" iterator), and store the left end of the result range in the view's iterators. `end` returns an iterator that wraps the "last" iterator. This design preserves bidirectionality of the underlying range and never needs to store trailing iterators.

* For random access sized ranges, the view doesn't cache anything and is always a bounded range: it can calculate the "last" iterator in O(1) as `begin(base()) + size(base()) - n_ + 1`.

tobimpub:
> For bidirectional bounded (end() returns an iterator) ranges, I flip everything from left to right: cache end(base()) - n_ + 1 in the view (known as the "last" iterator), and store the left end of the result range in the view's iterators. end returns an iterator that wraps the "last" iterator. This design preserves bidirectionality of the underlying range and never needs to store trailing iterators.

That is more elegant, i admit.

CaseyCarter:
There's a bug# for the GCC6 optimizer bug in the test. This should be ready to merge now.

PR #524, opened 2016-12-07
Rework contract enforcement system:

CaseyCarter:
* `RANGES_ASSERT(pred)`
  * Implement by hand on GCC4, instead of using `assert` from the standard library, so that `RANGES_ASSERT `is usable in `constexpr` functions. (This is not the case for `assert` in libstdc++ for GCC < 5.)

* Add `RANGES_ASSUME(pred)`
  * Has undefined behavior if `pred` is not `true`. (Makes an effort to inform the optimizer that `pred` must hold.)
  * Unlike `RANGES_ASSERT(pred)`, which does not evaluate `pred` in release builds, `pred` is *always* potentially evaluated. It must be inexpensive to evaluate.

* Add `RANGES_EXPECT(pred)`
  * When debugging (`!defined(NDEBUG)`), equivalent to `RANGES_ASSERT(pred)`.
  * When NOT debugging, equivalent to `RANGES_ASSUME(pred)`.

Replace `void()` with `void(0)` throughout the contract system to workaround https://llvm.org/bugs/show_bug.cgi?id=30592.

### DIFF
--- a/include/range/v3/detail/config.hpp
+++ b/include/range/v3/detail/config.hpp
@@ -16,21 +16,70 @@
 #define RANGES_V3_DETAIL_CONFIG_HPP
 
 #include <iosfwd>
+#if (defined(NDEBUG) && !defined(RANGES_ENSURE_MSG)) || \
+    (!defined(NDEBUG) && !defined(RANGES_ASSERT) && \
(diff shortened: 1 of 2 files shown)


ericniebler:
Love this. Docs would rule, but I'm taking it as-is because it's an improvement.

PR #525, opened 2016-12-07
Contracts system followup:

CaseyCarter:
* Replace many uses of `RANGES_ASSERT` with `RANGES_EXPECT`.
* Replace uses of `assert` in tests with `RANGES_ENSURE`.

"many" is actually "most." The majority of assertions provide useful optimizer hints and evaluate expressions that themselves optimize to nothing in release builds, making them excellent candidates for `RANGES_EXPECT`.

### DIFF
--- a/include/range/v3/action/drop.hpp
+++ b/include/range/v3/action/drop.hpp
@@ -63,7 +63,7 @@ namespace ranges
                     CONCEPT_REQUIRES_(Concept<Rng, D>())>
                 Rng operator()(Rng && rng, range_difference_t<Rng> n) const
                 {
-                    RANGES_ASSERT(n >= 0);
+                    RANGES_EXPECT(n >= 0);
                     ranges::action::erase(rng, begin(rng), ranges::next(begin(rng), n, end(rng)));
                     return std::forward<Rng>(rng);
                 }
--- a/include/range/v3/action/slice.hpp
+++ b/include/range/v3/action/slice.hpp
@@ -68,7 +68,7 @@ namespace ranges
                 Rng operator()(Rng && rng, range_difference_t<Rng> from,
                     range_difference_t<Rng> to) const
                 {
-                    RANGES_ASSERT(from <= to);
+                    RANGES_EXPECT(from <= to);
                     ranges::action::erase(rng, next(begin(rng), to), end(rng));
                     ranges::action::erase(rng, begin(rng), next(begin(rng), from));
                     return std::forward<Rng>(rng);
--- a/include/range/v3/action/stride.hpp
+++ b/include/range/v3/action/stride.hpp
@@ -66,7 +66,7 @@ namespace ranges
                 {
                     using I = range_iterator_t<Rng>;
                     using S = range_sentinel_t<Rng>;
-                    RANGES_ASSERT(0 < step);
+                    RANGES_EXPECT(0 < step);
                     if(1 < step)
                     {
                         I begin = ranges::begin(rng);
--- a/include/range/v3/action/take.hpp
+++ b/include/range/v3/action/take.hpp
@@ -64,7 +64,7 @@ namespace ranges
                     CONCEPT_REQUIRES_(Concept<Rng, D>())>
                 Rng operator()(Rng && rng, range_difference_t<Rng> n) const
                 {
-                    RANGES_ASSERT(n >= 0);
+                    RANGES_EXPECT(n >= 0);
                     ranges::action::erase(rng, ranges::next(begin(rng), n, end(rng)), end(rng));
                     return std::forward<Rng>(rng);
                 }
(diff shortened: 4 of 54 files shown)


gnzlbg:
@CaseyCarter In the diff I can only see the cases of assert that were replaced with expect, and github doesn't let me search "forked" repositories...

Could you maybe comment on some of those if they aren't many? In which situations can't assert be replaced with expect?

CaseyCarter:
> Could you maybe comment on some of those if they aren't many? In which situations can't assert be replaced with expect?

Tests of predicates that would have runtime impact in release builds - because the compiler thinks it has side effects, or its too complicated for the analyzer to tell - I left as asserts. I didn't examine assembly for every callsite to perform that determination, this is purely guesswork. [Here's the output from `git grep -nC 12 RANGES_ASSERT` for the 14 occurrences of `RANGES_ASSERT` outside of detail/config.hpp.](https://gist.github.com/CaseyCarter/fd903fed88e72a82a4f7935cfedf6694) This one from iterator_range.hpp is a good example:
```
 98-        public:
 99-            using iterator = I;
100-            using sentinel = S;
101-        #ifndef RANGES_DOXYGEN_INVOKED
102-            using const_iterator = I; // Mostly to avoid spurious errors in Boost.Range
103-        #endif
104-
105-            sized_iterator_range() = default;
106-            RANGES_NDEBUG_CONSTEXPR sized_iterator_range(I begin, S end, iterator_size_t<I> size)
107-              : rng_{detail::move(begin), detail::move(end)}, size_(size)
108-            {
109-            #ifndef NDEBUG
110:                RANGES_ASSERT(!ForwardIterator<I>() ||
111-                    static_cast<iterator_size_t<I>>(ranges::distance(rng_)) == size_);
112-            #endif
113-            }
114-            template<typename X, typename Y,
115-                CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
116-            RANGES_NDEBUG_CONSTEXPR sized_iterator_range(std::pair<X, Y> rng, iterator_size_t<I> size)
117-              : sized_iterator_range{detail::move(rng).first, detail::move(rng).second, size}
118-            {}
119-            template<typename X, typename Y,
120-                CONCEPT_REQUIRES_(Constructible<I, X &&>() && Constructible<S, Y &&>())>
121-            RANGES_NDEBUG_CONSTEXPR sized_iterator_range(iterator_range<X, Y> rng, iterator_size_t<I> size)
122-              : sized_iterator_range{detail::move(rng).first(), detail::move(rng).second, size}
```

I imagine the compiler would have a hard time eliminating

```c++
!ForwardIterator<I>() ||
    static_cast<iterator_size_t<I>>(ranges::distance(rng_)) == size_
```

which would introduce O(N) overhead into what is otherwise an O(1) constructor.

gnzlbg:
> which would introduce O(N) overhead into what is otherwise an O(1) constructor.

How? [The specification of clang's `__builtin_assume`](http://clang.llvm.org/docs/LanguageExtensions.html) says that it never evaluates the arguments, it just assumes them to be true: 

> If the condition is violated during execution, the behavior is undefined. The argument itself is never evaluated, so any side effects of the expression will be discarded.

Isn't this the case for MSVC and GCC? I thought they would all do the same :/

CaseyCarter:
> Isn't this the case for MSVC and GCC? I thought they would all do the same :/

MSVC's `__assume` is similar: it doesn't evaluate the expression. GCC doesn't have an assume, it must be emulated with e.g. `((predicate) ? void() : __builtin_unreachable())`. Since I can't provide the unevaluated semantics on all supported platforms - and I honestly find it offensive that a side effect might be present only in debug builds - I use the same `__builtin_unreachable` formulation on Clang.

gnzlbg:
> and I honestly find it offensive that a side effect might be present only in debug builds

Isn't this how assert works as well? or does RANGE_ASSERT evaluate the expression in release builds as well?

CaseyCarter:
> Isn't this how assert works as well?

Yes, and it (a) leads to subtle bugs when a predicate has side effects, and (b) results in having to cast-to-void parameters that are only used in asserts. Those semantics are too engrained in `assert` to ever change.

gnzlbg:
This LGTM as well. Thanks for the remarks @CaseyCarter .

I don't think that trying to do something "better/cleverer" in clang or MSVC is a good idea (like using the assume builtins there). It would make the execution of side-effects compiler-dependent. It doesn't affect that many cases, and if GCC is the minimum common denominator, that's just the way it is :/

CaseyCarter:
<3 Travis. The failure is one of the GCC builds that failed to install g++-4.9.

rsw0x:
Bit old but,

>Since I can't provide the unevaluated semantics on all supported platforms - and I honestly find it offensive that a side effect might be present only in debug builds - I use the same __builtin_unreachable formulation on Clang.

@CaseyCarter FYI, __builtin_unreachable in Clang is not equivalent to GCC in terms of code generation.
Compare https://godbolt.org/g/7Nw0Es to https://godbolt.org/g/nFLlbw

CaseyCarter:
>  __builtin_unreachable in Clang is not equivalent to GCC in terms of code generation.

That is pretty horrible. How about:
```c++
#define RANGES_ASSUME(...) (void(__VA_ARGS__), __builtin_assume(__VA_ARGS__))
```
which both uses `__builtin_assume` *and* always evaluates?

PR #527, opened 2016-12-08
return prvalues instead of xvalues from get_first & get_second

### LINKED ISSUE #526: view::ints seemingly not counting under optimisation
This code:

```c++
#include <vector>
#include <iostream>
#include <range/v3/all.hpp>

using namespace ranges;
using namespace std;

int main(int c, char **v) {
    vector<int> xs = {42, 100, -1234};

    auto exs = view::zip(view::ints, xs);

    cout << "Iteration over zip:";
    RANGES_FOR(const auto ix, exs)
        cout << " (" << ix.first << ", " << ix.second << ")";

    cout << "\nIteration over keys:";
    RANGES_FOR(const auto i, view::keys(exs))
        cout << " " << i;

    cout << "\n";
    return 0;
}
```
I would expect to produce
```
Iteration over zip: (0, 42) (1, 100) (2, -1234)
Iteration over keys: 0 1 2
```
and indeed it does under `g++ -O0`.  But under `g++ -O1`, or higher `-O`, it produces
```
Iteration over zip: (0, 42) (1, 100) (2, -1234)
Iteration over keys: 0 0 0
```
This is
```
g++ (Ubuntu 6.2.0-5ubuntu12) 6.2.0 20161005
```
Very similar behaviour under `clang`; `-O0` gives the expected output, but `-O1` gives
```
Iteration over zip: (0, 42) (1, 100) (2, -1234)
Iteration over keys: 4204694 4204694 4204694
```
Clang is
```
clang version 3.8.1-12ubuntu1 (tags/RELEASE_381/final)
```

As noted out-of-band by @aldanor, the problem seems to lie with `ints` rather than `keys`, because the problem remains if you swap the order of `zip`'s arguments and iterate over `values`, i.e., if you change the code to
```c++
    auto exs = view::zip(xs, view::ints);
    // ...
    RANGES_FOR(const auto i, view::values(exs))
        cout << " " << i;
```

The other two permutations —
* `zip(ints, xs)` and iterate over `values`;
* `zip(xs, ints)` and iterate over `keys`

— behave correctly.

I have not dug any further, sorry.  Many thanks for any insight!

CaseyCarter:
Fixes #526.

### DIFF
--- a/include/range/v3/view/map.hpp
+++ b/include/range/v3/view/map.hpp
@@ -17,6 +17,7 @@
 #include <utility>
 #include <meta/meta.hpp>
 #include <range/v3/range_fwd.hpp>
+#include <range/v3/utility/concepts.hpp>
 #include <range/v3/utility/functional.hpp>
 #include <range/v3/utility/static_const.hpp>
 #include <range/v3/view/view.hpp>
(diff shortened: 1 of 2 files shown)


aldanor:
Hmm, looks like it doesn't work on gcc 4.8/4.9 in release mode; I can reproduce the failure with gcc 4.8.5:

```sh
169: > ERROR: CHECK failed "*begin0 == *begin1"
169: >  ./range-v3/test/view/../test_utils.hpp(35)
169: >  EXPECTED: 0
169: >  ACTUAL: -76095120
169: > ERROR: CHECK failed "*begin0 == *begin1"
169: >  ./range-v3/test/view/../test_utils.hpp(35)
169: >  EXPECTED: 1
169: >  ACTUAL: -76095120
169: > ERROR: CHECK failed "*begin0 == *begin1"
169: >  ./range-v3/test/view/../test_utils.hpp(35)
169: >  EXPECTED: 2
169: >  ACTUAL: -76095120
1/1 Test #169: test.view.map, ...................***Failed    0.00 sec
```

ericniebler:
I kind of hate C++'s lack of true perfect forwarding. I don't really want xvalues to be converted into prvalues. Then again, I don't know how to distinguish prvalue arguments from xvalue arguments, so there's no way to make this safe. It's annoying.

gnzlbg:
I dont think it is possible, nur would really like to be proven wrong on this one

PR #528, opened 2016-12-09
 * Added support for the Buck build system

njlr:


ericniebler:
I have no easy way to test this at the moment, but if you say it works, that's good enough for me.

PR #530, opened 2016-12-19
GCC6.3 will fix PR78841; disable the workaround...

### LINKED ISSUE #529: GCC 6 optimizer bug in test/view/sliding.cpp
One of the `view::sliding` tests has a workaround for GCC PR78841, which will be fixed in the 6.3 release (Real Soon Now™). When that release occurs, we need to (a) ensure that the workaround is not necessary for 6.3 and (b) restrain the workaround to only be used when `__GNUC_MINOR__` < 3.

CaseyCarter:
... in test/view/sliding.cpp for GCC 6.3+. Fixes #529.

I decided to build my own gcc-6-branch compiler and perform the requisite test.


CaseyCarter:
Also update the PR number from 78841 to 78047.

Issue #451, opened 2016-08-19
Build warnings


aldanor:
If extra warnings are enabled, `ranges` triggers quite a few build warnings which is not cool if the downstream code has tight warnings settings.

Mostly, it's `-Wundef`, like this:

```
range/v3/utility/functional.hpp:186:13: error: "__apple_build_version__" is not defined [-Werror=undef]
         #if __apple_build_version__

range/v3/utility/random.hpp:79:9: error: "__i386__" is not defined [-Werror=undef]
     #if __i386__
```

I think these could be fixed e.g. like so `#if defined(__i386__) && __i386__`. This warning is an especial PITA since it can't be suppressed by pragmas due to long-standing bug in gcc.

or `-Wold-style-cast`:

```
range/v3/algorithm/max.hpp:59:53: error: use of old-style cast [-Werror=old-style-cast]
                         result = (decltype(tmp) &&) tmp;

range/v3/view_interface.hpp:102:41: error: use of old-style cast [-Werror=old-style-cast]
                 return (range_size_t<D>)Cardinality;
```

(and a whole bunch more).

Do we want these addressed (e.g. replacing casts with a more verbose but more modern `static_cast`)? It would be sure nice if the library didn't throw any warnings on both gcc/clang.

gnzlbg:
[Warnings have been discussed a lot of times](https://github.com/ericniebler/range-v3/issues?utf8=%E2%9C%93&q=warnings). Which warnings you use in your own code is up to you, but since _it is impossible_ to make code warning free in all compilers, at the end of the day you should be either including range-v3 as a system library (using `-isystem` instead of `-I`) or using `pragma push/pop` before the range-v3 hearders.

FWIW I think that the `-Wundef` warnings should be fixed (`#if defined(identifier)` is probably enough for most cases). I find old-style casts more convenient to write, but explicit casts more convenient to read.

Which other warnings are you getting?

aldanor:
> FWIW I think that the -Wundef warnings should be fixed (#if defined(identifier) is probably enough for most cases).
I think those should just get fixed for good, too.

> Which warnings you use in your own code is up to you, but since it is impossible to make code warning free in all compilers
But we can consciously try to make it *mostly warning free*.

[...]

Issue #195, opened 2015-07-14
Should we enable -Wshadow?


rhalbersma:
Most algorithm function objects currently shadow the `begin()` and `end()` functions in their `operator()` function parameters `begin` and `end`. This will trigger `-Wshadow` on Clang (currently avoided in CMakeLists.txt through `-Wno-shadow`). Even though the warning is not signalling an error, it still obfuscates the code IMO. Better parameter names would be `first` and `last`. Clearing all this up is rather elaborate, so hence this issue to inquire whether I should undertake this cleanup.

gnzlbg:
This was also discussed in #2 .

rhalbersma:
Yeah, sorry, I should have searched the archive first. Let's focus the discussion to shadowing only here. Both libc++ and libstdc++ don't shadow `begin`/`end` (and it would surprise me if MSVC wouldn't do the same) and `first`/`last` is much easier on the eyes. For one, it's how all STL documentation has named its parameters as well. Plus, if (when?) range-v3 is standardized, the STL implementors will have to go through the cleanup as well. Dito for `size` parameters which could be renamed to `n`. Note that this is something else than postfixing ugly underscores to names as proposed in #2 .

ericniebler:
Shadow warnings don't concern me. If someone submitted a patch that replaced locals `begin`/`end` with `first`/`last`, I would accept it. No strong feelings about `size` --> `n`.

ericniebler:
No patch appears to be forthcoming. Closing this.

PR #534, opened 2016-12-20
Enable compile with strict clang

bradleygibson:
range-v3 will not compile when compiling under Clang (I am on v4.0.0, but no reason to believe the same won't happen on any version) with -Weverything -Werror*.  I ignored the warnings that the compiler reported, initially at the highest level (all.hpp), but this did not seem to be effective.  I then went one level down (same result), so I ended up disabling warnings in the offending files themselves.

If you know of a better way to do this, let me know, and I will re-submit the PR.

*The exact set of warning switches I am using for my project is `-Weverything -Werror -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-newline-eof -Wno-exit-time-destructors -Wno-weak-vtables`.  This PR suppresses any remaining warnings in range-v3 only.

### DIFF
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -26,6 +26,14 @@
 #pragma GCC diagnostic ignored "-Wunknown-pragmas"
 #pragma GCC diagnostic ignored "-Wpragmas"
 #pragma GCC diagnostic ignored "-Wdocumentation-deprecated-sync"
+
+#pragma clang diagnostic ignored "-Wold-style-cast"
+#pragma clang diagnostic ignored "-Wshadow"
+#pragma clang diagnostic ignored "-Wdocumentation"
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wpadded"
+#pragma clang diagnostic ignored "-Wdocumentation-unknown-command"
+#pragma clang diagnostic ignored "-Wundef"
 #endif
 
 /// \defgroup meta Meta
--- a/include/range/v3/action.hpp
+++ b/include/range/v3/action.hpp
@@ -19,6 +19,9 @@
 #pragma clang diagnostic ignored "-Wshadow"
 #pragma clang diagnostic ignored "-Wdocumentation"
 #pragma clang diagnostic ignored "-Wunused-parameter"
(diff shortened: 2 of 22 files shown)


pfultz2:
Whats the point of ignoring the warnings? The warnings are not enabled in the CI.

gnzlbg:
> range-v3 will not compile when compiling under Clang with -Weverything -Werror

@bradleygibson This is deliberate. A lot of the warnings there don't make much sense. If you try to do the same with libc++ or libstdc++ you will realize that they don't compile with `-Weverything -Werror` either. The difference is that you are probably including your standard library using `-isystem` instead of just `-I`, and warnings are ignored for system libraries.

gnzlbg:
@pfultz2 If we decide compiling `-Weverything -Werror` with clang/gcc/... is something worth pursuing, I don't really see a better way of doing it than with something "like" this PRs (but with  the logic for pop/pushing offending warnings abstracted away to some config files instead of duplicating it on every file). We then would also need to enable `-Weverything -Werror` in Travis CI, and test with _all_ compiler versions in debug and optimized builds, since the set of warnings in `-Weverything` changes form version to version, and some warnings only trigger with/without optimizations (at least in clang).

I don't really care (since I just use `-isystem` for range-v3 and am done with it), but if somebody wants to do the work I wouldn't mind as long as the result is easy to maintain and it doesn't clutter every file with lots of warning disabling/reenabling logic.

bradleygibson:
@gnzlbg agreed.  BTW, some older versions of GCC didn't like this, so that would also need to be factored in to the refactored version.

I think your -isystem suggestion is the most pragmatic for now.

gnzlbg:
Cross-referencing this with the respective issues for the people of the future:  #451 #195 #2

ericniebler:
The diagnostics at these very high warning levels tend to be uninteresting. We could spend all of our time playing whack-a-mole with them for every version of every compiler. I'm seriously not interesting in doing that, and I don't much care to have the range-v3 source code littered with macro work-arounds like in this patch. I've even seen people _introduce bugs_ and make code less efficient and harder to read just to silence some misguided compiler warning. It's the tail wagging the dog.

I think compiling clean with `-Wall -Werror` or the equivalent is a good target to aim for.

PR #539, opened 2016-12-29
Rework Callable concepts to TS spec + ericniebler/stl2#189

CaseyCarter:
* Implement `invoke`, use it instead of `as_function` and `function_type`. Implement `result_of` in terms of `invoke`, use it in place of `std::result_of`. (Put `invoke`, `result_of`, and `reference_wrapper` in a separate invoke.hpp since functional.hpp is getting a bit enormous.)

* Make `Callable` & `IndirectCallable` concepts independent of `Function` concepts.

* Replace uses of `Function` family concepts with `Callable` family concepts.

* `reference_wrapper` updates:
  * Delete constructor from "wrong" value category
  * Use `invoke` in `operator()`

* Properly constrain `variant`'s `visit` and `emplace` functions.

* Remove (unused) stuff from functional.hpp:
  * `reference_of`
  * `function_type`
  * `as_function`
  * `Function` concept family

* Great `Callable` renaming:
  * Rename `Callable` => `Invocable`
  * `RegularCallable` => `RegularInvocable`
  * `CallablePredicate` => `Predicate`
  * `CallableRelation` => `Relation`
  * `IndirectCallable` => `IndirectInvocable`
  * `IndirectRegularCallable` => `IndirectRegularInvocable`
  * `IndirectCallablePredicate` => `IndirectPredicate`
  * `IndirectCallableRelation` => `IndirectRelation`

* Update `Callable` concepts to use perfect forwarding, and update uses correspondingly.

* Correct some algorithm return types:
  * `for_each`: return `tagged_pair` of `{end, fun}`
  * `adjacent_difference` and `partial_sum`: return `tagged_pair` instead of plain `std::pair`.
  * `iota`: return `range_safe_iterator_t<Rng`>

* `generate_view`:
  * Use `unreachable` instead of `default_sentinel`. (Seems a bit more idiomatic for an infinite range.)
  * Try to fully constrain all usage of the generator function's return type.
  * Reuse the `Concept` and "bad" overload of `operator()` in `generate_n_view`.
* `generate_n_view`:
  * Don't store the count in the iterators, instead update the count in the range.

### DIFF
--- a/include/range/v3/action/action.hpp
+++ b/include/range/v3/action/action.hpp
@@ -39,7 +39,7 @@ namespace ranges
                     static auto bind(Ts &&...ts)
                     RANGES_DECLTYPE_AUTO_RETURN
                     (
-                        A::bind(std::forward<Ts>(ts)...)
+                        A::bind(detail::forward<Ts>(ts)...)
                     )
                 };
             };
@@ -49,7 +49,7 @@ namespace ranges
                 template<typename Fun>
                 action<Fun> operator()(Fun fun) const
                 {
-                    return {std::move(fun)};
+                    return {detail::move(fun)};
                 }
             };
 
@@ -65,7 +65,7 @@ namespace ranges
                 friend pipeable_access;
                 template<typename Rng>
                 using ActionPipeConcept = meta::and_<
-                    Function<Action, Rng>,
+                    Invocable<Action&, Rng>,
                     Range<Rng>,
                     meta::not_<std::is_reference<Rng>>>;
                 // Pipeing requires things are passed by value.
@@ -74,7 +74,7 @@ namespace ranges
                 static auto pipe(Rng && rng, Act && act)
                 RANGES_DECLTYPE_AUTO_RETURN
                 (
-                    act.action_(std::move(rng))
+                    invoke(act.action_, detail::move(rng))
                 )
             #ifndef RANGES_DOXYGEN_INVOKED
                 // For better error messages:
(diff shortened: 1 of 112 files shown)


include/range/v3/action/action.hpp
```diff
@@ -86,42 +86,42 @@ namespace ranges
                         "The type Rng must be a model of the Range concept.");
                     // BUGBUG This isn't a very helpful message. This is probably the wrong place
                     // to put this check:
-                    CONCEPT_ASSERT_MSG(Function<Action, Rng>(),
+                    CONCEPT_ASSERT_MSG(Invocable<Action&, Rng>(),
                         "This action is not callable with this range type.");
                     static_assert(!std::is_reference<Rng>(),
-                        "You can't pipe an lvalue into an action. Try using std::move on the argument, "
+                        "You can't pipe an lvalue into an action. Try using detail::move on the argument, "
```
ericniebler: No, no. Please don't tell people to use `detail::move`.

include/range/v3/algorithm/is_sorted.hpp
```diff
@@ -32,26 +32,26 @@ namespace ranges
...
-            /// Works on ForwardViews
+            /// Works on ForwardRanges
             ///
-            /// \pre `Rng` is a model of the `ForwardView` concept
+            /// \pre `Rng` is a model of the `ForwardRange` concept
             /// \pre `I` is a model of the `ForwardIterator` concept
-            /// \pre `S` is a model of the `Sentinel<I>` concept
-            /// \pre `R` is a model of the `Relation<Value_Type<I>>` concept
+            /// \pre `S` and `I` model the `Sentinel<S, I>` concept
+            /// \pre `R` and `projected<I, P>` model the `IndirectRelation<R, projected<I, P>>` concept
```
ericniebler: Thanks!

include/range/v3/action/action.hpp
CaseyCarter: I must have caught this with an overzealous `s/std::move/detail::move/`.

PR #544, opened 2017-01-06
Add operator-> customization to basic_iterator

CaseyCarter:
Readable cursors can implement `arrow() const` to override `operator->`. If the cursor has no arrow member, `operator->` returns the address of the current element for lvalue iterators, and otherwise does not participate in overload resolution.

* Implement conforming `reverse_iterator::operator->`
* Implement conforming `common_iterator::operator->`
  * Hack around GCC PR 77537 for `common_cursor::arrow`

* Drive-by:
  * make `common_iterator` difference work cross-type
  * Implement `iterator - sentinel` for test_iterators.hpp
  * Add diagnostic suppression support for clang's -Wunneeded-member-function


include/range/v3/utility/common_iterator.hpp
```diff
@@ -27,6 +31,105 @@ namespace ranges
...
+              : meta::and_<
+                    std::is_constructible<First, Arg1>,
+                    std::is_constructible<Second, Arg2>,
+                    meta::defer<Constructible, std::pair<First, Second>, std::pair<Arg1, Arg2>>>
+            {};
+
+            template<typename T, typename... Args>
+            using is_nothrow_constructible_hack = meta::and_<
+                pair_construct_hack<T, Args...>,
+                meta::defer<std::is_nothrow_constructible, T, Args...>>;
```
ericniebler: I just threw up in my mouth.

test/utility/reverse_iterator.cpp
```diff
@@ -456,9 +456,18 @@ int main() {
...
+      CHECK ( ri.operator->() == ranges::prev(ri.base()) );
+      ++ri;
+      CHECK ( ri->get() == 1 );
+      CHECK ( (*ri).get() == 1 );
+      CHECK ( ri.operator->() == ranges::prev(ri.base()) );
+      ++ri;
+      CHECK ( ri->get() == 0 );
+      CHECK ( (*ri).get() == 0 );
+      CHECK ( ri.operator->() == ranges::prev(ri.base()) );
+      ++ri;
```
ericniebler: Love this, thanks!

include/range/v3/utility/iterator.hpp
```diff
@@ -657,11 +657,16 @@ namespace ranges
...
+                {
+                    return *arrow();
+                }
+                RANGES_CXX14_CONSTEXPR
+                I arrow() const
                 {
                     I tmp(it_);
-                    return *(--tmp);
+                    --tmp;
+                    return tmp;
```
ericniebler: 👍

include/range/v3/utility/common_iterator.hpp
```diff
@@ -27,6 +31,105 @@ namespace ranges
...
+            template<typename T, typename... Args>
+            using is_nothrow_constructible_hack = std::is_nothrow_constructible<T, Args...>;
+#endif
+
+            template<typename T>
+            class operator_arrow_proxy
+            {
+                T value_;
+            public:
+                template<typename U, CONCEPT_REQUIRES_(pair_construct_hack<T, U>())>
```
ericniebler: Why is this hack needed here but not elsewhere?

include/range/v3/utility/common_iterator.hpp
```diff
@@ -27,6 +31,105 @@ namespace ranges
...
+            template <class T>
+            struct HasMemberOperatorArrow<T, meta::void_<
+                decltype(std::declval<T>().operator->())>>
+              : std::true_type
+            {};
+
+            template<typename I,
+                CONCEPT_REQUIRES_(std::is_pointer<I>::value ||
+                    HasMemberOperatorArrow<I>())>
+            constexpr I operator_arrow_(const I& i, int)
```
ericniebler: `I const&`, please. Here and elsewhere.

include/range/v3/utility/common_iterator.hpp
```diff
@@ -27,6 +31,105 @@ namespace ranges
...
+            }
+
+            template<typename I,
+                CONCEPT_REQUIRES_(std::is_reference<iterator_reference_t<I>>())>
+            auto operator_arrow_(const I& i, long)
+                noexcept(noexcept(*i)) ->
+                decltype(std::addressof(std::declval<iterator_reference_t<I>&>()))
+            {
+                auto && tmp = *i;
+                return std::addressof(tmp);
```
ericniebler: if `*i` is a reference (as is required), this can simply be `std::addressof(*i)`.

CaseyCarter: Since it's almost entirely a backwards compat mechanism, `common_iterator` implements `operator->` when `*i` is a *glvalue*, unlike `move_iterator` which only implements `operator->` for *lvalue* `*i`.  `tmp` is an lvalue here whenever `*i` is a glvalue, as required by `std::addressof`.

include/range/v3/utility/common_iterator.hpp
CaseyCarter: Yeah; when "hack" appears in the name of a component and not just in a comment you know something truly horrible is happening. (Aside: should this be `defer_trait` instead of `defer`?

ericniebler: Either works since both `std::is_nothrow_constructible<T>` and `std::is_nothrow_constructible<T>::type` model `meta::Integer`.

include/range/v3/utility/common_iterator.hpp
ericniebler: 👍

include/range/v3/utility/common_iterator.hpp
CaseyCarter: The hack is here to make `view::zip(...) | view::bounded` work, specifically on test/view/zip.cpp line 214. Now that 6.3 is released - with PR 77537 fixed - I'm hoping the problem will die soon and we'll be able to ignore the fact that unpatched 6.2 is broken.

include/range/v3/utility/common_iterator.hpp
CaseyCarter: Ah, I was thinking of `std::conjunction` which derives from the argument type directly, and not `meta::and_` which always derives from a `meta::bool_` specialization. Both would be correct in either case, but with `std::conjunction` there would be a tiny throughput improvement with `defer_trait` simply from shortening the inheritance hierarchy.

include/range/v3/utility/common_iterator.hpp
CaseyCarter: Curses. I have some left-hugging `&&`-qualifiers in here as well.

PR #549, opened 2017-01-11
header fixes:

### LINKED ISSUE #548: Compilation failed if only join.hpp included
Use case:
```
#include <range/v3/view/join.hpp>

using namespace ranges::view;

int main()
{
    return 0;
}
In file included from ../Ranges/range/v3/numeric.hpp:15:0,
                 from ../Ranges/range/v3/view/join.hpp:23,
                 from ../Ranges/main.cpp:1:
../Ranges/range/v3/numeric/adjacent_difference.hpp:52:25: error: 'in' is not a member of 'ranges::v3::tag'
             tagged_pair<tag::in(I), tag::out(O)>
                         ^
../Ranges/range/v3/numeric/adjacent_difference.hpp:52:25: note: suggested alternatives:
In file included from ../Ranges/range/v3/view/join.hpp:19:0,
                 from ../Ranges/main.cpp:1:
../Ranges/meta/meta.hpp:2698:46: note:   'meta::v1::in'
         using in = not_<empty<find<List, T>>>;
                                              ^
../Ranges/meta/meta.hpp:2698:46: note:   'meta::v1::in'
../Ranges/meta/meta.hpp:2705:42: note:   'meta::v1::lazy::in'
             using in = defer<in, List, T>;
```
Including of all.hpp or algorithm/tagspec.hpp before join.hpp solves the problem. Bug is reproducable with gcc version 4.9.2 (Debian 4.9.2-10) and mingw 5.3.

CaseyCarter:
* adjacent_difference.hpp and partial_sum.hpp need to include algorithm/tagspec.hpp

* invoke.hpp needs to include concepts.hpp

Fixes #548.

### DIFF
--- a/include/range/v3/numeric/adjacent_difference.hpp
+++ b/include/range/v3/numeric/adjacent_difference.hpp
@@ -21,6 +21,7 @@
 #include <range/v3/begin_end.hpp>
 #include <range/v3/range_traits.hpp>
 #include <range/v3/range_concepts.hpp>
+#include <range/v3/algorithm/tagspec.hpp>
 #include <range/v3/utility/functional.hpp>
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/iterator_concepts.hpp>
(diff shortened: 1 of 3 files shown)


ericniebler:
Thanks

PR #551, opened 2017-01-11
include view/set_algorithm.hpp in view.hpp

Mizuchi:
Not sure whether the original code is intended, it looks like overlook to me. test plan: Travis-CI


ericniebler:
Jeez, why is this header called `view/set_algorithm.hpp`, anyway? It should be `view/set_view.hpp` or just `view/set.hpp`. Or broken up into `view/set_intersection.hpp`, `view/set_difference.hpp`,... We've inherited weirdness from Boost.Range.

I'll merge this for now (once the tests pass), but I wouldn't mind another patch that fixed the header names in some sane way.

PR #552, opened 2017-01-11
Add operator++(int) customization to basic_iterator

CaseyCarter:
* Necessary to implement TS-conformant `istream_iterator` and `istreambuf_iterator`

### DIFF
--- a/include/range/v3/range_access.hpp
+++ b/include/range/v3/range_access.hpp
@@ -125,6 +125,14 @@ namespace ranges
                         concepts::model_of<concepts::SignedIntegral>(c.distance_to(s))
                     ));
             };
+            struct HasCursorPostincrement
+            {
+                template<typename T>
+                auto requires_(T && t) -> decltype(
+                    concepts::valid_expr(
+                        (t.post_increment(), concepts::void_)
+                    ));
+            };
             struct OutputCursor
               : concepts::refines<WritableCursor, Cursor(concepts::_1)>
             {};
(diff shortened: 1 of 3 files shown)


include/range/v3/range_access.hpp
```diff
@@ -247,6 +255,12 @@ namespace ranges
             (
                 pos.next()
             )
+            template<typename Cur>
+            static RANGES_CXX14_CONSTEXPR auto post_increment(Cur & pos)
+            RANGES_DECLTYPE_AUTO_RETURN
+            (
+                pos.post_increment()
+            )
```
ericniebler: I gotta say that I hate that the name of the customization points are `next()` and `post_increment()`. No symmetry. I actually prefer `next()` and `next(int)`, even though the `int` hack is not necessary here.

CaseyCarter: Keeping in mind that I also hate the name `post_increment`: it is intentional that `next` and `post_increment` are asymmetric, because they are not completely orthogonal. When the cursor does not have `post_increment` and does have `next`, `next` is used to implement both the prefix and postfix increment operators. If the two cursor operations had nicely symmetric names like `next` and `next(int)` I'm concerned that users will get the impression that they must implement both. I *want* `post_increment` to have a long, crappy name to deter people from implementing it unnecessarily.

That said, I'm not convinced that the long, crappy name approach is the best way to address the problem.

ericniebler:
I'm not comfortable with this. When designs get this ugly, it usually indicates that it's time for a rethink. Or that we're expecting too much from the generic component. Maybe it's OK that not every valid iterator can be created with the help of `basic_iterator`. Or maybe there is a better set of Cursor concepts that better cover the design space. This needs more thought.

PR #557, opened 2017-01-24
Add view::shared

FloopCZ:
**Description:** view::shared was designed to allow the construction of temporary containers in the middle of the range pipeline (such as inside view::for_each or view::transform). Such temporary containers can be afterwards used in the remaining part of the pipeline, without the danger of dangling references. view::shared allows elegant constructions, which would be otherwise very difficult to express and the code readability would be significantly degraded (see examples at the end of this description).

Internally, view::shared stores only std::shared_ptr to the underlying range, and therefore construction, destruction, copying and moving of view::shared are still O(1)-time operations, except for the last destruction, which can take up to O(N) time. The linear complexity of the last destruction, however, can be amortized with the time needed to construct the underlying container.

It can be constructed:
- from an rvalue reference of a range
- from std::shared_ptr of a range
- as a function
- as a range pipeline element

**Example 1:** Apply a transformation function returning a vector, to a range of integers, and flatten the result.

```
std::vector<int> foo(int a);
auto rng =
    view::iota(1)
  | view::transform(foo)
  | view::transform(view::shared)
  | view::join;
```

**Example 2:** Build a range which repeats the numbers from a pre-specified vector, but in each epoch, the numbers are randomly shuffled (as is usual e.g., for the training data in machine learning).

```
std::vector<int> base{5, 1, 8};
auto rng =
    view::repeat(base)
  | view::for_each([](std::vector<int> tmp) {
        return std::move(tmp)
             | action::shuffle(gen)
             | view::shared;
    });
```

**Stand-alone views:** Naturally, view::shared can be used to encapsulate the initial data structure inside a view without bothering with the structure's scope. Therefore, it allows the creation of stand-alone views inside of functions:

```
auto build_view()
{
    std::vector<int> base{5, 1, 8};
    return std::move(base) | view::shared | ...
}
```


FloopCZ:
An additional improvement would be to build the view::shared automatically (e.g., whenever the input range of a view is an rvalue reference of a non-view). That would completely solve the problem of temporary containers.

One could directly apply functions returning containers:
```
std::vector<int> foo(int a);
auto rng = view::iota(1) | view::transform(foo) | view::join;
```
Build stand-alone views in-place:
```
auto rng = std::vector<int>{2, 3, 5, 7} | view::transform(foo) | view::join;
```
And yield from temporary containers:
```
auto rng = view::iota(1) | view::for_each(foo);
```

What is your opinion on such a functionality?

gnzlbg:
> What is your opinion on such a functionality?

Can you provide two side-by-side examples of how the code would look like with/without this functionality?

FloopCZ:
Sure. Without the automatic construction, one has to insert the view::shared right after each rvalue referenced container. So two of the previous examples would look like this:

Direct application of functions returning containers and flattening the result:
```
std::vector<int> foo(int a);
// before
view::iota(1) | view::transform(foo) | view::transform(view::shared) | view::join;
// after
view::iota(1) | view::for_each(foo);
```

Build stand-alone views in-place:
```
std::vector<int> foo(int a);
// before
std::vector<int>{2, 3, 5, 7} | view::shared | ...
// after
std::vector<int>{2, 3, 5, 7} | ...
```

And if you are asking for functionality without the view::shared whatsoever, than please refer to the @CaseyCarter 's [code snippet on StackOverflow](http://stackoverflow.com/a/39810436/4932728). I think his words characterize the alternative approach quite well: "This isn't novice-friendly. Heck, it isn't expert-friendly.". :-)

ericniebler:
Just acknowledging that this is on my radar. Casey and I are busy right now preparing for the upcoming committee meeting, so we're not likely to have time to look at this in detail until we get back in mid-Feb. Apologies.

FWIW, I posted a rather straightforward solution to the original stackoverflow question here: http://stackoverflow.com/a/41858405/195873.

FloopCZ:
Great, thanks. The code you provided on StackOverflow is basically what this pull request does, minus the size() function, tests, etc. I look forward to hearing from you!

FloopCZ:
Squashed and rebased on v0.2.0.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+    {
+        // test bidirectional range, which does not know its size
+        auto base_list = std::list<int>{1, 2, 3};
+        auto iter_range = make_iterator_range(begin(base_list), end(base_list));
+        auto list_view = std::move(iter_range) | view::shared;
+
+        has_type<int &>(*begin(list_view));
+        ::models_not<concepts::SizedRange>(list_view);
+        ::models<concepts::BidirectionalRange>(list_view);
+        ::models<concepts::BidirectionalIterator>(begin(list_view));
```
CaseyCarter: Ditto redundant.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    {
+        // test random access range
+        auto vec_view = std::vector<int>{1, 2, 3} | view::shared;
+
+        CHECK(vec_view.size() == 3u);
+        has_type<int &>(*begin(vec_view));
+        ::models<concepts::SizedRange>(vec_view);
+        ::models<concepts::RandomAccessRange>(vec_view);
+        ::models<concepts::RandomAccessIterator>(begin(vec_view));
```
CaseyCarter: Ditto redundant.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        CHECK(vec_view[0] == 1);
+        CHECK(vec_view[1] == 2);
+        CHECK(vec_view[2] == 3);
+
+        // test random access range iterator
```
CaseyCarter: This test is redundant with the above.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        auto vec_iter = begin(vec_view);
+        CHECK(*vec_iter == 1);
+        CHECK(vec_iter[0] == 1);
+        CHECK(vec_iter[1] == 2);
+        CHECK(vec_iter[2] == 3);
+    }
+
+    {
+        // test const random access range
```
CaseyCarter: What coverage does this provide over the previous test? Nothing in the view definition is sensitive to the constness of the range type or its elements.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        // test const random access range iterator
+        auto vec_iter = begin(vec_view);
+        CHECK(*vec_iter == 1);
+        CHECK(vec_iter[0] == 1);
+        CHECK(vec_iter[1] == 2);
+        CHECK(vec_iter[2] == 3);
+    }
+
+    {
+        // check ranges::for_each
```
CaseyCarter: Redundant. We've already tested iteration.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        auto vec_view = std::vector<int>{1, 2, 3} | view::shared;
+        std::vector<int> vec_view_copy;
+        for_each(vec_view, [&](int a) { vec_view_copy.push_back(a); });
+        ::check_equal(vec_view, vec_view_copy);
+    }
+
+    {
+        // check RANGES_FOR
```
CaseyCarter: Ditto redundant.

FloopCZ:
Thank you for the review @CaseyCarter!

> I'm not positive that putting memory allocation and views together is the right design, but I do think that it's interesting enough to put in the library for some usage experience.

Maybe there is a better alternative. What if the whole range-v3 library would consider a shared_ptr of a range to be a range? Similarly as it does with std::reference_wrapper. Then the user would be in charge of memory allocation (he would have to explicitly call std::make_shared somewhere), but all the library functions would just work and deallocate the the shared object properly.

> I do think that constructing view::shared implicitly is a terrible idea. We don't want to hide allocation and or copying of ranges from users: it should be something explicit.

Agreed. It would have been solved by the previous note though.

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,154 @@
...
+        struct shared_view
+          : view_interface<shared_view<T>>
+        {
+        private:
+            // shared storage
+            std::shared_ptr<T> rng_ptr_;
+
+            // range of the shared storage
+            using range_t = view::all_t<T>;
+            range_t rng_;
```
FloopCZ: I also wanted to call the stored range directly, but I was getting SEGFAULTs in tests. It was a bug in view::cycle though (#595), so I will do it as you say.

include/range/v3/view/shared.hpp
```diff
@@ -91,49 +85,49 @@ namespace ranges
...
 
-                template<typename T,
-                    CONCEPT_REQUIRES_(Range<T>()),
-                    typename std::enable_if<std::is_rvalue_reference<T&&>::value, bool>::type = 0>
-                shared_view<typename std::remove_reference<T>::type> operator()(T && t) const
+                template<typename Rng,
+                    CONCEPT_REQUIRES_(Range<Rng>()
+                                      && !View<Rng>()
+                                      && std::is_rvalue_reference<Rng&&>::value)>
+                shared_view<typename std::remove_reference<Rng>::type> operator()(Rng && t) const
```
CaseyCarter: `std::is_rvalue_reference_v<Rng&&>` implies `std::is_same_v<Rng, meta::_t<std::remove_reference<Rng>>>`, so `typename std::remove_reference<Rng>::type` here (and on 112) can simply be `Rng`.

CaseyCarter:
> What if the whole range-v3 library would consider a shared_ptr of a range to be a range? Similarly as it does with std::reference_wrapper.

Pretending that pointers are references is rarely a good idea. `shared_ptr`s already have their own set of behaviors that don't match the semantics we expect of range types. For example, two `shared_ptr`s compare equal only if they point at the same thing, whereas most range types compare equal if they denote ranges with equal elements. There's too much semantic mismatch.

Honestly, I'm still on the fence about `reference_wrapper`. Every time we hack up some library component to treat `reference_wrapper` specially - `bind`, `make_pair`, etc. - we make it that much harder for users to provide their own replacement for `reference_wrapper`.  I haven't proposed ripping the `reference_wrapper` support out of range-v3, though, because I haven't yet convinced myself that the level of semantic mismatch is significant.

include/range/v3/view/shared.hpp
ericniebler: Why the convoluted `std::is_rvalue_reference<Rng&&>::value` and not `!std::is_reference<Rng>::value`?

ericniebler:
> we make it that much harder for users to provide their own replacement for `reference_wrapper`.

Do we have evidence that this is a thing users want to do? My rationale is that `reference_wrapper` already _is_ special. It is intended to behave as much like a `T&` as possible. I'm not sure yet that I would go so far as to say that _anywhere_ a `Concept<T&>` is satisfied, `Concept<reference_wrapper<T>>` should be satisfied also; there are obvious exceptions (`Constructible`). But for the domain-specific abstraction (e.g. `Range`) it seems reasonable to me.

CaseyCarter:
> Do we have evidence that this is a thing users want to do?

https://github.com/ericniebler/range-v3/blob/master/include/range/v3/utility/invoke.hpp#L133

> But for the higher-level domain-specific abstractions (e.g. `Range`) it seems reasonable to me.

Yeah, like I said, I have only a sense of uneasiness - not a concerete argument.

ericniebler:
> > Do we have evidence that this is a thing users want to do?
> 
> https://github.com/ericniebler/range-v3/blob/master/include/range/v3/utility/invoke.hpp#L133

No fair! :-) I had to provide dupes of many `std::` facilities because they weren't meeting the needs of a constrained library. But I suppose that's your point. :-P

FloopCZ:
@CaseyCarter you have a point. For now, it may be better to just require the users to write view::shared explicitly. It is not such a pain.

P. S. All the review requests have been implemented. Thank you both for these.

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        ::check_equal(vec_view, {1, 2, 8, 8, 8});
+    }
+
+    {
+        // test bidirectional range
+        auto list_view = std::list<int>{1, 2, 3} | view::shared;
+
+        CHECK(list_view.size() == 3u);
+        has_type<int &>(*begin(list_view));
+        ::models<concepts::SizedRange>(list_view);
```
FloopCZ: done

test/view/shared.cpp
```diff
@@ -0,0 +1,294 @@
...
+        CHECK(*(--list_iter) == 1);
+    }
+
+    {
+        // test random access range
+        auto vec_view = std::vector<int>{1, 2, 3} | view::shared;
+
+        CHECK(vec_view.size() == 3u);
+        has_type<int &>(*begin(vec_view));
+        ::models<concepts::SizedRange>(vec_view);
```
FloopCZ: done

FloopCZ:
@ericniebler @CaseyCarter Any new thoughts or requests on this? To me, the shared view has actually proven to be very useful during the last few months.

ericniebler:
Awfully sorry to keep you waiting (and waiting) on this. I _still_ haven't had time to think deeply about it. It is a new direction for Ranges and so not something I can comfortably accept without some deep thought.  I'm using my spare time to prepare for an upcoming talk. I'm hopeful that after that (Wednesday) I'll have some free cycles to spend on this.

FloopCZ:
Sure thing. Thank you for your reply. :-)

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,148 @@
...
+        {
+            struct shared_fn : pipeable<shared_fn>
+            {
+            public:
+                template<typename Rng,
+                    CONCEPT_REQUIRES_(Range<Rng>())>
+                shared_view<Rng> operator()(std::shared_ptr<Rng> pt) const
+                {
+                    return {std::move(pt)};
+                }
```
ericniebler: I don't love this overload. APIs with `shared_ptr` arguments are almost always poorly designed, in my experience.

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,148 @@
...
+#endif
+
+                template<typename Rng,
+                    CONCEPT_REQUIRES_(Range<Rng>()
+                                      && !View<Rng>()
+                                      && !std::is_reference<Rng>::value)>
+                shared_view<Rng> operator()(Rng && t) const
+                {
+                    return shared_view<Rng>{std::move(t)};
+                }
```
ericniebler: I like that this rejects `View`s.

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,148 @@
...
+                                      || std::is_reference<Rng>::value)>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(Range<Rng>(),
+                        "The object on which view::shared operates must be a "
+                        "model of the Range concept.");
+                    CONCEPT_ASSERT_MSG(!View<Rng>(),
+                        "view::shared cannot be constructed from a view."
+                        " Please copy the original view instead.");
+                    CONCEPT_ASSERT_MSG(std::is_rvalue_reference<Rng&&>::value,
```
ericniebler: Better to mirror the constraint here with `CONCEPT_ASSERT_MSG(!std::is_reference<Rng>::value,...`.

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,148 @@
...
+            range_size_type_t<Rng> size()
+            {
+                return ranges::size(*rng_ptr_);
+            }
+
+            // shared storage access
+            const std::shared_ptr<Rng> & get_shared() const
+            {
+                return rng_ptr_;
+            }
```
ericniebler: Strong preference to remove this API. `shared_ptr` is only one way to implement this functionality. Don't tie us to one implementation detail.

include/range/v3/view/shared.hpp
```diff
@@ -0,0 +1,148 @@
...
+            // shared storage
+            std::shared_ptr<Rng> rng_ptr_;
+
+        public:
+            shared_view() = default;
+
+            // construct from a shared_ptr
+            shared_view(std::shared_ptr<Rng> pt) noexcept
+              : rng_ptr_{std::move(pt)}
+            {}
```
ericniebler: Remove me.

ericniebler:
> **Example 1**: Apply a transformation function returning a vector, to a range of integers, and flatten the result.
> 
>     std::vector<int> foo(int a);
>     auto rng =
>         view::iota(1)
>       | view::transform(foo)
>       | view::transform(view::shared)
>       | view::join;

I understand why you would want this to work, but it is guaranteed to do an allocation (`make_shared`) at each step. This would perform far better (if the range isn't infinite):

```
std::vector<std::vector<int>> tmp = view::indices(42) | view::transform(foo);
auto rng = tmp | view::join;
```

If you need to handle infinite ranges, then `view::shared` seems like an ok solution here.

> **Example 2**: Build a range which repeats the numbers from a pre-specified vector, but in each epoch, the numbers are randomly shuffled (as is usual e.g., for the training data in machine learning).
> 
>     std::vector<int> base{5, 1, 8};
>     auto rng =
>         view::repeat(base)
>       | view::for_each([](std::vector<int> tmp) {
>             return std::move(tmp)
>                  | action::shuffle(gen)
>                  | view::shared;
>         });

Like Example 1, this is doing lots of unnecessary allocations. `view::shared` it making it too easy to come up with the wrong solution. This is better:

    std::vector<int> base{5, 1, 8};
    std::mt19937 gen;
	auto rng =
		view::generate([=,&gen]() mutable -> decltype(auto) {
			base |= action::shuffle(gen);
			return std::as_const(base);
		})
	  | view::join;

Yes, this view "owns" a `vector`, but the range is infinite, and copying it is not O(N) so it's fine.

> Therefore, it allows the creation of stand-alone views inside of functions:

I don't really understand this example.

include/range/v3/view/shared.hpp
FloopCZ: ok, done

include/range/v3/view/shared.hpp
FloopCZ: ok, done

include/range/v3/view/shared.hpp
FloopCZ: ok, done. I thought that it could come in handy when I already have a shared_ptr to a range in hand. However, if the range is efficiently move constructible, it should not affect the performance if it is moved instead of passing the shared_ptr.

include/range/v3/view/shared.hpp
FloopCZ: It was actually @CaseyCarter, who suggested that. :-)

include/range/v3/view/shared.hpp
FloopCZ: true, done

FloopCZ:
> I understand why you would want this to work, but it is guaranteed to do an allocation (make_shared) at each step. This would perform far better (if the range isn't infinite):
> ```
> std::vector<std::vector<int>> tmp = view::indices(42) | view::transform(foo);
> auto rng = tmp | view::join;
> ```

True. Originally, my example was not motivated by infinite ranges, but by very large data (sequences of images) which would not fit into memory all at once. But if it not the case, your solution is of course, better.

> Like Example 1, this is doing lots of unnecessary allocations. view::shared it making it too easy to come up with the wrong solution. This is better:
> 
> ```
> std::vector<int> base{5, 1, 8};
> std::mt19937 gen;
> auto rng =
> 	view::generate([=,&gen]() mutable -> decltype(auto) {
> 		base |= action::shuffle(gen);
> 		return std::as_const(base);
> 	})
>   | view::join;
> ```
> Yes, this view "owns" a vector, but the range is infinite, and copying it is not O(N) so it's fine.

This one has not occurred to me. Thanks. Although how is it that copying of such a view is not O(`size(base)`)? As I understand it, the vector stored inside the lambda function gets copied every time the whole view is copied, doesn't it?

> I don't really understand this example.

Sorry for not making it clear. Sometimes, it is just convenient to have a view that owns the base range and when all the copies of this view get destroyed, the base range gets destroyed too. Why? Consider e.g., [Python NumPy's advanced indexing](https://docs.scipy.org/doc/numpy-1.12.0/reference/arrays.indexing.html#integer-array-indexing). In the world of range-v3 views, it would be something like `view::slice`, where the slice can be an arbitrary range of indices. How to implement this? An obvious option would be to keep a reference to the range of indices inside the `view::slice` and make sure that the lifetime of this range of indices is at least as long as the lifetime of the view.

Why is this options often inconvenient? Let's assume, that I have a library function that returns a view, which selects some important elements from a vector.
```
auto important_view()
{
    return view::...
}

// usage:
std::vector<int> my_data = ...;
auto important_data = my_data | important_view();
```

As long as `important_view()` simply `filters` the data or `takes` the first 5 of them etc., this interface would work just fine. But what if I want to, for instance, ask a user about the important elements? Or just take elements 1, 50, 1930 and 7? I could use the "advanced slice" like this:
```
auto important_view()
{
    std::vector<int> some_indices = ask_user();
    view::slice(some_indices);
}
```

Naturally, this would not work, because `some_indices` would get destroyed at the end of the function. What would probably work is to store `some_indices` to a smart pointer and return it along with the view.
```
auto important_view()
{
    auto some_indices_ptr = make_shared<std::vector<int>>(ask_user());
    return make_tuple(view::slice(*some_indices_ptr), some_indices_ptr);
}
```

But with this, I would have to change the API of the function and the user would be bothered with the lifetime of the `some_indices_ptr`. However, I would like to be able to change the internal implementation of my library function without breaking the public API in such a terrible way. With view::shared, this could be done such as this:
```
auto important_view()
{
    auto some_indices = ask_user() | view::shared;
    return view::slice(some_indices);
}
```

If there is a better way to do this, please educate me. :-)

FloopCZ:
> Would you be opposed to putting this in an experimental namespace?

Well, I would not be opposed, but I would prefer for it to be directly in the `ranges::view`, naturally. The reason is that the GitHub title of `range-v3` says: 

> Experimental range library for C++11/14/17

and being experimental in an experimental library is kind of too much experimental. :-)

Do you prefer `ranges::experimental::view` or `ranges::view::experimental`?

ericniebler:
Only addressing this part for lack of time:

> how is it that copying of such a view is not O(size(base))? As I understand it, the vector stored inside the lambda function gets copied every time the whole view is copied, doesn't it?

It does. But with respect to the size of the `generate` view, it is still O(1) since the view is infinite. Makes sense?

ericniebler:
> The reason is that the GitHub title of range-v3 says:
> 
> > Experimental range library for C++11/14/17
>
> and being experimental in an experimental library is kind of too much experimental. :-)

LOL. I can probably take "experimental" off the description of range-v3. The experiment has largely been successful.

> Do you prefer `ranges::experimental::view` or `ranges::view::experimental`?

Probably `ranges::experimental::view`.

FloopCZ:
> Only addressing this part for lack of time:

Sure. If you ever find some more time, I would be really interested in your solution for the last example. That is something that generally bothers me when I return to C/C++ after writing some code in a garbage collected language.

> It does. But with respect to the size of the generate view, it is still O(1) since the view is infinite. Makes sense?

However, since an algorithm has to terminate (from definition) in a finite time, it can never iterate through an infinite number of elements. Therefore, we should not analyze algorithmic complexity with respect to the size of an infinite view. Otherwise we could consider any finite operation to be O(1), which is not interesting. Well actually, we should not analyze complexity with respect to the size of any view, since views are just algorithms and their size does not have to correspond to their actual running time (especially for infinite views).

Practically, in my original motivation I needed to shuffle thousands or millions of large elements (sequences of images), and iterate through this shuffled range a hundred times. In such a case, the price for copy would be just too high.

> Probably `ranges::experimental::view`.

`ranges::experimental::view` it is. However, in such a case, one has to write a full namespace specification for everything in `ranges::experimental`. Because if one would do:
```
using namespace ranges;
using namespace ranges::experimental;
view::iota(0)  // this will probably not work, view namespace is ambiguous
```
then the `view::` namespace would be ambiguous. Are you still sure you don't prefer `ranges::view::experimental`?

FloopCZ:
- `view::shared` has been moved to the `ranges::experimental::view` namespace (and folder `include/experimental/view`)
- this PR has been updated by merging the most recent master branch

include/range/v3/experimental.hpp
```diff
@@ -0,0 +1,19 @@
...
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_EXPERIMENTAL_HPP
+#define RANGES_V3_EXPERIMENTAL_HPP
+
+#include <range/v3/experimental/view/shared.hpp>
+
+#endif
```
ericniebler: Please delete this header.

include/range/v3/experimental/view/shared.hpp
```diff
@@ -0,0 +1,133 @@
...
+//  Copyright Filip Matzner 2017
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+#ifndef RANGES_V3_VIEW_SHARED_HPP
```
ericniebler: Nit: `RANGES_V3_EXPERIMENTAL_VIEW_SHARED_HPP`.

include/range/v3/experimental/view/shared.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/size.hpp>
+#include <range/v3/view/all.hpp>
+
+namespace ranges
+{
+namespace experimental
+{
+    inline namespace v3
+    {
```
ericniebler: The namespace should be `ranges::v3::experimental`, not `ranges::experimental::v3`. See ranges/v3/experimental/utility/generator.hpp (which should probably also be moved into the experimental/view/ directory now that it's a view).

include/range/v3/experimental/view/shared.hpp
```diff
@@ -0,0 +1,133 @@
...
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(Range<Rng>(),
+                        "The object on which view::shared operates must be a "
+                        "model of the Range concept.");
+                    CONCEPT_ASSERT_MSG(!View<Rng>(),
+                        "view::shared cannot be constructed from a view."
+                        " Please copy the original view instead.");
+                    CONCEPT_ASSERT_MSG(!std::is_reference<Rng>::value,
+                        "view::shared needs an rvalue reference"
```
ericniebler: Needs a space after "rvalue reference" and before the closing quote.

include/range/v3/experimental/view/shared.hpp
```diff
@@ -0,0 +1,133 @@
...
+                        "to build a shared object.");
+                }
+#endif
+            };
+
+            /// \relates all_fn
+            /// \ingroup group-views
+            RANGES_INLINE_VARIABLE(shared_fn, shared)
```
ericniebler: I think this should be `RANGES_INLINE_VARIABLE(view<shared_fn>, shared)`.

FloopCZ: Are you sure about it? I seem to be unable to compile in this form:
```
.../range-v3/include/range/v3/detail/config.hpp:403:63: error: template argument 1 is invalid
             constexpr auto &name = ::ranges::static_const<type>::value; \
                                                               ^
.../range-v3/include/range/v3/experimental/view/shared.hpp:121:17: note: in expansion of macro ‘RANGES_INLINE_VARIABLE’
                 RANGES_INLINE_VARIABLE(view<shared_fn>, shared)
```

FloopCZ:
@ericniebler Oh, I remember why I cannot use `RANGES_INLINE_VARIABLE(view<shared_fn>, shared)` but only `RANGES_INLINE_VARIABLE(shared_fn, shared)`!

The reason is that the `view<>` structure [does not allow](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/view.hpp#L109) piping from rvalue containers, but `view::shared` needs just that. It is the same case as for `view::all`, which also [does not use](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/all.hpp#L103) `view<all_fn>`.

All the other comments were addressed, thank you.

Issue #452, opened 2016-08-19
view::single shouldn't require SemiRegular


ericniebler:
See #445

PR #559, opened 2017-01-29
Release the constraints on view::single

FloopCZ:
- Use CopyConstructible instead of SemiRegular
- Make the view SemiRegular using "detail::optional"
- closes ericniebler/range-v3#452
- almost closes ericniebler/range-v3#441

### DIFF
--- a/include/range/v3/view/single.hpp
+++ b/include/range/v3/view/single.hpp
@@ -16,6 +16,7 @@
 
 #include <utility>
 #include <type_traits>
+#include <range/v3/detail/optional.hpp>
 #include <range/v3/detail/satisfy_boost_range.hpp>
 #include <range/v3/range_fwd.hpp>
 #include <range/v3/begin_end.hpp>
@@ -38,20 +39,20 @@ namespace ranges
         {
         private:
             friend struct range_access;
-            Val value_;
+            optional<Val> value_opt_;
             struct cursor
             {
             private:
-                Val value_;
+                optional<Val> value_opt_;
                 bool done_;
             public:
                 cursor() = default;
-                explicit cursor(Val value)
-                  : value_(std::move(value)), done_(false)
+                explicit cursor(optional<Val> value_opt)
+                  : value_opt_{std::move(value_opt)}, done_{false}
                 {}
                 Val read() const
                 {
-                    return value_;
+                    return *value_opt_;
                 }
                 bool equal(default_sentinel) const
                 {
@@ -82,12 +83,12 @@ namespace ranges
             };
             cursor begin_cursor() const
             {
-                return cursor{value_};
+                return cursor{value_opt_};
             }
         public:
             single_view() = default;
             constexpr explicit single_view(Val value)
-              : value_(detail::move(value))
+              : value_opt_{std::move(value)}
             {}
             constexpr std::size_t size() const
             {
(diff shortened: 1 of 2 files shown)


include/range/v3/view/single.hpp
```diff
@@ -38,20 +39,20 @@ namespace ranges
         {
         private:
             friend struct range_access;
-            Val value_;
+            optional<Val> value_opt_;
             struct cursor
             {
             private:
-                Val value_;
+                optional<Val> value_opt_;
```
ericniebler: This should use the `semiregular_t` helper instead of using `optional` directly. See how the other views do it; for example, `istream_view`.

include/range/v3/view/single.hpp
```diff
@@ -82,12 +83,12 @@ namespace ranges
...
             cursor begin_cursor() const
             {
-                return cursor{value_};
+                return cursor{value_opt_};
             }
         public:
             single_view() = default;
             constexpr explicit single_view(Val value)
-              : value_(detail::move(value))
+              : value_opt_{std::move(value)}
```
ericniebler: This should stay `detail::move`. `std::move` is not `constexpr` in C++11.

FloopCZ:
Then I hope you enjoyed the honeymoon! And welcome back. :-)

Thank you very much for the review, the changes you requested have been implemented.
However, the issue #441 is still not completely solved.

include/range/v3/view/single.hpp
```diff
@@ -38,16 +40,16 @@ namespace ranges
...
             private:
-                Val value_;
+                semiregular_t<Val> value_;
                 bool done_;
             public:
                 cursor() = default;
-                explicit cursor(Val value)
-                  : value_(std::move(value)), done_(false)
+                explicit cursor(semiregular_t<Val> value_opt)
+                  : value_{detail::move(value_opt)}, done_{false}
```
ericniebler: By the time you create a `cursor`, you better have an actual value! IMO, these changes to `cursor` should be reverted.

include/range/v3/view/single.hpp
```diff
@@ -99,22 +101,21 @@ namespace ranges
         {
             struct single_fn
             {
-                template<typename Val, CONCEPT_REQUIRES_(SemiRegular<Val>())>
+                template<typename Val, CONCEPT_REQUIRES_(CopyConstructible<Val>())>
                 single_view<Val> operator()(Val value) const
                 {
-                    return single_view<Val>{std::move(value)};
+                    return single_view<Val>{detail::move(value)};
```
ericniebler: `detail::move` only matters in `constexpr` functions.

include/range/v3/view/single.hpp
FloopCZ: You are very right! Sorry for that.

include/range/v3/view/single.hpp
FloopCZ: Should I revert it too?

FloopCZ:
Thanks for another review. Reverted whatever was not neccessary. Now the changes relative to the upstream should be minimal.

include/range/v3/view/single.hpp
```diff
@@ -38,11 +39,11 @@ namespace ranges
         {
         private:
             friend struct range_access;
-            Val value_;
+            semiregular_t<Val> value_;
             struct cursor
             {
             private:
-                Val value_;
+                semiregular_t<Val> value_;
```
ericniebler: I think you missed this one. This should be `Val` instead of `semiregular_t<Val>`.

CaseyCarter: `cursor` needs `value_` to have a `Semiregular` type in order to be `Semiregular` itself.

ericniebler: Ah, right. Never mind, then.

FloopCZ: That was my thinking as well. Thank you for the merge!

Issue #556, opened 2017-01-21
segfault when initializing vector-of-vector with range-of-range


h-2:
According to the documentation you can initialize a sequence container with a view over the same value-type. Is this supposed to work for two-dimensional views and containers, too?

This program crashes for me in the last assignment:
```c++
#include <iostream>
#include <range/v3/all.hpp>
#include <string>

int main()
{
    std::string s{"abc"};
    ranges::any_random_access_view<ranges::any_random_access_view<char>> v1 = ranges::view::single(s | ranges::view::drop(1));
    ranges::any_random_access_view<ranges::any_random_access_view<char>> v2 = ranges::view::single(s | ranges::view::drop(2));
    
    auto v3 = ranges::view::concat(v1, v2);

// both of these crash:
//    std::vector<std::vector<char>> owner = v3;
//    std::vector<std::string> owner = v3;

    return 0;
}
```

Thanks!

edit: clarified which calls create a crash

FloopCZ:
Hi h-2, could you provide your compiler version and compiler flags, please (`g++ -v`)?

I have been unable to reproduce the bug with neither g++ 6.3.1 or clang++ 3.9.1 on my Arch. Both versions of your code snippet compile fine, produce no error when run and valgrind reports no leaks.

h-2:
Hm, that's curious. From my home computer I also cannot reproduce this now. Tried with gcc-versions between 4.9 and 7 and clang versions between 3.5 and 4.0.
I will double-check in the office on Monday. Maybe it's also something that has been fixed in the meantime?

FloopCZ:
According to Travis, the problem affects gcc 4.7-4.9
https://travis-ci.org/ericniebler/range-v3/builds/202624667
I have been able to reproduce it with gcc 4.9.2 just moments ago.

In #563, there is an ugly workaround if you really need that functionality right now. :-)

h-2:
@FloopCZ thanks for investigating! I don't have urgent need for a fix, I am just doing smaller tests right now to see if/how things work. And I will be targetting only newer compilers anyway!

PR #563, opened 2017-02-17
Add test for issue #556

FloopCZ:
Let travis check #556 on multiple compiler versions.

### DIFF
--- a/include/range/v3/begin_end.hpp
+++ b/include/range/v3/begin_end.hpp
@@ -87,8 +87,10 @@ namespace ranges
             private:
                 template<typename Rng>
                 static constexpr auto impl(Rng && rng, long)
-                    noexcept(noexcept(begin(static_cast<Rng &&>(rng)))) ->
-                    decltype(begin(static_cast<Rng &&>(rng)))
+#if !defined(__GNUC__) || __GNUC__ >= 5
+                    noexcept(noexcept(begin(static_cast<Rng &&>(rng))))
+#endif
+                    -> decltype(begin(static_cast<Rng &&>(rng)))
                 {
                     return begin(static_cast<Rng &&>(rng));
                 }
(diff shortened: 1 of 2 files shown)


FloopCZ:
The compilation fails on g++ 4 with the following error:
```
/home/travis/build/ericniebler/range-v3/include/range/v3/begin_end.hpp:89:39: error: declaration of ‘static constexpr decltype (ranges::v3::adl_begin_end_detail::begin(static_cast<Rng&&>(rng))) ranges::v3::adl_begin_end_detail::begin_fn::impl(Rng&&, long int) noexcept (<uninstantiated>) [with Rng = ranges::v3::indexed_element<ranges::v3::any_view<char, (ranges::v3::category)3>, 0ul>&; decltype (ranges::v3::adl_begin_end_detail::begin(static_cast<Rng&&>(rng))) = ranges::v3::basic_iterator<ranges::v3::detail::any_cursor<char, (ranges::v3::category)3> >]’ has a different exception specifier
                 static constexpr auto impl(Rng && rng, long)
                                       ^
/home/travis/build/ericniebler/range-v3/include/range/v3/begin_end.hpp:89:39: error: from previous declaration ‘static constexpr decltype (ranges::v3::adl_begin_end_detail::begin(static_cast<Rng&&>(rng))) ranges::v3::adl_begin_end_detail::begin_fn::impl(Rng&&, long int) noexcept (false) [with Rng = ranges::v3::indexed_element<ranges::v3::any_view<char, (ranges::v3::category)3>, 0ul>&; decltype (ranges::v3::adl_begin_end_detail::begin(static_cast<Rng&&>(rng))) = ranges::v3::basic_iterator<ranges::v3::detail::any_cursor<char, (ranges::v3::category)3> >]’
```

include/range/v3/begin_end.hpp
```diff
@@ -103,8 +105,10 @@ namespace ranges
             public:
                 template<typename Rng>
                 constexpr auto operator()(Rng && rng) const
-                    noexcept(noexcept(begin_fn::impl(static_cast<Rng &&>(rng), 0))) ->
-                    detail::decay_t<decltype(begin_fn::impl(static_cast<Rng &&>(rng), 0))>
+#if !defined(__GNUC__) || __GNUC__ >= 5
+                    noexcept(noexcept(begin_fn::impl(static_cast<Rng &&>(rng), 0)))
+#endif
+                    -> detail::decay_t<decltype(begin_fn::impl(static_cast<Rng &&>(rng), 0))>
```
ericniebler: This is not acceptable. We need to find another workaround.

FloopCZ: I strongly agree. It seems the problem only occurs if the range is any_view, but I have not been able to provide a better solution.

It was more for the Travis to check whether it passes the tests since I did not have a machine with gcc4. It would actually pass the tests I think, since Travis failed with unrelated apt-get error on two clang tests before even trying to build.

FloopCZ:
I was unable to fix this issue for gcc 4. Those may be related if anyone else wants give it a try:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56643
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59526

(On the other hand, the issue only concerns multi-dimensional any_view and a fairly old gcc 4, so it is questionable if it is worth fixing. I will close this PR.)

PR #568, opened 2017-02-28
drop_exactly_view should work with mutable-only random-access base views

CaseyCarter:


include/range/v3/view/drop_exactly.hpp
```diff
@@ -51,10 +51,18 @@ namespace ranges
             difference_type_ n_;
 
             // RandomAccessRange == true
+            template<typename BaseRng = Rng,
+                CONCEPT_REQUIRES_(RandomAccessRange<BaseRng const>())>
```
ericniebler: `RandomAccessRange<BaseRng const &>()`

include/range/v3/view/drop_exactly.hpp
```diff
@@ -51,10 +51,18 @@ namespace ranges
...
             range_iterator_t<Rng> get_begin_(std::true_type) const
             {
                 return next(ranges::begin(rng_), n_);
             }
+            template<typename BaseRng = Rng,
+                CONCEPT_REQUIRES_(!RandomAccessRange<BaseRng const>())>
```
ericniebler: `!RandomAccessRange<BaseRng const &>()`

include/range/v3/view/drop_exactly.hpp
CaseyCarter: The `Range` concept family is oblivious to reference qualifiers.

include/range/v3/view/drop_exactly.hpp
CaseyCarter: Ditto.

include/range/v3/view/drop_exactly.hpp
CaseyCarter: ....since `iterator_t<T>` is `decltype(begin(declval<T&>()))` and `sentinel_t<T>` is `decltype(end(declval<T&>()))`.

CaseyCarter:
The thing that worries me most here is the expectation that `range_iterator_t<T>` and `range_iterator_t<T const>` are always the same type when `View<T>() && Range<T const>()` is satisfied. That seems like a sensible requirement to me, but I don't think we require it anywhere - despite this and perhaps other views relying on that assumption.

PR #570, opened 2017-03-09
Bring range-v3 closer in line with the current draft wording of the Ranges TS

ericniebler:


include/range/v3/utility/counted_iterator.hpp
```diff
@@ -25,183 +25,344 @@ namespace ranges
...
+            {
+                return y.count();
+            }
+
+            template<typename I,
+                CONCEPT_REQUIRES_(RandomAccessIterator<I>())>
+            counted_iterator<I>
+            operator+(difference_type_t<I> n, const counted_iterator<I>& x)
+            {
+                RANGES_ASSERT(n <= x.count());
```
CaseyCarter: Why check the contract redundantly here?

include/range/v3/utility/counted_iterator.hpp
```diff
@@ -25,183 +25,344 @@ namespace ranges
...
+                using reference = void;
+                using pointer = void;
+            };
+
+            template<typename I>
+            struct iterator_traits_<I, meta::if_<InputIterator<I>>>
+            {
+                using iterator_category =
+                    meta::if_c<
+                        (bool) ForwardIterator<I>() &&
```
CaseyCarter: Unnecessary `(bool)` cast.

include/range/v3/utility/iterator.hpp
```diff
@@ -726,80 +742,150 @@ namespace ranges
...
-        /// \cond
-        namespace detail
+        template<typename I>
+        struct move_iterator
         {
-            template<typename I>
-            struct move_cursor
+        private:
+            CONCEPT_ASSERT(InputIterator<I>());
+            I current_ = detail::value_init{};
```
CaseyCarter: `InputIterator` does not require convertibility from `value_init_t`; this could explode.

include/range/v3/view/generate.hpp
```diff
@@ -39,7 +39,7 @@ namespace ranges
           : view_facade<generate_view<G>, infinite>
         {
         private:
-            friend struct range_access;
+            friend struct ranges::range_access;
```
CaseyCarter: Yowch. The compiler would have caught this bug if these were simply `friend ranges::range_access;` Without the class-key, the friend declaration is not a forward declaration and is ill-formed if the named type is not already declared.

Version.cmake
```diff
@@ -1,5 +1,5 @@
 # To update the range-v3 version, from a *CLEAN* working directory, update the version numbers below.
 # This makefile will generate a new version.hpp, *AMEND THE MOST RECENT COMMIT*, and git-tag the commit.
 set(RANGE_V3_MAJOR 0)
-set(RANGE_V3_MINOR 1)
+set(RANGE_V3_MINOR 2)
```
ericniebler: In my understanding, as long as the major version says '0', I can break anything at any time with impunity. Mwah-hah-ha!

include/range/v3/utility/swap.hpp
```diff
@@ -29,6 +29,19 @@ namespace ranges
...
     {
+        /// \cond
+        namespace detail
+        {
+            template<typename T>
+            struct is_movable_
+              : meta::and_<
+                    std::is_object<T>,
+                    std::is_constructible<T, T>,
+                    std::is_assignable<T &, T>>
```
ericniebler: &lt;shrug&gt; My way saves two characters. ;-)

include/range/v3/utility/counted_iterator.hpp
```diff
@@ -25,183 +25,344 @@ namespace ranges
...
+                reference_t<I> operator*() const
+                RANGES_AUTO_RETURN_NOEXCEPT
+                (
+                    *static_cast<I2 const &>(current_)
+                )
+
+                counted_iterator& operator++()
                 {
-                    *it_ = (T &&) t;
+                    RANGES_ENSURE(cnt_ > 0);
```
ericniebler: You would prefer an assert here instead?

include/range/v3/utility/swap.hpp
CaseyCarter: They *are* equivalent, but e.g. `is_move_constructible<T>` is a bit higher level whereas `is_constructible<T, T>` requires analysis on the reader's part to figure out what it's doing.

But yeah, not a big deal either way.

include/range/v3/utility/counted_iterator.hpp
CaseyCarter: It's side-effect free and zero-cost, I'd use `RANGES_EXPECT` here and try to help the optimizer. `RANGES_ASSERT` is now really only used for side-effect-free but potentially expensive checks.

include/range/v3/utility/iterator.hpp
ericniebler: `detail::value_init_t` is implicitly convertible to any default constructible type. I don't see how this could explode.

ericniebler:
Rebased. I also took care of the `is_swappable` -> `is_swappable_with` change in fbe1ec0.

PR #582, opened 2017-03-15
Add test for issue #556

### LINKED ISSUE #556: segfault when initializing vector-of-vector with range-of-range

FloopCZ:
Since range-v3 dropped support for g++-4.9.2, issue #556 is probably not an issue anymore. If tests pass, this closes #556.

### DIFF
--- a/test/container_conversion.cpp
+++ b/test/container_conversion.cpp
@@ -13,9 +13,14 @@
 #include <set>
 #include <list>
 #include <sstream>
+#include <string>
 #include <vector>
 #include <range/v3/core.hpp>
+#include <range/v3/view/any_view.hpp>
+#include <range/v3/view/concat.hpp>
+#include <range/v3/view/drop.hpp>
 #include <range/v3/view/take.hpp>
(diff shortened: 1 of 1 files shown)


gnzlbg:
It seems that tests do pass, the failure is spurious. We should close #556 .

PR #584, opened 2017-03-15
Add ranges::index for unchecked indexed range access.

### LINKED ISSUE #575: ranges::at improvements
I think we could improve ranges::at for `SizedRanges` by adding a `RANGES_EXPECT(n < ranges::size(rng));` that asserts in debug and assumes in release builds.

### LINKED ISSUE #585: ranges::at vocabulary
All std:: containers use `at` for checked access, but range-v3 uses `at` for unchecked access.

I like it this way but I would be surprised if no-one would complain about this when trying to standardize it.

gnzlbg:
This PR:

- Makes `ranges::at` perform checked indexed range access (like the `.at` methods in the standard containers). 
- Adds an overload to `ranges::at` for non-sized ranges that performs unchecked range access (since their length cannot be computed in O(1)). This overload is marked as deprecated and is provided for backwards compatibility only (should be removed in the future).
- Adds a new core function: `ranges::index`, that performs unchecked indexed range access. For sized ranges, this functions asserts in debug mode that the range index is in bounds, and assumes this in release mode. 
- Makes ranges::distance conditionally C++14 constexpr for SizedRanges. This is required for `ranges::index` to be constexpr due to the assertion.

Closes #575 .
Closes #585 .


include/range/v3/at.hpp
```diff
@@ -30,7 +31,20 @@ namespace ranges
...
+                decltype(begin(rng)[n])
+            {
+                RANGES_EXPECT(static_cast<std::size_t>(n) < ranges::size(rng));
+                return begin(rng)[n];
+            }
+
+            /// \return `begin(rng)[n]`
+            template<typename Rng,
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>() and
+                                  not SizedRange<Rng>())>
```
ericniebler: Ick. No iso646 alternate keywords here, please.

Also, this can be collapsed into one overload with:

```c++
if (SizedRange<Rng>())
  RANGES_EXPECT(static_cast<std::size_t>(n) < ranges::distance(rng));
...
```

CaseyCarter: `RANGES_EXPECT(static_cast<std::size_t>(n) < ranges::distance(rng));`

`ranges::distance` returns `difference_type_t`, there's no need for the cast.

CaseyCarter:
> I'm somewhat conflicted about `ranges::at`. `vector::at` will make people think that `ranges::at` is checked and throws on out of range, and this doesn't. I'm tempted to rename this to `ranges::index`, and add a `ranges::at` that requires `SizedRange` and throws on out of range. Thoughts?

[`index`](https://www.freebsd.org/cgi/man.cgi?query=index&sektion=3) could be challenging to standardize thanks to the existence of the legacy BSD string function of the same name. POSIX dropped the function in 2008, so we may be able to get away with it.

Also we should be discussing this in #585.

include/range/v3/at.hpp
gnzlbg: @ericniebler I've fixed it.

@CaseyCarter I get an error if I remove the cast:

```shell
 error: comparison of integers of different signs: 'range_difference_type_t<int (&)[4]>'
      (aka 'long') and 'decltype(fn::impl_(r, 42))' (aka 'unsigned long')
```

CaseyCarter: > @CaseyCarter I get an error if I remove the cast

That's because you didn't replace the call to `ranges::size` with a call to `ranges::distance`.

CaseyCarter: > That's because you didn't replace the call to `ranges::size` with a call to `ranges::distance`.

Which is necessary for correctness: `ranges::distance` is valid for all ranges, `ranges::size` is not. Since that `if` isn't `if constexpr`, the code needs to use `distance` instead of `size`.

gnzlbg: Indeed, sorry about that.

include/range/v3/at.hpp
```diff
@@ -35,10 +36,15 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
+                if (SizedRange<Rng>()) {
+                    RANGES_EXPECT(n < ranges::distance(rng));
```
CaseyCarter: We've made you change this like ten times already, but I observe this whole conditional could be:
```c++
RANGES_EXPECT(!SizedRange<Rng>() || n < ranges::distance(rng));
```
I'll let you decide if you want to change it again ;)

gnzlbg: I will do so tomorrow :D 
Keep them coming ;)

gnzlbg: Fixed. I'll love this OCD of yours. I wonder if I might be able to have it someday as well.

include/range/v3/at.hpp
```diff
@@ -18,6 +18,7 @@
 #include <range/v3/begin_end.hpp>
 #include <range/v3/range_traits.hpp>
 #include <range/v3/range_concepts.hpp>
+#include <range/v3/size.hpp>
```
CaseyCarter: In the spirit of churning this PR forever, I'll point out that we could remove size.hpp and utility/iterator.hpp and instead include distance.hpp.

gnzlbg: fixed :)

gnzlbg:
So I've updated the PR with the new changes.

gnzlbg:
There is a new change here that was not previously discussed. I've made `ranges::distance` conditionally C++14 constexpr for `SizedRange`s so that `ranges::index` can be constexpr (due to the assertion), since I thought that we want to make distance constexpr at some point anyways.

We can probably make `ranges::distance` C++11 constexpr as well but I've preferred to keep this as simple as possible for now. I want to submit a PR that makes one algorithm constexpr, `ranges::equal`, rather soonish anyways, with the aim of starting somewhere with the algorithms and make them constexpr one algorithm per PR at a time. I'll send a PR before that with some changes to the core utilities so that this will work, so I could make `ranges::distance` C++11 constexpr then if desired (I prefer not to, C++11 constexpr makes my eyes bleed).

gnzlbg:
> index could be challenging to standardize thanks to the existence of the legacy BSD string function of the same name. POSIX dropped the function in 2008, so we may be able to get away with it.

The test for `ranges::index` cannot use `using namespace ranges;` because clang modules pulls the precompiled POSIX headers into scope even when the headers are not included due to bug: https://bugs.llvm.org//show_bug.cgi?id=32192 @zygoloid ;)

gnzlbg:
I've added a test that's commented out: 

```c++
       auto rng = ranges::view::ints(std::uint64_t{0}, std::numeric_limits<std::uint64_t>::max() - 1);
       CHECK(ranges::index(rng, std::numeric_limits<std::uint64_t>::max()) ==
             std::numeric_limits<std::uint64_t>::max() - 2);
       CHECK(ranges::at(rng, std::numeric_limits<std::uint64_t>::max()) ==
             std::numeric_limits<std::uint64_t>::max() - 2);
```

it is at least blocked by #593 but I am not sure if we can make this to work at all due to an implicit convertion to a signed type on `begin()[n]`.

include/range/v3/at.hpp
```diff
@@ -14,35 +15,64 @@
...
             template<typename Rng,
-                CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>() && SizedRange<Rng>())>
             RANGES_CXX14_CONSTEXPR
-            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
+            auto operator()(Rng &&rng, range_size_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
+                const auto dist = ranges::distance(rng);
+                RANGES_EXPECT(dist >= 0);
```
CaseyCarter: This is a post-condition of `ranges::distance(rng)`; ideally it should be `RANGES_EXPECT`ed there.

include/range/v3/at.hpp
```diff
@@ -14,35 +15,64 @@
...
 #include <range/v3/begin_end.hpp>
 #include <range/v3/range_traits.hpp>
 #include <range/v3/range_concepts.hpp>
-#include <range/v3/utility/iterator.hpp>
+#include <range/v3/distance.hpp>
 #include <range/v3/utility/static_const.hpp>
+#include <range/v3/index.hpp>
+
+RANGES_DISABLE_WARNINGS
+RANGES_DIAGNOSTIC_IGNORE_SIGN_CONVERSION
```
CaseyCarter: These are big hammers, and they don't seem to be silencing any warnings in here; the only conversion I see is a `static_cast`.

include/range/v3/at.hpp
```diff
@@ -14,35 +15,64 @@
...
         /// \ingroup group-core
         struct at_fn
         {
             /// \return `begin(rng)[n]`
             template<typename Rng,
-                CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>() && SizedRange<Rng>())>
             RANGES_CXX14_CONSTEXPR
-            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
+            auto operator()(Rng &&rng, range_size_type_t<Rng> n) const ->
```
CaseyCarter: I would prefer both `at` and `index` to accept difference type and require (`RANGES_EXPECT`) that  `n >= 0`. IMO `-21` is not "outside the range of valid indices," it's a bug: negative numbers are *never* valid indices.

include/range/v3/index.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/begin_end.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/distance.hpp>
+#include <range/v3/utility/static_const.hpp>
+
+RANGES_DISABLE_WARNINGS
+RANGES_DIAGNOSTIC_IGNORE_SIGN_CONVERSION
```
CaseyCarter: Ditto un-targeted warning hammer.

include/range/v3/index.hpp
```diff
@@ -0,0 +1,61 @@
...
+        {
+            /// \return `begin(rng)[n]`
+            template<typename Rng,
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
+            RANGES_CXX14_CONSTEXPR
+            auto operator()(Rng &&rng, range_size_type_t<Rng> n) const ->
+                decltype(begin(rng)[n])
+            {
+                RANGES_EXPECT(!SizedRange<Rng>() ||
+                    (ranges::distance(rng) >= 0 &&
```
CaseyCarter: Ditto postcondition of `ranges::distance`.

include/range/v3/at.hpp
gnzlbg: They are silencing the conversion in here: 

```c++
    begin(rng)[n];
```

The solution to this seems obvious now:

```c++
    begin(rng)[static_cast<range_difference_t<Rng>>(n)];
```

include/range/v3/at.hpp
gnzlbg: @ericniebler thoughts? 

I would prefer to use `difference_type` here as well, I went for `range_size_type` because that's what all containers with `at` and `operator[]` in the STL actually use.

If we use `difference_type` people calling `ranges::at(vec, vec.size() - 1)` will get warnings about `vec.size() - 1` being implicitly converted to a  signed integer type that might truncate its value and that is going to be very annoying.

Maybe we should offer two overloads? One that takes a size type and one that takes a difference type?

ericniebler: > If we use `difference_type` people calling `ranges::at(vec, vec.size() - 1)` will get warnings

Yup. In this case, they should just be using `back`, but the point stands. The question is how annoying and prescriptive we want to be in range-v3, given that the code range-v3 must play with does things a different way. I'm still undecided.

gnzlbg: We could also complement the "two-overload" proposal with a `RANGES_V3_STL1_COMPATIBILITY` macro that allows users turning the STL1 compatibility on and off. I don't know whether I like this idea though.

CaseyCarter: Two overloads for distinct integral types will result in nasty ambiguities for other arithmetic types. What about:
```c++
template<typename Rng,
    CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const
(
    (RANGES_EXPECT(!SizedRange<Rng>() || n < ranges::distance(rng)),
     ranges::begin(rng)[n])
)
template<typename Rng, typename T, typename Fn = at_fn,
    CONCEPT_REQUIRES_(RandomAccessRange<Rng>() &&
        ConvertibleTo<T, range_difference_type_t<Rng>>())>
auto operator()(Rng &&rng, T &&t) const
(
    Fn{}((Rng &&) rng, static_cast<range_difference_type_t<Rng>>((T &&) t))
)
```

CaseyCarter: > If we use `difference_type` people calling `ranges::at(vec, vec.size() - 1)` will get warnings about `vec.size() - 1` being implicitly converted to a signed integer type that might truncate its value and that is going to be very annoying.

I think a silently bi-convertible wrapper would help here, a la:
```c++
template<typename I>
struct wrapper
{
    CONCEPT_ASSERT(Integral<T>());
    T value_;

    constexpr operator meta::_t<std::make_signed<T>> () const noexcept
    {
        auto const result = static_cast<meta::_t<std::make_signed<T>>>(value_);
        RANGES_EXPECT(0 <= result);
        return result;
    }
    constexpr operator meta::_t<std::make_unsigned<T>> () const noexcept
    {
        return static_cast<meta::_t<std::make_unsigned<T>>>(value_);
    }
};
```

include/range/v3/at.hpp
gnzlbg: fixed

include/range/v3/index.hpp
gnzlbg: fixed

include/range/v3/index.hpp
gnzlbg: fixed

include/range/v3/at.hpp
gnzlbg: I've fixed it using your first solution.

CaseyCarter: > I've fixed it using your first solution.

I completely failed to express that the implicit conversion wrapper was for use in the return expression a la `begin()[wrapper<range_difference_type_t<Rng>>{n}]` to avoid conversion warnings if someone defines `operator[]` for their iterator as taking an unsigned type. That said, I can't imagine *why* someone would do so or why we should encourage such a thing. Consider that suggestion withdrawn.

include/range/v3/at.hpp
```diff
@@ -14,31 +15,83 @@
...
         struct at_fn
         {
             /// \return `begin(rng)[n]`
             template<typename Rng,
-                CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>() && SizedRange<Rng>())>
             RANGES_CXX14_CONSTEXPR
-            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
+            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const 
+                noexcept(noexcept(begin(rng)[n]) && noexcept(ranges::distance(rng)))
```
CaseyCarter: Umm, `noexcept`? On a function designed to throw exceptions?

include/range/v3/at.hpp
```diff
@@ -14,31 +15,83 @@
...
             }
+            /// \return `begin(rng)[n]`
+            template<typename Rng,
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>() && !SizedRange<Rng>())>
+            RANGES_DEPRECATED(
+                "Checked indexed range access (ranges::at) on !SizedRanges is deprecated! "
+                "This version performs unchecked access (the range size cannot be computed in O(1) for !SizedRanges)! "
+                "Use ranges::index for unchecked access instead!")
+            RANGES_CXX14_CONSTEXPR
+            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const
```
CaseyCarter:
```c++
auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const
(
    ranges::index(std::forward<Rng>(rng), n)
)
```

include/range/v3/index.hpp
```diff
@@ -0,0 +1,95 @@
...
+        /// Unchecked indexed range access.
+        ///
+        /// \ingroup group-core
+        struct index_fn
+        {
+            /// \return `begin(rng)[n]`
+            template<typename Rng,
+                CONCEPT_REQUIRES_(RandomAccessRange<Rng>())>
+            RANGES_CXX14_CONSTEXPR
+            auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const
```
CaseyCarter: The conditional noexcept clause is missing a contribution from `distance`:
```c++
(
    (RANGES_EXPECT(!SizedRange<Rng>() || n < ranges::distance(rng)),
        ranges::begin(rng)[n])
)
```

test/index.cpp
```diff
@@ -0,0 +1,65 @@
...
+        bool did_throw = false;
+        try {
+            ranges::at(vi, 4);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            did_throw = true;
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("ranges::at")));
+        }
+        CHECK(did_throw);
```
CaseyCarter: This is redundant with the `CHECK(false)`.

gnzlbg:
@CaseyCarter it seems that GCC 4.9 and GCC 5.0 do not like the `RANGES_EXPECT(d >= d)` in `ranges::distance`:

```shell
/home/travis/build/gnzlbg/range-v3/include/range/v3/distance.hpp: In member function ‘D ranges::v3::distance_fn::operator()(Rng&&, D) const [with Rng = const S (&)[10]; D = long int; int _concept_requires_87 = 42; typename std::enable_if<((_concept_requires_87 == 43) || (ranges::v3::concepts::models<ranges::v3::concepts::Integral, D>() && ranges::v3::concepts::models<ranges::v3::concepts::Range, T>())), int>::type <anonymous> = 0]’:
/home/travis/build/gnzlbg/range-v3/include/range/v3/detail/config.hpp:55:29: error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]
 #define RANGES_ASSUME(COND) static_cast<void>((COND) ? void(0) : __builtin_unreachable())
                             ^
/home/travis/build/gnzlbg/range-v3/include/range/v3/detail/config.hpp:65:29: note: in expansion of macro ‘RANGES_ASSUME’
 #define RANGES_EXPECT(COND) RANGES_ASSUME(COND)
                             ^
/home/travis/build/gnzlbg/range-v3/include/range/v3/distance.hpp:94:17: note: in expansion of macro ‘RANGES_EXPECT’
                 RANGES_EXPECT(result >= d);
                 ^
/home/travis/build/gnzlbg/range-v3/include/range/v3/distance.hpp: In member function ‘D ranges::v3::distance_fn::operator()(Rng&&, D) const [with Rng = const T (&)[10]; D = long int; int _concept_requires_87 = 42; typename std::enable_if<((_concept_requires_87 == 43) || (ranges::v3::concepts::models<ranges::v3::concepts::Integral, D>() && ranges::v3::concepts::models<ranges::v3::concepts::Range, T>())), int>::type <anonymous> = 0]’:
/home/travis/build/gnzlbg/range-v3/include/range/v3/detail/config.hpp:55:29: error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]
 #define RANGES_ASSUME(COND) static_cast<void>((COND) ? void(0) : __builtin_unreachable())
                             ^
/home/travis/build/gnzlbg/range-v3/include/range/v3/detail/config.hpp:65:29: note: in expansion of macro ‘RANGES_ASSUME’
 #define RANGES_EXPECT(COND) RANGES_ASSUME(COND)
                             ^
/home/travis/build/gnzlbg/range-v3/include/range/v3/distance.hpp:94:17: note: in expansion of macro ‘RANGES_EXPECT’
                 RANGES_EXPECT(result >= d);
                 ^
cc1plus: all warnings being treated as errors
```

GCC 6 works just fine. See the builds here: https://travis-ci.org/gnzlbg/range-v3/builds/213998988

CaseyCarter:
> @CaseyCarter it seems that GCC 4.9 and GCC 5.0 do not like the `RANGES_EXPECT(result >= d)` in `ranges::distance` (the warning seems to be a false positive)

Best fix seems to be killing the overzealous warning for GCC <= 5 (https://github.com/gnzlbg/range-v3/pull/3).

ericniebler:
This error looks legit: https://travis-ci.org/ericniebler/range-v3/jobs/214390032

CaseyCarter:
> This error looks legit: https://travis-ci.org/ericniebler/range-v3/jobs/214390032

Curses, foiled again. I re-enabled C++14 constexpr for GCC5 builds in my last adjustment of the test matrix under the belief that 5.3 or 5.4 had fixed the constexpr array bugs. It would appear that there are still others for which we didn't have test coverage.

I strongly suspect that trying to workaround this bug would be a waste of time, and that we should go back to testing GCC5 with `-DRANGES_CXX_CONSTEXPR=RANGES_CXX_CONSTEXPR11`.

include/range/v3/at.hpp
ericniebler: This will find the `ranges::index` object and not do ADL, right? How is the explicit qualification an improvement?

test/index.cpp
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            ranges::at(vi, 4);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("ranges::at")));
+        }
+    }
+
+    // { // Blocked by #593 and an implicit conversion of `n` to signed in begin()[n]:
```
ericniebler: #593 is merged now. Can you pls. fix this up?

include/range/v3/at.hpp
CaseyCarter: IIRC, this function had separate `decltype`, `noexcept`, and function body when I made this comment. I have a mild preference for explicitly qualifying function objects in the `ranges` namespace: it has no effect on correctness, but it helps readability for those who haven't memorized the entire content of the library.

gnzlbg: @ericniebler are `index` and `at` customization points? AFAIK only `begin` and `end` are, so IIUC this doesn't matter.

include/range/v3/index.hpp
gnzlbg: RANGES_EXPECT failed in some builds, but i've added another `noexcept(...)` clause.

gnzlbg:
travis is all green

include/range/v3/at.hpp
ericniebler: They are not customization points. This does not strictly need to be explicitly qualified.

PR #586, opened 2017-03-15
view::linear_distribute

### LINKED ISSUE #102: ints() double support, step/stride syntax sugar.
Does `ints(0.5, 5.5)` return a range of doubles [0.5, 5.5) that is incremented by 1.0 ?

If so, then the name `ints` is a bit misleading and otherwise, why not? 

Both python `xrange` and D's `iota` support a step or a stride, as well as integers and floating point, so I think that this common case deserves some syntax sugar for creating [from, to) strided ranges.  

A solution to this could be to offer both, an `integer_range` and a `numeric_range`. The numeric range could have the stride bolted in (I don't think `view::strided` can or should be made to work with `numeric` strides). The `numeric_range` could use by default `a < b && a > b` as implementation for `a != b` but this should probably be customizable. 

Then something like:
-  `range(0, 1)` returns an `integer_range`,
-  `range(0, 4, 2)` returns an strided view of an integer range, and
-  `range(0.0, 5.0, 1.0)` returns a `numeric_range`. 

(TODO: replace `range` with a better name).

gnzlbg:
Adds `view::linear_distribute(from, to, n)` which linearly distributes `n`
elements in the closed range `[from, to]`. It models `ForwardRange`.

Its semantics are:

- Precondition: `from <= to and n > 0`.
- Requirements: `std::is_arithmetic<T>`.
- Returns:
  - If `n == 1`, it returns `to`.
  - If `from == to`, it returns `n`-times to.
  - Otherwise, distirbutes `n` equally-spaced points in the interval
  `[from, to]` (the end-points are always included).

Possible future extensions:

- Support user-defined integer types: see issues #605 ande #591.
- Model `BidirectionalRange`.
- Model `RandomAccessRange` for `std::is_integral<T>`.

Closes #102.


include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
+/// \file
```
CaseyCarter: This header needs to be added to `range/v3/view.hpp`.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+        private:
+            friend range_access;
+
+            T from_, to_, spacing_;
+
+            constexpr T read() const
+            {
+                return from_;
+            }
+            constexpr void next()
```
CaseyCarter: `RANGES_CXX14_CONSTEXPR`

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+            constexpr bool equal(default_sentinel) const
+            {
+                return from_ == to_ + spacing_;
+            }
+            CONCEPT_REQUIRES(std::is_floating_point<T>())
+            constexpr bool equal(default_sentinel) const
+            {
+                return from_ > to_;
+            }
+            constexpr void prev()
```
CaseyCarter: `RANGES_CXX14_CONSTEXPR`

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+            {
+                from_ -= spacing_;
+            }
+            constexpr void advance(std::ptrdiff_t n)
```
CaseyCarter: `RANGES_CXX14_CONSTEXPR`

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+            {
+                RANGES_EXPECT(to_ - from_ >= n * spacing_);
```
CaseyCarter: Maybe `RANGES_EXPECT(std::numeric_limits<T>::max() / spacing_ >= n);` before this?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+                return (that.from_ - from_) / spacing_;
+            }
+            CONCEPT_REQUIRES(std::is_floating_point<T>())
+            constexpr std::ptrdiff_t distance_to(linear_distribute_view const &that) const
+            {
+                return static_cast<std::ptrdiff_t>(std::floor((that.from_ - from_) / spacing_));
+            }
+        public:
+            constexpr linear_distribute_view() = default;
+            constexpr linear_distribute_view(T from, T to, std::size_t n)
```
CaseyCarter: Either `RANGES_CXX14_CONSTEXPR`, or encode the body as comma expressions in the member-initializer-list, a la
```c++
constexpr linear_distribute_view(T from, T to, std::size_t n)
  : from_((RANGES_EXPECT(n > 1), detail::move(from))),
    to_((
        RANGES_EXPECT(to > from_),
        RANGES_EXPECT((to - from_) / static_cast<T>(n - 1) > T{0}),
        detail::move(to)))
{}
```

Love those C++11 `constexpr` hacks.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+        namespace view
+        {
+            /// Distributes `n >= 2` values linearly in the closed interval `[from, to]`.
+            struct linear_distribute_fn
+            {
+                template<typename T,
+                    CONCEPT_REQUIRES_(std::is_arithmetic<T>())>
+                constexpr auto operator()(T from, T to, std::size_t n) const
+                    RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT(linear_distribute_view<T>{from, to, n})
```
CaseyCarter: Typical style is to not indent `RANGES_DECLTYPE_AUTO_XXX`, and to put the parens where braces would go in a "real" function body:
```c++
constexpr auto operator/*...*/
(
    linear_distribute_view<T>{from, to, n}
)
```

gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
gnzlbg: Fixed: I went with `RANGES_CXX14_CONSTEXPR`, nice workaround though.

include/range/v3/view/linear_distribute.hpp
gnzlbg: indeed

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,127 @@
...
+            {
+                return from_ > to_;
+            }
+            RANGES_CXX14_CONSTEXPR void prev()
+            {
+                from_ -= spacing_;
+            }
+            RANGES_CXX14_CONSTEXPR void advance(std::ptrdiff_t n)
+            {
+                RANGES_EXPECT(std::numeric_limits<T>::max() / spacing_ >= n);
```
CaseyCarter: `#include<limits>`

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,127 @@
...
+            }
+            CONCEPT_REQUIRES(std::is_integral<T>())
+            constexpr std::ptrdiff_t distance_to(linear_distribute_view const &that) const
+            {
+                return (that.from_ - from_) / spacing_;
+            }
+            CONCEPT_REQUIRES(std::is_floating_point<T>())
+            constexpr std::ptrdiff_t distance_to(linear_distribute_view const &that) const
+            {
+                return static_cast<std::ptrdiff_t>(std::floor((that.from_ - from_) / spacing_));
```
CaseyCarter: `#include<cmath>`

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,35 @@
...
+// Project home: https://github.com/ericniebler/range-v3
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+#include <range/v3/view/linear_distribute.hpp>
+
+int main() {
+
+  using ranges::view::linear_distribute;
+
+  auto float_eq = [](double a, double b) { return (int)std::floor(std::abs(a - b)) == 0; };
```
CaseyCarter: `#include<cmath>`

gnzlbg:
The failures here also look spurious, @ericniebler could you attempt to retry the failed jobs please?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+            T from_, to_, spacing_;
+
+            constexpr T read() const
+            {
+                return from_;
+            }
+            RANGES_CXX14_CONSTEXPR void next()
+            {
+                from_ += spacing_;
```
ericniebler: From Sean (emphasis mine):
> There are many ways to deal with the error at each step, two common ways are error accumulation and stochastic rounding. A simple method of error accumulation is to subtract off each segment **and re-divide the remaining part into the remaining number of segments**.

You're not doing it right. ;-) I think this is closer to Sean's intent:

```c++
from_ += (to_ - from_) / n_--;
```

... where `n_` is the (integral) number of segments. You're done when `n_` is zero.

include/range/v3/view/linear_distribute.hpp
ericniebler: I think this is a `ForwardRange`. Otherwise, the corrections for errors at each step gives you different answers. I suggest removing `prev`, `advance`, and `distance_to`, at least for floating point types.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,129 @@
...
+            }
+        public:
+            constexpr linear_distribute_view() = default;
+            RANGES_CXX14_CONSTEXPR linear_distribute_view(T from, T to, std::size_t n)
+            : from_(detail::move(from)), to_(detail::move(to))
+            {
+              RANGES_EXPECT(n > 1);
+              RANGES_EXPECT(to_ > from_);
+              spacing_ = (to_ - from_)/ static_cast<T>(n - 1);
+              RANGES_EXPECT(spacing_ > T{0});
```
ericniebler: Spacing.

include/range/v3/view/linear_distribute.hpp
gnzlbg: pinging @sean-parent . The implementation provided supports `RandomAccess`, has no divide in the tight loop, and produces the same results as Matlab's `linspace` function.

I think it is a mistake to enable `RandomAccess` for floating point values (i'll fix that) since `from_ = n * spacing_` and ` while (n > 0) from_ += spacing` won't produce the same result due to floating point errors, but with a constant spacing it can still be a `BidirectionalRange`.

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed

include/range/v3/view/linear_distribute.hpp
gnzlbg: So i've refactored the view into one for floating point numbers and one for integers, and for the floats one I've implemented it like @sean-parent suggested, so its not random access anymore. We can tweak it later to something different if we feel like it in a smaller PR.

gnzlbg:
I still need to look into why the floating point range is not bidirectional.

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,60 @@
...
+                            float_eq));
+    }
+    {
+        auto frng = linear_distribute(1.0, 3.0, 21);
+        CHECK(ranges::equal(frng,
+                            {1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9,
+                             2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8, 2.9,
+                             3.0},
+                            float_eq));
+    }
```
ericniebler: You need to test the `to==from && n==1` case, for both integral and floating point types. And I'm curious what happens in the `to==from && n>1` case.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+            private:
+                friend range_access;
+                T from_, to_, spacing_;
+                constexpr T read() const
+                {
+                    return from_;
+                }
+                constexpr bool equal(default_sentinel) const
+                {
+                  return spacing_ == T{0}? from_ != to_ : from_ == to_ + spacing_;
```
CaseyCarter: Couldn't this simply `return from_ > to_;`?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                    if (spacing_ == T{0})
+                    {
+                      RANGES_EXPECT(to_ == from_);
+                      to_ += 1;
+                    }
+                    from_ += spacing_;
+                }
+                RANGES_CXX14_CONSTEXPR void prev()
+                {
+                    from_ -= spacing_;
```
CaseyCarter: Doesn't this need `if (spacing_ == T{0}) { --to_; RANGES_EXPECT(to_ == from_); }` to mirror `next`?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                }
+                RANGES_CXX14_CONSTEXPR void advance(std::ptrdiff_t n)
+                {
+
```
CaseyCarter: `advance(1)` seems to not be equivalent to `next()` when `spacing_ == 0`.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                }
+                RANGES_CXX14_CONSTEXPR void advance(std::ptrdiff_t n)
+                {
+                    
+                  RANGES_EXPECT(spacing_ == T{0} || std::numeric_limits<T>::max() / spacing_ >= n);
+                    RANGES_EXPECT(to_ - from_ >= n * spacing_);
+                    from_ += n * spacing_;
+                }
+                constexpr std::ptrdiff_t distance_to(linear_distribute_integral_view const &that) const
+                {
```
CaseyCarter: `RANGES_EXPECT(spacing_ != T{0});`?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,183 @@
...
+                    RANGES_EXPECT(n > 0);
+                    RANGES_EXPECT(to_ >= from_);
+                    if (to_ == from_)
+                    {
+                        spacing_ = 0;   
+                    }
+                    else
+                    {
+                        RANGES_EXPECT(n > 1);
+                        spacing_ = (to_ - from_)/ static_cast<T>(n - 1);
```
CaseyCarter: > You can replace the `if`/`else` with:

If `to_ == from_` and `n == 1`, that divides by zero.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                constexpr std::ptrdiff_t distance_to(linear_distribute_integral_view const &that) const
+                {
+                    return (that.from_ - from_) / spacing_;
+                }
+            public:
+                linear_distribute_integral_view() = default;
+                linear_distribute_integral_view(T from, T to, std::ptrdiff_t n)
+                  : from_(from), to_(to) 
+                {
+                    RANGES_EXPECT(n > 0);
```
CaseyCarter: This is redundant with the checks in the `if` / `else`, and `to >= from_` is partially redundant. Why not replace the whole body with:
```c++
if (to_ == from_)
{
    RANGES_EXPECT(n == 1);
    spacing_ = 0;
}
else
{
    RANGES_EXPECT(n > 1);
    RANGES_EXPECT(from < to_);
    spacing = (to_ - from_) / static_cast<T>(n - 1);
}
```

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                {
+                    return spacing_ == T{0}? 1 : static_cast<std::size_t>((to_ - from_) / spacing_ + 1);
+                }
+            };
+
+            template <typename T>
+            struct linear_distribute_floating_point_view
+              : view_facade<linear_distribute_floating_point_view<T>, finite>
+            {
+                static_assert(std::is_floating_point<T>(), "");
```
CaseyCarter: Ditto `CONCEPT_ASSERT`

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,183 @@
...
+                std::ptrdiff_t n_;
+                
+                constexpr T read() const
+                {
+                    return from_;
+                }
+                constexpr bool equal(default_sentinel) const
+                {
+                    return n_ == 0;
+                }
```
CaseyCarter: > You need to add an `equal` that compares...

...in order to model `Sentinel<I, I>()` and `EqualityComparable<I>()`. Is there any motivation for doing so in addition to modeling `InputIterator<I>()` and `Sentinel<default_sentinel, I>()`?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+                RANGES_CXX14_CONSTEXPR void next()
+                {
+                    RANGES_EXPECT(n_ > 0);
+                    from_ += (to_ - from_) / T(--n_);
+                }
+                RANGES_CXX14_CONSTEXPR void prev()
+                {
+                    from_ -= (to_ - from_) / T(++n_);
```
CaseyCarter: I suspect there are corner cases where this does not perfectly invert the behavior of `next`. No correction yet suggested, but we need to be careful here.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,183 @@
...
+                }
+            public:
+                linear_distribute_floating_point_view() = default;
+                linear_distribute_floating_point_view(T from, T to, std::ptrdiff_t n)
```
CaseyCarter: and `noexcept`.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+        /// \endcond
+
+        /// \addtogroup group-views
+        /// @{
+
+        /// Distributes `n >= 2` values linearly in the closed interval [`from`, `to`].
+        template <typename T>
+        using linear_distribute_view
+        = meta::if_c<std::is_floating_point<T>::value,
```
CaseyCarter: Either this line needs an indent, or the next two need outdents.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,183 @@
...
+                )
+
+            #ifndef RANGES_DOXYGEN_INVOKED
+                template<typename T,
+                    CONCEPT_REQUIRES_(not std::is_arithmetic<T>())>
+                void operator()(T, T, std::ptrdiff_t) const
+                {
+                    CONCEPT_ASSERT_MSG(std::is_arithmetic<T>(),
+                        "The object passed to view::linear_distribute must model the Arithmetic "
+                        "concept; that is, it must be one of the native integer of floating-point types");
```
CaseyCarter: Do we need to perform that relaxation now, before accepting this PR?

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,80 @@
...
+    auto float_eq = [](double a, double b) {
+        return (int)std::floor(std::abs(a - b)) == 0;
+    };
+
+    {
+        auto irng = linear_distribute(0, 1, 2);
+        CHECK(ranges::size(irng) == std::size_t{2});
+        CHECK(ranges::at(irng, 0) == 0);
+        CHECK(ranges::at(irng, 1) == 1);
+        static_assert(ranges::RandomAccessRange<decltype(irng)>(), "");
```
CaseyCarter: `CONCEPT_ASSERT` all of these, please.

gnzlbg:
So this was all wrong. In Matlab, there are two different types of these kinds of ranges:

1. `linspace`: what is now `view::linear_distribute`, with the following semantics: 
  - [`x1`, `x2`]: Point interval, specified as a pair of numeric scalars. `x1` and `x2` define the interval over which linspace generates points. `x1` and `x2` can be real or complex, and `x2` can be either larger or smaller than `x1`. If `x2` is smaller than `x1`, then the vector contains descending values.
 - `n`: Number of points, specified as a real numeric scalar. If `n` is `1`, linspace returns `x2`. If `n` is zero or negative, linspace returns an empty 1-by-0 matrix. If `n` is not an integer, linspace rounds down and returns `floor(n)` points.

2. The colon operator (`:`): `x = j:i:k`creates a regularly-spaced vector `x` using `i` as the increment between elements. The vector elements are roughly equal to `[j,j+i,j+2*i,...,j+m*i]` where `m = fix((k-j)/i)`. However, if `i` is not an integer, then floating point arithmetic plays a role in determining whether colon includes the endpoint `k` in the vector, since `k` might not be exactly equal to `j+m*i`. If you specify nonscalar arrays, then MATLAB interprets `j:i:k` as `j(1):i(1):k(1)`.

I've now chosen for `view::linear_distribute` to stick to the semantics of `linspace`. The only change is that it does not support "descending" ranges, since in range-v3 one can just use `view::reverse` to opt into that. 

The colon operator `:` should probably belong into something like `view::linear_step`, where instead of the number of points, one specifies the step to use.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,115 @@
...
+                    return from_;
+                }
+                constexpr bool equal(default_sentinel) const
+                {
+                    return n_ == 0;
+                }
+                RANGES_CXX14_CONSTEXPR void next()
+                {
+                    RANGES_EXPECT(n_ > 0);
+                    if (n_ == 1)
```
CaseyCarter: `if (--n_ == 0)`, and remove the decrements in the branches?

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,115 @@
...
+        namespace view
+        {
+            /// \addtogroup group-views
+            /// @{
+
+            template <typename T>
+            struct linear_distribute_view
+              : view_facade<linear_distribute_view<T>, finite>
+            {
+                static_assert(std::is_arithmetic<T>(), "");
```
CaseyCarter: Please `CONCEPT_ASSERT(foo)` all of these `static_assert(foo, "")` declarations.

include/range/v3/view/linear_distribute.hpp
gnzlbg: I've moved the decrement outside the conditional.

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed here and in the .cpp file

test/view/linear_distribute.cpp
gnzlbg: done

include/range/v3/view/linear_distribute.hpp
gnzlbg: Indeed, this cannot work. @ericniebler 

I am making the range `ForwardRange` for now. Making it `Bidirectional` is a bit trickier, might do that in a follow-up PR.

include/range/v3/view/linear_distribute.hpp
gnzlbg: @ericniebler I've stick to a `ForwardRange` for all types now. Relaxing this to a `BidirectionalRange` turned out to be tricky to get right for floating point types.

include/range/v3/view/linear_distribute.hpp
gnzlbg: fixed.

include/range/v3/view/linear_distribute.hpp
gnzlbg: I'm with Casey on this, we should have a solid concept for user-defined "integers" before trying to bolt it on here. Having a couple of views that require `std::is_integral` (or `std::is_arithmetic`), will help getting that concept right.

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,60 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014
```
gnzlbg: fixed

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,60 @@
...
+// Project home: https://github.com/ericniebler/range-v3
+#include <iostream>
+#include <cmath>
+#include <range/v3/algorithm/equal.hpp>
+#include <range/v3/view/linear_distribute.hpp>
+#include <range/v3/view/transform.hpp>
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+
+int main() {
```
gnzlbg: fixed

test/view/linear_distribute.cpp
gnzlbg: done

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+
+        /// \cond
+        namespace detail
+        {
+            template <typename T>
+            struct linear_distribute_integral_view
+              : view_facade<linear_distribute_integral_view<T>, finite>
+            {
+                static_assert(std::is_integral<T>(), "");
```
gnzlbg: fixed

gnzlbg:
So I've simplified this as much as possible, it now only models `ForwardRange`, anything else is left as an extension to follow up PRs. I've updated the PR info and the commit message and rebased.

sean-parent:
It has been a _long_ time since I've done any significant work in this space, but providing some info that may help. linear_distribute is a simple case of a linear recurrence algorithm. These are very important in many domains, including graphics and physics. In graphics, a linear_distribution is the answer to "how do you draw a line". The problem with doing this with a discrete machine is how to do error propagation. The simple way is to simply recalculate at each step:
```cpp
template <class Out>
void iterative(int f, int l, unsigned n, Out out) {
    while (n != 0) {
        f += (l - f) / n;
        --n;
        *out++ = f;
    }
    assert((n == 0) || (f == l));
}
```
(just using int here as a placeholder). This gives a very simple solution to error distribution. To do this quickly, you can use forward differencing. Bresenham's algorithm for drawing a line is a forward differencing algorithm to do linear distribution. Another common approach is to recursively subdivide. Recursive subdivision has a couple of advantages over forward differencing, it typically requires far fewer bits of precision at each step, and it allows you to drill to any location within the solution in log(n) steps. The downside is higher cost at each step and log(n) storage. However, if you want a "random access" linear distribution, this is how to get it. The recursive form of the above code is:
```cpp
template <class Out>
void recursive(int f, int l, int n, Out out) {
    if (n == 0) return;
    if (n & 1) {
        f += (l - f) / n;
        --n;
        *out++ = f;
    }
    auto m = f + (l - f) / 2;

    recursive(f, m, n / 2, out);
    recursive(m, l, n / 2, out);
}
```
Of course, one can use hybrid solutions between the two. There is a _lot_ of work that has been done in this space for graphics. The same basic approach is used for drawing nearly all graphics primitives (lookup how to draw cubic and quadratic Bézier curves). In Physics you see the same approaches to plot trajectories at a given time intervals (as one example), and to find collision points.

The same basic algorithm applies to distributing floating point values, you simply have a smaller epsilon in that case. Note that iterators behave enough like integers that you can (and it is very useful to) have a distribution of iterators.

include/range/v3/utility/iterator.hpp
```diff
@@ -189,7 +189,7 @@ namespace ranges
             {
                 RANGES_EXPECT((Same<I, S>() || 0 <= n));
                 D d = bound - it;
-                RANGES_EXPECT(0 <= n ? 0 <= d : 0 >= d);
+                //RANGES_EXPECT(0 <= n ? 0 <= d : 0 >= d);
```
CaseyCarter: Commenting out a precondition check is literally *never* the right thing.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,116 @@
...
+                constexpr T read() const noexcept
+                {
+                    return from_;
+                }
+                constexpr bool equal(default_sentinel) const noexcept
+                {
+                    return n_ == 0;
+                }
+                constexpr bool equal(linear_distribute_view const& other) const noexcept
+                {
```
CaseyCarter: `RANGES_EXPECT(from_ == other.from_ && to_ == other.to_)`?

test/view/linear_distribute.cpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+        auto irng = linear_distribute(1, 3, 3);
+        auto il = {1, 2, 3};
+        check_equal(irng, il);
+        CHECK(ranges::size(irng) == std::size_t{3});
+    }
+    {
+        auto irng = linear_distribute(0, 21, 22);
+        auto frng = linear_distribute(0., 21., 22);
+        check_equal(irng, frng | ranges::view::transform([](double v) { return int(v); }));
```
CaseyCarter: `[](double v) { return int(v); }` could be `ranges::convert_to<int>{}` from `functional.hpp`.

include/range/v3/utility/iterator.hpp
gnzlbg: oops, this shouldn't haven't been commented out. I had a lot of debugging logic in there when i was trying to reverse engineer what `distance_to(view const& other)` actually was supposed to do when doing the RandomAccess version.

include/range/v3/view/linear_distribute.hpp
gnzlbg: I guess now is the time to ask my questions about `equal(view const& other)` and `distance_to(view const& other)` when are these actually called and for what? 

If they are only called for sub-views of the same range, then the assertion is correct, but if they can be called for completely independent ranges, then that should be part of the result.

sean-parent:
One more comment - there is more than one way to distribute error in a linear distribution. For example, if you are a drawing a linear gradient (a linear distribution of colors) than often it is better to use a stochastic model where the accumulated error is a probability of the position. This avoids visible rounding artifacts.

include/range/v3/view/linear_distribute.hpp
CaseyCarter: `equal(view const&)` is called by `operator==(basic_iterator<view> const&, basic_iterator<view> const&)`. You can support (slowly) comparing iterators from different ranges by returning `std::tie(from_, to_, n_) == std::tie(that.from_, that.to_, that.n_)`, or you can declare that only iterators from the same range are in the domain of equality and return `(RANGES_EXPECT(std::tie(from_, to_) == std::tie(that.from_, that.to_)), n_ == that.n_)` or `(RANGES_EXPECT(std::tie(to_, n_) == std::tie(that.to_, that.n_)), from_ == that.from_)`. The flavor that compares only a single integer will probably be the fastest in release builds. In the name of "don't pay for what you don't use," range-v3 always chooses speed for comparison with ranges over enlarging the domain of `==`. (It would be awesome if STL2 were to have distinct hooks for "fast comparison" using operators, and "total ordering" using some other customization point so that we could usefully provide both operations on iterators.)

`distance_to(view const& that)` is called by `operator-(basic_iterator<view> const&, basic_iterator<view> const&)`. Similar reasoning applies here, although I think you would have a much harder time making this operation work across distinct ranges than `equal`.

gnzlbg:
@sean-parent thanks for the pointers. The current implementation is using the simple solution you mention, I'll try to extend it to recursive subdivision, since I think the `log(N)` stack space cost is worth it for `RandomAccess`. 

I think that at least for floating point values it is important to document the error distribution being used. 

Using a stochastic model would probably belong into a different `view`. I don't really see an easy way of allowing the user to customize the error distribution without basically having to re-implement the whole `view`.

include/range/v3/view/linear_distribute.hpp
gnzlbg: That makes it clear. I will do this for integers but will omit the `RANGES_EXPECT` for floating point values (since comparing for equality there does not make sense without some sort of tolerance).

sean-parent:
My algorithms also only handle the case where the range is larger than n correctly (technically, the distance(f,l) measured in epsilons must not be < n). Otherwise iterative and recursive differ and you get a pile-up with iterative at the beginning). The normal way to deal with that in graphics is to flip coordinates (there is inherent symmetry here, a linear distribution is a line (f,0),(l, n) when distance(f,l) < n then you can swap x and y and use the same algorithm to get proper distribution. Algorithmically it is useful to have a distribution of 0 elements (returning nothing). Don't know if that fits in the view model.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,124 @@
...
+                bool equal(linear_distribute_view const& other) const noexcept
+                {
+                    RANGES_EXPECT(from_ == other.from_ && to_ == other.to_);
+                    return n_ == other.n_;
+                }
+                CONCEPT_REQUIRES(std::is_floating_point<T>())
+                constexpr bool equal(linear_distribute_view const& other) const noexcept
+                {
+                    return n_ == other.n_;
+                }
```
ericniebler: Why keep these two `equal` overloads separate instead of combining them with an `EXPECT` like:

```
RANGES_EXPECT(std:: is_floating_point <T>()  || (from_ == other.from_ && to_ == other.to_));
 ```

OK, so the floating point case is no longer C++11 `constexpr`, but I don't think that matters.

File: include/range/v3/view/linear_distribute.hpp [hunk 36]
```diff
@@ -0,0 +1,124 @@
...
+                    {
+                        from_ = to_;
+                    }
+                    else
+                    {
+                        from_ += (to_ - from_) / T(n_);
+                    }
+                }
+            public:
+                RANGES_CXX14_CONSTEXPR linear_distribute_view() noexcept = default;
```
ericniebler: s/`noexcept = default`/`= default`/

include/range/v3/view/linear_distribute.hpp
gnzlbg: Because floating point comparison with `==` (even if it is never invoked) did raise warnings (unless one uses `if constexpr`). This was the "nicest" solution I found :/

CaseyCarter: > the floating point case is no longer C++11 constexpr

*Neither* case would be C++11 constexpr, although that's easily fixed by combining `RANGES_EXPECT(/**/)` and `n_ == other.n_` into a comma expression.

CaseyCarter: > Because floating point comparison with == (even if it is never invoked) did raise warnings

Far from expert on anything to do with fp, but I think the warnings are spurious here. If we cannot consistently and exactly repeat the values of `from_` and `to_` on repeated passes over the same range - and I think this implementation correctly does so - then we are violating the iterator requirements and we need to fix it.

ericniebler: Agreed. Copies of floating point values should be equality comparable, and floating point operations are deterministic and equality preserving to the best of my knowledge. (I hate to think of the implications if they weren't.)

I think one `equal` that does `RANGES_EXPECT(from_ == other.from_ && to_ == other.to_)` with a warning suppression would be best.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -0,0 +1,118 @@
...
+                    return from_;
+                }
+                constexpr bool equal(default_sentinel) const noexcept
+                {
+                    return n_ == 0;
+                }
+                RANGES_CXX14_CONSTEXPR
+                bool equal(linear_distribute_view const& other) const noexcept
+                {
+                    RANGES_EXPECT(from_ == other.from_ && to_ == other.to_);
```
ericniebler: This needs a warning suppression.

```
/Users/travis/build/ericniebler/range-v3/include/range/v3/view/linear_distribute.hpp:52:41: error: comparing floating point with == or != is unsafe [-Werror,-Wfloat-equal]
                    RANGES_EXPECT(from_ == other.from_ && to_ == other.to_);
                                  ~~~~~ ^  ~~~~~~~~~~~
```

CaseyCarter: > This needs a warning suppression.

https://github.com/gnzlbg/range-v3/pull/6

gnzlbg:
travis is all green

ericniebler:
Thanks for all your work on this, @gnzlbg !

skgbanga:
Ah! That's very informative. Thanks!

PR #587, opened 2017-03-15
action::reverse

### LINKED ISSUE #477: No action::reverse?
I have some code that tries to set up a map/dictionary for sorting purposes. My intended usage is:

``` cpp
rmap[join(split(line, '.') | action::reverse, '.')] = line;
```

This is to map the Python code as follows:

``` python
rmap['.'.join(reversed(line.split('.')))] = line
```

Regretfully, the code does not work. I also tried `view::reverse`, but ranges-v3 complains with a static assertion:

> You can't pipe an rvalue container into a view. First, save the container into a named variable, and then pipe it to the view.

Should `action::reverse` be provided? Or any workaround (except following the ranges-v3 prompt, which I know already and regard a bit ugly)?

gnzlbg:
Implements action::reverse.

Closes #477.

### DIFF
--- a/include/range/v3/action/reverse.hpp
+++ b/include/range/v3/action/reverse.hpp
@@ -1,0 +1,88 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2014
+//  Copyright Gonzalo Brito Gadeschi 2017
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_ACTION_REVERSE_HPP
+#define RANGES_V3_ACTION_REVERSE_HPP
+
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/action/action.hpp>
+#include <range/v3/algorithm/reverse.hpp>
+#include <range/v3/utility/iterator_concepts.hpp>
+#include <range/v3/utility/static_const.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        /// \addtogroup group-actions
+        /// @{
+        namespace action
+        {
+            /// \cond
+            namespace reverse_detail
+            {
+                template<typename Rng>
+                using Concept = meta::and_<
+                    BidirectionalRange<Rng>,
+                    // Only evaluate this one if the previous one succeeded
+                    meta::lazy::invoke<
+                        meta::compose<
+                            meta::quote<Permutable>,
+                            meta::quote<iterator_t>>,
+                        Rng>>;
+            }  // namespace reverse_detail
+            /// \endcond
+
+            /// Reversed the source range in-place.
+            struct reverse_fn
+            {
+            private:
+                friend action_access;
+
+            public:
+                template<typename Rng, typename I = iterator_t<Rng>,
+                    CONCEPT_REQUIRES_(BidirectionalRange<Rng>() && Permutable<I>())>
+                Rng operator()(Rng && rng) const
+                {
+                    ranges::reverse(rng);
+                    return std::forward<Rng>(rng);
+                }
+
+            #ifndef RANGES_DOXYGEN_INVOKED
+                template<typename Rng,
+                         CONCEPT_REQUIRES_(!reverse_detail::Concept<Rng>())>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(BidirectionalRange<Rng>(),
+                        "The object on which action::reverse operates must be a model of the "
+                        "BidirectionalRange concept.");
+                    CONCEPT_ASSERT_MSG(Permutable<iterator_t<Rng>>(),
+                        "The iterator type of the range passed to action::reverse must allow its "
+                        "elements to be permuted; that is, the values must be movable and the "
+                        "iterator must be mutable.");
+                }
+            #endif
+            };
+
+            /// \ingroup group-actions
+            /// \relates reverse_fn
+            /// \sa action
+            RANGES_INLINE_VARIABLE(action<reverse_fn>, reverse)
+        }
+        /// @}
+    }
+}
+
+#endif
--- a/test/action/reverse.cpp
+++ b/test/action/reverse.cpp
@@ -1,0 +1,36 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#include <vector>
+#include <range/v3/core.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/repeat_n.hpp>
+#include <range/v3/view/for_each.hpp>
+#include <range/v3/algorithm/equal.hpp>
+#include <range/v3/action/reverse.hpp>
+#include <range/v3/action/unique.hpp>
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+
+using namespace ranges;
+
+int main()
+{
+    // [1,2,2,3,3,3,4,4,4,4,5,5,5,5,5,...]
+    std::vector<int> v =
+        view::for_each(view::ints(1,6), [](int i){
+            return yield_from(view::repeat_n(i,i));
+        });
+    check_equal(v, {1,2,2,3,3,3,4,4,4,4,5,5,5,5,5});
+
+    v |= action::unique | action::reverse;
+    CHECK(equal(v, {5,4,3,2,1}));
+
+    return ::test_result();
+}
--- a/include/range/v3/action.hpp
+++ b/include/range/v3/action.hpp
@@ -23,6 +23,7 @@
 #include <range/v3/action/push_back.hpp>
 #include <range/v3/action/push_front.hpp>
 #include <range/v3/action/remove_if.hpp>
+#include <range/v3/action/reverse.hpp>
 #include <range/v3/action/shuffle.hpp>
 #include <range/v3/action/slice.hpp>
 #include <range/v3/action/sort.hpp>
--- a/test/action/CMakeLists.txt
+++ b/test/action/CMakeLists.txt
@@ -19,6 +19,9 @@ add_test(test.act.push_back act.push_back)
 add_executable(act.remove_if remove_if.cpp)
 add_test(test.act.remove_if act.remove_if)
 
+add_executable(act.reverse reverse.cpp)
+add_test(test.act.reverse act.reverse)
+
 add_executable(act.shuffle shuffle.cpp)
 add_test(test.act.shuffle act.shuffle)
 


include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,79 @@
+/// \file
```
CaseyCarter: This header needs to be added to `include/range/v3/action.hpp`.

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,79 @@
...
+                }
+                template<typename Rng, typename I = iterator_t<Rng>,
+                    CONCEPT_REQUIRES_(BidirectionalRange<Rng>() and !Permutable<I>())>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(Permutable<I>(),
+                                   "The iterator type of the range passed to action::reverse must allow its "
+                                   "elements to be permuted; that is, the values must be movable and the "
+                                   "iterator must be mutable.");
+              }
```
CaseyCarter: +2

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,79 @@
...
+                    CONCEPT_ASSERT_MSG(BidirectionalRange<Rng>(),
+                        "The object on which action::reverse operates must be a model of the "
+                        "BidirectionalRange concept.");
+                }
+                template<typename Rng, typename I = iterator_t<Rng>,
+                    CONCEPT_REQUIRES_(BidirectionalRange<Rng>() and !Permutable<I>())>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(Permutable<I>(),
+                                   "The iterator type of the range passed to action::reverse must allow its "
```
CaseyCarter: these lines have some extra indentation.

include/range/v3/action/reverse.hpp
gnzlbg: added

include/range/v3/action/reverse.hpp
gnzlbg: fixed

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,79 @@
...
+        namespace action
+        {
+            struct reverse_fn
+            {
+            private:
+                friend action_access;
+
+            public:
+                template<typename Rng, typename I = iterator_t<Rng>,
+                         CONCEPT_REQUIRES_(BidirectionalRange<Rng>() and Permutable<I>())>
```
CaseyCarter: Woops - I missed this over-indented line on the first pass. Sorry.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <range/v3/action/sort.hpp>
+#include <range/v3/action/reverse.hpp>
+#include <range/v3/action/unique.hpp>
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+
+int main()
+{
+    using namespace ranges;
+    std::mt19937 gen;
```
CaseyCarter: This is unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#include <random>
```
CaseyCarter: Unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <vector>
+#include <range/v3/core.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/repeat_n.hpp>
+#include <range/v3/view/for_each.hpp>
+#include <range/v3/view/take.hpp>
+#include <range/v3/algorithm/shuffle.hpp>
```
CaseyCarter: Unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <range/v3/algorithm/equal.hpp>
+#include <range/v3/algorithm/is_sorted.hpp>
+#include <range/v3/action/shuffle.hpp>
```
CaseyCarter: Unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <random>
+#include <vector>
+#include <range/v3/core.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/repeat_n.hpp>
+#include <range/v3/view/for_each.hpp>
+#include <range/v3/view/take.hpp>
+#include <range/v3/algorithm/shuffle.hpp>
+#include <range/v3/algorithm/equal.hpp>
+#include <range/v3/algorithm/is_sorted.hpp>
```
CaseyCarter: Unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <range/v3/action/shuffle.hpp>
+#include <range/v3/action/sort.hpp>
```
CaseyCarter: Unused.

test/action/reverse.cpp
```diff
@@ -0,0 +1,43 @@
...
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+
+#include <random>
+#include <vector>
+#include <range/v3/core.hpp>
+#include <range/v3/view/iota.hpp>
+#include <range/v3/view/repeat_n.hpp>
+#include <range/v3/view/for_each.hpp>
+#include <range/v3/view/take.hpp>
```
CaseyCarter: Unused.

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,79 @@
...
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_ACTION_REVERSE_HPP
+#define RANGES_V3_ACTION_REVERSE_HPP
+
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/action/action.hpp>
+#include <range/v3/action/erase.hpp>
```
CaseyCarter: Unused?

include/range/v3/action/reverse.hpp
gnzlbg: I wish we had a clang-format file... when I want to edit range-v3 I need to indent everything manually by inserting spaces, and working against the indentation mechanism of my editor (and the tab key, which for me just invokes clang-format on the whole file) :/ 

clang-format isn't perfect though :/

include/range/v3/action/reverse.hpp
gnzlbg: yep, fixed

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,78 @@
...
+        namespace action
+        {
+            struct reverse_fn
+            {
+            private:
+                friend action_access;
+
+            public:
+                template<typename Rng, typename I = iterator_t<Rng>,
+                    CONCEPT_REQUIRES_(BidirectionalRange<Rng>() and Permutable<I>())>
```
ericniebler: No iso646 keywords, please.

include/range/v3/action/reverse.hpp
```diff
@@ -0,0 +1,78 @@
...
+                }
+                template<typename Rng, typename I = iterator_t<Rng>,
+                    CONCEPT_REQUIRES_(BidirectionalRange<Rng>() and !Permutable<I>())>
+                void operator()(Rng &&) const
+                {
+                    CONCEPT_ASSERT_MSG(Permutable<I>(),
+                        "The iterator type of the range passed to action::reverse must allow its "
+                        "elements to be permuted; that is, the values must be movable and the "
+                        "iterator must be mutable.");
+              }
```
ericniebler: I have typically put all such `CONCEPT_ASSERT`s in a single overload to keep boilerplate under control. I would prefer that here.

Also, that trailing `}` looks funny hanging out in space there.

include/range/v3/action/reverse.hpp
gnzlbg: fixed both of them

include/range/v3/action/reverse.hpp
gnzlbg: Fixed :)

gnzlbg:
@ericniebler the nitpicks should be fixed, but I wonder whether merging the "static asserts" overloads into one really works. In particular, if an user calls it with something that is not a range, the overload won't produce a `static_assert`, because `range_iterator_t<Rng>` will fail to instantiate. Or should I remove it from the `CONCEPT_REQUIRES_` clause aus well?

ericniebler:
> if an user calls it with something that is not a range, the overload won't produce a `static_assert`, because `range_iterator_t<Rng>` will fail to instantiate.

Oh, you're right. This is a problem elsewhere as well (looking at `action::sort` for instance). I suspect we can change it to something like:

```c++
template <class Rng>
using Concept =
    meta::let<
        meta::lazy::and_<
            meta::defer<BidirectionalRange, Rng>,
            meta::defer<Permutable, meta::defer<iterator_t, Rng> > > >;
```

And then constrain the two overloads with `Concept<Rng>()` and `!Concept<Rng>()`. I don't particularly care for this solution because `meta::let` with nested `defer`s is expensive at compile time. A lighter weight alternative would be something like:

```c++
template <class Rng, bool = (bool) BidirectionalRange<Rng>()>
struct Concept : std::false_type {};
template <class Rng>
struct Concept<Rng, true> : Permutable<iterator_t<Rng>> {};
```

This would compile faster, but it doesn't generalize very well.

I'm really looking forward to real Concepts and this awful hackery can go away.

gnzlbg:
@ericniebler can you restart the two jobs that failed? The errors _might_ be spurious, but I haven't seen them before: 

```shell

fatal error: file '/usr/include/asm-generic/errno-base.h' has been modified
      since the module file
      '/home/travis/build/ericniebler/range-v3/build/module.cache/37BAM9G4AAKMY/range_v3-IUI8NG69BABM.pcm'
      was built
note: '/usr/include/asm-generic/errno-base.h' required by
      '/home/travis/build/ericniebler/range-v3/build/module.cache/37BAM9G4AAKMY/std-5743FB4UA0N1.pcm'
note: 
      '/home/travis/build/ericniebler/range-v3/build/module.cache/37BAM9G4AAKMY/std-5743FB4UA0N1.pcm'
      required by
      '/home/travis/build/ericniebler/range-v3/build/module.cache/37BAM9G4AAKMY/range_v3-IUI8NG69BABM.pcm'
note: please rebuild precompiled header
      '/home/travis/build/ericniebler/range-v3/build/module.cache/37BAM9G4AAKMY/range_v3-IUI8NG69BABM.pcm'
1 error generated.
```

That just shouldn't happen, nothing should have modified the `limits.h`, but maybe the clock of the build bot was unsynchronized somehow.

---

About the issue with the static asserts, I think we should open an issue to come up with a solution that we can apply library wide. From all the alternatives you propose, I actually prefer my first approach of just having two overloads :/

ericniebler:
> About the issue with the static asserts, I think we should open an issue to come up with a solution that we can apply library wide.

I could but I'm not sure what would be better. The two overloads are a bit nasty. It gets worse the more constraints an algorithm has. I kinda hate the `!Concept<Rng>()` overloads; we could simply remove them, but that leaves users with mysterious SFINAE errors and little guidance. 

For now, can you please use:

```c++
template<typename Rng>
using Concept = meta::and_<
    BidirectionalRange<Rng>,
    // Only evaluate this one if the previous one succeeded
    meta::lazy::invoke<
        meta::compose<
            meta::quote<Permutable>,
            meta::quote<iterator_t>>,
        Rng>>;
```

It's awful, but has the benefit of being consistent with what is done elsewhere in the library.

gnzlbg:
Done :)

PR #589, opened 2017-03-15
Clang/C2 fixes

CaseyCarter:
Clang/C2 warns on unneeded members - as do all clangs - and doesn't define `__GNUC__`; use `RANGES_DIAGNOSTIC_IGNORE_UNNEEDED_MEMBER` to disable the warning.

Don't check `is_nothrow_destructible` with an incomplete type.

Drive-by: `is_nothrow_swappable` is specified to symmetrically check both `noexcept(swap(a, b)) and noexcept(swap(b, a))`

Don't instantiation `is_swappable<Base>` in `tagged` when the class is instantiated. (VC++ apparently doesn't constrain `pair`'s assignment operator.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ Supported Compilers
 
 The code is known to work on the following compilers:
 
-- clang 3.5.2
-- GCC 4.9.4 (C++14 support requires GCC 5.2; C++14 "extended constexpr" support is poor before 6.1.)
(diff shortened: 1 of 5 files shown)


CaseyCarter:
Bah, killing this PR to stop spawning Travis jobs while I run down more compile errors.

PR #590, opened 2017-03-16
Drop for_each_view, return join(transform(x)) instead

### LINKED ISSUE #566: view::move and view::for_each do not work together
view::for_each passes an lvalue reference to its functor, even if it is preceded by view::move. A workaround is to combine view::transform and view::join instead of using view::for_each directly.

Consider the following code:
```c++
#include <vector>
#include <memory>
#include <range/v3/view.hpp>

using namespace ranges;

int main()
{
  std::vector<std::unique_ptr<int>> data;
  data.emplace_back(std::make_unique<int>(42));

  // does not compile
  auto rng1 = data | view::move | view::for_each([](std::unique_ptr<int> ptr) {
    return yield(42);
  });

  // does compile
  auto rng2 = data | view::move | view::for_each([](const std::unique_ptr<int>& ptr) {
    return yield(42);
  });

  // workaround
  auto rng3 = data | view::move | view::transform([](std::unique_ptr<int> ptr) {
    return yield(42);
  }) | view::join;
}
```

FloopCZ:
Fixes #566, but I am not sure what could be the consequences.

### DIFF
--- a/include/range/v3/view/for_each.hpp
+++ b/include/range/v3/view/for_each.hpp
@@ -32,18 +32,6 @@ namespace ranges
 {
     inline namespace v3
     {
-        /// \addtogroup group-views
-        /// @{
-        template<typename Rng, typename F>
-        struct for_each_view
-          : join_view<transform_view<Rng, F>>
-        {
-            for_each_view() = default;
-            for_each_view(Rng rng, F f)
-              : for_each_view::join_view{{std::move(rng), std::move(f)}}
-            {}
-        };
-
         namespace view
         {
             struct for_each_fn
@@ -58,35 +46,11 @@ namespace ranges
                 )
             public:
                 template<typename Rng, typename F>
-                using Concept = meta::and_<
-                    Range<Rng>,
-                    IndirectInvocable<F, iterator_t<Rng>>,
-                    Range<concepts::Invocable::result_t<F&, range_common_reference_t<Rng>>>>;
-
-                template<typename Rng, typename F,
-                    CONCEPT_REQUIRES_(Concept<Rng, F>())>
-                for_each_view<all_t<Rng>, F> operator()(Rng && rng, F f) const
-                {
-                    return {all(std::forward<Rng>(rng)), std::move(f)};
-                }
-
-            #ifndef RANGES_DOXYGEN_INVOKED
-                // For better error reporting
-                template<typename Rng, typename F,
-                    CONCEPT_REQUIRES_(!Concept<Rng, F>())>
-                void operator()(Rng &&, F) const
-                {
-                    CONCEPT_ASSERT_MSG(Range<Rng>(),
-                        "Rng is not a model of the Range concept.");
-                    CONCEPT_ASSERT_MSG(IndirectInvocable<F, iterator_t<Rng>>(),
-                        "The function F is not callable with arguments of the type of the range's "
-                        "common reference type.");
-                    CONCEPT_ASSERT_MSG(Range<concepts::Invocable::result_t<F&,
-                        range_common_reference_t<Rng>>>(),
-                        "To use view::for_each, the function F must return a model of the Range "
-                        "concept.");
-                }
-            #endif
+                auto operator()(Rng && rng, F f) const
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    join(transform(std::forward<Rng>(rng), std::move(f)))
+                )
             };
 
             /// \relates for_each_fn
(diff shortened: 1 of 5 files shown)


include/range/v3/view/for_each.hpp
```diff
@@ -58,35 +46,11 @@ namespace ranges
...
-                        range_common_reference_t<Rng>>>(),
-                        "To use view::for_each, the function F must return a model of the Range "
-                        "concept.");
-                }
-            #endif
+                auto operator()(Rng && rng, F f) const
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    join(transform(std::forward<Rng>(rng), std::move(f)))
+                )
```
CaseyCarter: I'd like to do this without losing the error messages. I think the best way to do that is something like:
```c++
struct for_each_fn
{
private:
    template <class Rng, class F>
    static auto impl(int, Rng &&rng, F f)
    (
        join(transform(std::forward<Rng>(rng), std::move(f)))
    )
    template <class Rng, class F>
    static void impl(long, Rng &&, F)
    {
        // Like the pre-existing operator() with the CONCEPT_ASSERT_MSGs,
        // but updated to the actual current requirements of join and transform.
        // These messages may get out of date, but the view will keep working.
    }
public:
    auto operator()(Rng &&rng, F f) const
    (
        for_each_fn::impl(42, std::forward<Rng>(rng), std::move(f))
    )
};
```

CaseyCarter: No, Casey, that won't work. `join` and `transform` accept everything and have their own "bad" overloads of `operator()` that will diagnose failures. FloopCZ has done exactly the right thing here by not even constraining the parameters to `for_each_fn::operator()`, and just passing them through.

ericniebler: My concern is that users will use `view::for_each` and get an error in `view::join`, and not know why. We try to validate template parameters at the API boundaries, and this violates that. My preference would be to fix up the constraints to bring them into agreement with what `join` and `transform` expect.

CaseyCarter: > My preference would be to fix up the constraints to bring them into agreement with what join and transform expect.

I'd like a reminder near where `join` and `transform` define their constraints to tell us to update `for_each` when we change them.

FloopCZ:
Thank you for the review.

Added concept checks as requested by @ericniebler. The concepts are taken directly from view::transform. Added reminder to view::transform concepts as requested by @CaseyCarter.

`F f` renamed to `Fun fun` to comply with view::transform.

CaseyCarter:
> Added concept checks as requested by @ericniebler. The concepts are taken directly from `view::transform`.

What about the `view::join` requirements?

FloopCZ:
> What about the view::join requirements?

Those should always be satisfied. We require the base range to satisfy InputRange and we require view::transform to return InputRange's. InputRange of InputRange's should be joinable.

include/range/v3/view/for_each.hpp
```diff
@@ -32,61 +32,54 @@ namespace ranges
...
-                    CONCEPT_REQUIRES_(Concept<Rng, F>())>
-                for_each_view<all_t<Rng>, F> operator()(Rng && rng, F f) const
-                {
-                    return {all(std::forward<Rng>(rng)), std::move(f)};
-                }
+                template<typename Rng, typename Fun,
+                    CONCEPT_REQUIRES_(
+                      transform_fn::Concept<Rng, Fun>()
+                      && InputRange<concepts::Invocable::result_t<
+                          Fun&, range_reference_t<Rng>>>())>
```
ericniebler: If `Rng` is not a range, then `range_reference_t<Rng>` is ill-formed and this overload will not be selected. I suggest this:

```c++
template<typename Rng, typename F>
using Concept = meta::and_<
    transform_fn::Concept<Rng, Fun>,
    meta::lazy::invoke<
        meta::compose<
            meta::quote<InputRange>,
            meta::bind_front<meta::quote<concepts::Invocable::result_t>, Fun&>,
            meta::quote<range_reference_t>>,
        Rng>>;
```

Then use `Concept<Rng>()` to constrain the "good" overload and `!Concept<Rng>()` to constrain the "bad" one.

FloopCZ: Right, that one is tricky. Fixed.

FloopCZ:
Travis job is [broken](https://travis-ci.org/ericniebler/range-v3/jobs/212299335), it fails on installation of packages, so this pull request will not pass travis tests.

By the way, congratulations to 1000 GitHub stars! Great job!

Issue #122, opened 2015-03-25
view for creating half-open Integral ranges


gnzlbg:
There are currently two views for creating closed (aka inclusive) `Integral` ranges (`view::iota` and `view::ints`). There aren't any utilities for creating half-open `Integral` ranges. Most range interfaces offer an utility for half-open `Integral` ranges: e.g. `boost::counting_range`, Python's `xrange`/`range`, D's `iota`, ... I'd argue that half-open integer ranges are more useful than inclusive ones (e.g. indexing over a container is terribly common), but inclusive ones are also very useful (e.g. Matlab's `linspace` is inclusive and issue #102 ).

**Naming**: no idea.  In #16  I used `irng` to name an utility function for creating a half-open integer range but I don't like this name. I wish that both `ints` and `iota` would return half-open ranges, and that there would be a `linspace`-like function for creating inclusive ranges as mentioned in issue #102. 

**Notes**: for indexing a container `irng(0, size)` where `0 : int` and `size : std::size_t` is very common so the view should handle this.

I use this _a lot_ via: 

```c++
template <class From, class To, class CT = std::common_type_t<From, To>>
constexpr auto irng(From from, To to) {
  return take(iota(CT(from)), CT(to) - CT(from));
}
```

but this is not the best solution to the problem.

ericniebler:
Agree there's a need for both. Not sure we need a differently named view for it or whether the existing ones can be pressed into service. Maybe `view::ints(a, b)` is open and `view::ints({a,b})` is closed. Or `view::ints(closed{a,b})`.

gnzlbg:
I don't know yet about the `view::ints({a,b})` syntax (might be more trouble than it's worth), but I like that you also prefer half-open ranges from `ints` by default.

ericniebler:
... or `view::ints[{a,b}]` for closed. Square brackets are used to denote a closed range. Too bad I can't use `view::ints[a,b)` for half-open. ;-)

viboes:
... or `view::ints(closed{a}, open{b})`. The question is if a default for `view::ints(a, b)` would be needed?

gnzlbg:
_Iff_ a default is chosen, I think that it should be half-open. There are multiple good reasons for this (ability to handle empty ranges, indexing a container, ...) but the most important I think is consistency with all other ranges which are half-open.

ericniebler:
Fixed in 963ac26

PR #592, opened 2017-03-16
[breaking change] view::ints(N) returns [0, N) instead of [N, ...)

### LINKED ISSUE #277: `ints` and `iota` are not compatible with similar facilities in other languages.
If you call `view::ints`/`view::iota` with a single value, it assumes that this value is the starting point of an unbounded range. This is inconsistent with similar facilities in other languages.

In Python, [`range` will assume a single value is the terminus of a half-open range starting with 0](http://pythoncentral.io/pythons-range-function-explained/). This is done to facilitate looping over integer ranges:

```
for i in range(11)
```

D's [`iota` function works the same way](http://dlang.org/phobos/std_range.html#iota). A single value means a half-open range starting at 0 and ending with the given value.

I think that the Range library versions should follow existing practice. People may find it confusing to write:

```
for(auto i : view::iota(50))
```

And get an unbounded range starting at 50.

I recognize the importance of being able to create an unbounded range. But I think it would be better to do something like this: `view::iota(50, unbounded)`.

gnzlbg:
**This is a breaking change**

- Change the behavior of single argument `view::ints(N)` to return the half-open range `[0, N)` instead of the infinite range `[N, ...)`. This makes `view::ints` behave like Python's `range(N)`.
- Add single argument version of `view::closed_ints(N)` to return the closed-range `[0, N]`.

Closes #277.

### DIFF
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -488,17 +488,23 @@ namespace ranges
             {
                 ints_fn() = default;
 
-                template<typename Val,
-                    CONCEPT_REQUIRES_(Integral<Val>())>
-                iota_view<Val> operator()(Val value) const
-                {
-                    return iota_view<Val>{value};
-                }
                 template<typename Val>
                 meta::if_c<
                     (bool)Integral<Val>(),
                     detail::take_exactly_view_<iota_view<Val>, true>>
-                operator()(Val from, Val to) const;
+                operator()(Val from, Val to) const
+                {
+                    return {iota_view<Val>{from}, detail::iota_minus_(to, from)};
+                }
+
+                template<typename Val,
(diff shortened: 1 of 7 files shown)


ericniebler:
If this is the right design decision (and I'm not convinced it is) I would prefer simply deprecating the single argument overload and making people say what they mean for a release or two.

Also, as pointed out in #122, `iota` must do what the std algorithm does, and I don't like `ints` doing something different. Perhaps the answer is to deprecate `ints` entirely and add `view::indices` with the behavior you want.

gnzlbg:
@ericniebler I think that deprecation of `ints`/`closed_intst` and using `indices/inclusive_indices` is a good idea. I'll close this PR for now, and kick-start the discussion about the semantics that the view should have in #122 again. Until we all agree on the semantics it doesn't make sense to pursue this further.

Issue #591, opened 2017-03-16
Relax difference_type to support user-defined wrappers over `Integer` types


gnzlbg:
In issue #514, @CaseyCarter argues:

> One of the critical features of Integral types is that all of their operations take constant time. The complexity statements for the algorithms are vastly simplified by the requirement that all iterator operations are O(1). Relaxing the Integral requirement on difference types without maintaining that O(1) requirement - which is necessary to support difference types that can represent arbitrary distances in infinite ranges - will require reanalyzing algorithm complexities.

I would like to relax `difference_type` to support user-defined `Integral` types, where by `Integral` types I mean that the user-defined types _must_ satisfy the complexity guarantees that the `Integral` concept impose on native and implementation defined integer types. 

This would allow using "type safe"/"opaque" wrappers over native `Integral` types as `difference_types`, removing a another barrier to use them in combination with range-v3 (see #514 #356).

I am _not_ proposing relaxing the complexity guarantees of `Integer` to support `BigInt`s as different types to represent distances on infinite ranges, nor am I proposing adding a `BigInt` concept that can do that. While this _could_ be a useful feature, I think that if someone wants to pursue it, that should happen after this change anyways (and this feature is much more complex that what I am proposing as @CaseyCarter mentions).

gnzlbg:
So my proposed changes would be as follows: 

- Add an `BikeshedIntegral` concept with the exact same requirements as `Integral`, except the requirement that the integer type must be "part of the language or the implementation". 
- Use `BikeshedIntegral` instead of `Integral` in range-v3 where appropriate.

I would prefer to rename `Integral` to `NativeIntegral`, and `BikeshedIntegral` to `Integral` instead, but backwards compatibility with `std::is_integral` probably would prevent that. Idk, thoughts?

CaseyCarter:
I would be strongly in favor of experimentation on this area. I've been kicking around a semantic wrapper template `nonnegative` that wraps a signed type, `RANGES_EXPECT`s that values assigned to it are non-negative, and `RANGES_ASSUME`s that values retrieved from it are non-negative. The intent is to use such a wrapper where today's standard library uses unsigned types when it really means "non-negative" and not "has mod-2^n behavior."

[...]

PR #593, opened 2017-03-16
add view::indices, deprecate closed_ints

### LINKED ISSUE #277: `ints` and `iota` are not compatible with similar facilities in other languages.

### LINKED ISSUE #470: closed_ints is broken
It seems that `closed_ints(first, last)` maps to `iota_view(first, last + 1)` which is obviously broken if `last + 1` overflows, like here:

``` c++
auto rng = view::closed_ints((std::uint64_t)0, std::numeric_limits<std::uint64_t>::max());
// works (weirdly, because of overflow):
  std::uint64_t last0 = *(ranges::end(rng) - 1);
  CHECK(last0 == std::numeric_limits<std::uint64_t>::max());
// fails (remark: unsigned overflow is defined behavior)
  CHECK(ranges::size(rng) != (unsigned long)0); // size == 0 because of overflow
// fails (as a consequence)
  std::uint64_t last1 = 2;
  for (auto i : rng) { last1 = i; }  // because size is zero, nothing happens here
  CHECK(last1 == std::numeric_limits<std::uint64_t>::max());
```

Is there a reason why closed ints maps to `iota_view` instead of closed_iota [here](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/iota.hpp#L494)? Why isn't `closed_iota` actually used? 

Note: `closed_ints((T)0, (T)max)` actually works for most `T`s in 64bit platforms since a _fast_ integer type is chosen (I only got it to fail for 64bit integers in my platform).

gnzlbg:
- Deprecate `closed_ints`. 
- add `view::indices/inclusive_indices`
- [bugfix]: fix unsigned integer wrap around which caused weird behavior like a range of size of `0` that can be iterated for`ints`/`closed_ints`/`indices`/`inclusive_indices` :
- assert on signed integer overflow by requiring `to - from < std::numeric_limits<T>::max()` for`ints`/`closed_ints`/`indices`/`inclusive_indices`

Closes #470 (at least the "bug" part of it. @CaseyCarter: it might be worth it to move your ideas for infinite ranges to its own issue to pursue them further).

Closes #277 .

Supersedes #592 .

### DIFF
--- a/include/range/v3/view/indices.hpp
+++ b/include/range/v3/view/indices.hpp
@@ -1,0 +1,119 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler 2013-2014
+//  Copyright Gonzalo Brito Gadeschi
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_INDICES_HPP
+#define RANGES_V3_VIEW_INDICES_HPP
+
+#include <meta/meta.hpp>
+#include <range/v3/detail/satisfy_boost_range.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/utility/static_const.hpp>
+#include <range/v3/view/take_exactly.hpp>
+#include <range/v3/view/iota.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+
+        namespace view
+        {
+            /// Half-open range of indices: [from, to).
+            struct indices_fn
+              : iota_view<std::ptrdiff_t>
+            {
+                indices_fn() = default;
+
+                template<typename Val, CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val from, Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    detail::take_exactly_view_<iota_view<Val>, true>
+                        {iota_view<Val>{from}, detail::ints_open_distance_(from, to)}
+                )
+
+                template<typename Val, typename Self = indices_fn,
+                    CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    Self{}(Val(), to)
+                )
+
+            #ifndef RANGES_DOXYGEN_INVOKED
+                template<typename Val, CONCEPT_REQUIRES_(!Integral<Val>())>
+                void operator()(Val) const
+                {
+                    CONCEPT_ASSERT_MSG(Integral<Val>(),
+                        "The object passed to view::indices must be Integral");
+                }
+                template<typename Val, CONCEPT_REQUIRES_(!Integral<Val>())>
+                void operator()(Val, Val) const
+                {
+                    CONCEPT_ASSERT_MSG(Integral<Val>(),
+                        "The object passed to view::indices must be Integral");
+                }
+            #endif
+            };
+
+            /// Inclusive range of indices: [from, to].
+            struct closed_indices_fn
+              : iota_view<std::ptrdiff_t>
+            {
+                template<typename Val, CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val from, Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    detail::take_exactly_view_<iota_view<Val>, true>
+                        {iota_view<Val>{from}, detail::ints_closed_distance_(from, to)}
+                )
+
+                template<typename Val, typename Self = closed_indices_fn,
+                    CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    Self{}(Val(), to)
+                )
+
+            #ifndef RANGES_DOXYGEN_INVOKED
+                template<typename Val, CONCEPT_REQUIRES_(!Integral<Val>())>
+                void operator()(Val) const
+                {
+                    CONCEPT_ASSERT_MSG(Integral<Val>(),
+                        "The object passed to view::closed_indices must be Integral");
+                }
+                template<typename Val, CONCEPT_REQUIRES_(!Integral<Val>())>
+                void operator()(Val, Val) const
+                {
+                    CONCEPT_ASSERT_MSG(Integral<Val>(),
+                        "The object passed to view::closed_indices must be Integral");
+                }
+            #endif
+            };
+
+            /// \relates indices_fn
+            /// \ingroup group-views
+            RANGES_INLINE_VARIABLE(indices_fn, indices)
+
+            /// \relates closed_indices_fn
+            /// \ingroup group-views
+            RANGES_INLINE_VARIABLE(closed_indices_fn, closed_indices)
+
+        }  // namespace view
+    }
+}
+
+#endif  // RANGES_V3_VIEW_INDICES_HPP
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -163,6 +163,55 @@ namespace ranges
                     :  static_cast<D>(v0 - v1);
             }
 
+            template<typename Val, CONCEPT_REQUIRES_(SignedIntegral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_open_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                RANGES_EXPECT(from <= to);
+                static_assert(sizeof(iota_difference_t<Val>) >= sizeof(Val),
+                    "iota_difference_type must be at least as wide as the signed integer type; "
+                    "otherwise the expression below might overflow when to - from would not.");
+                return static_cast<D>(to) - static_cast<D>(from);  
+            }
+
+            template<typename Val, CONCEPT_REQUIRES_(UnsignedIntegral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_open_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                using UD = meta::_t<std::make_unsigned<D>>;
+                // Disable wrap-semantics for UnsignedIntegral types.
+                // See discussion in: https://github.com/ericniebler/range-v3/pull/593
+                RANGES_EXPECT(from <= to);
+                // view::take_exactly(Rng rng, difference_type_t<Rng> n) takes a
+                // number of type difference_type_t<Rng>, which for view::iota is
+                // equal to iota_difference_t<Val>.
+                //
+                // The expression (to - from) shall not overflow this signed
+                // difference type. This is ensured by the RANGES_EXPECT below,
+                // in which:
+                // - the expression is evaluated using the unsigned integer type of the inputs, and
+                // - compared against the maximum value representable by the
+                //   signed difference_type (which is always positive and can
+                //   always be represented by an unsigned integer type of the
+                //   same width).
+                //
+                static_assert(sizeof(D) >= sizeof(Val),
+                    "the difference type of view::iota must be at least as wide as the UnsignedIntegral type");
+                RANGES_EXPECT(Val(to - from) <= static_cast<UD>(std::numeric_limits<D>::max()));
+                return static_cast<D>(Val(to - from));
+            }
+
+            template<typename Val, CONCEPT_REQUIRES_(Integral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_closed_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                auto dist = ints_open_distance_(from, to);
+                // Check whether dist + 1 would overflow the signed integer type,
+                // introducing undefined behavior:
+                RANGES_EXPECT(dist < std::numeric_limits<D>::max());
+                return dist + D(1);
+            }
+
             template<typename Val,
                 typename C = common_type_t<Val, iota_difference_t<Val>>>
             auto iota_plus_(Val const &v, iota_difference_t<Val> n, std::true_type)
@@ -494,11 +543,14 @@ namespace ranges
                 {
                     return iota_view<Val>{value};
                 }
-                template<typename Val>
-                meta::if_c<
-                    (bool)Integral<Val>(),
-                    detail::take_exactly_view_<iota_view<Val>, true>>
-                operator()(Val from, Val to) const;
+                
+                template<typename Val, CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val from, Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    detail::take_exactly_view_<iota_view<Val>, true>
+                        {iota_view<Val>{from}, detail::ints_open_distance_(from, to)}
+                )
 
             #ifndef RANGES_DOXYGEN_INVOKED
                 template<typename Val,
@@ -518,22 +570,14 @@ namespace ranges
             #endif
             };
 
-            template<typename Val>
-            meta::if_c<
-                (bool)Integral<Val>(),
-                detail::take_exactly_view_<iota_view<Val>, true>>
-            ints_fn::operator()(Val from, Val to) const
-            {
-                return {iota_view<Val>{from}, detail::iota_minus_(to, from)};
-            }
-
             struct closed_ints_fn
             {
                 template<typename Val,
                     CONCEPT_REQUIRES_(Integral<Val>())>
+                RANGES_DEPRECATED("view::closed_ints is deprecated; use view::closed_indices instead!")
                 detail::take_exactly_view_<iota_view<Val>, true> operator()(Val from, Val to) const
                 {
-                    return {iota_view<Val>{from}, detail::iota_minus_(to, from) + 1};
+                    return {iota_view<Val>{from}, detail::ints_closed_distance_(from, to)};
                 }
             #ifndef RANGES_DOXYGEN_INVOKED
                 template<typename Val,
@@ -550,8 +594,6 @@ namespace ranges
             /// \ingroup group-views
             RANGES_INLINE_VARIABLE(ints_fn, ints)
 
-            /// \relates closed_ints_fn
-            /// \ingroup group-views
             RANGES_INLINE_VARIABLE(closed_ints_fn, closed_ints)
         }
         /// @}
--- a/test/view/iota.cpp
+++ b/test/view/iota.cpp
@@ -12,6 +12,7 @@
 
 #include <range/v3/core.hpp>
 #include <range/v3/view/iota.hpp>
+#include <range/v3/view/indices.hpp>
 #include <range/v3/view/take.hpp>
 #include <range/v3/view/indirect.hpp>
 #include <range/v3/view/c_str.hpp>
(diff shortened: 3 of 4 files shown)


include/range/v3/view/iota.hpp
```diff
@@ -524,6 +524,7 @@ namespace ranges
                 detail::take_exactly_view_<iota_view<Val>, true>>
             ints_fn::operator()(Val from, Val to) const
             {
+                RANGES_EXPECT(to >= from);
```
CaseyCarter: I disagree that we should require `to >= from`, at least for unsigned types. I don't particularly care about `[0, -2)`, since we already have `view::iota(0)` to name that infinite ranges, but `[(1u << 30), 2u)` is a perfectly well-defined range of unsigned integers.

This could easily be `RANGES_EXPECT(UnsignedIntegral<Val>() || to >= from);`

include/range/v3/view/iota.hpp
```diff
@@ -533,6 +534,8 @@ namespace ranges
                     CONCEPT_REQUIRES_(Integral<Val>())>
                 detail::take_exactly_view_<iota_view<Val>, true> operator()(Val from, Val to) const
                 {
+                    RANGES_EXPECT(to >= from);
```
CaseyCarter: Ditto.

include/range/v3/view/iota.hpp
gnzlbg: @CaseyCarter note that `[(1u << 30), 2u)` is currently broken in range-v3, because `iota_minus_` returns a negative integer for the distance (it performs the substraction correctly), so `take_exactly` will fail when passed a negative integer. . . 

I've made some changes to the PR to fix this, that hopefully do not break any other cases (I've added some new tests for unsigned wrap around as well). 

Please take a careful look at `iota_distance_` and let me know what you think. 

The good think from doing things this way is that we catch with an assertion all cases in which weird behavior would happen. 

The thing that worries me the most is that `view::take_exactly` takes a `difference_type`, which means that for ranges of unsigned integers, a range size of at most the largest representable value by the unsigned integers difference type can be produced. For 32 bit unsigned integers on 32 bit platforms this means that the range shall contain at most `std::numeric_limits<uint32_t>::max() / 2` elements.

include/range/v3/view/iota.hpp
```diff
@@ -533,7 +552,7 @@ namespace ranges
                     CONCEPT_REQUIRES_(Integral<Val>())>
                 detail::take_exactly_view_<iota_view<Val>, true> operator()(Val from, Val to) const
                 {
-                    return {iota_view<Val>{from}, detail::iota_minus_(to, from) + 1};
+                  return {iota_view<Val>{from}, detail::ints_distance_(from, to) + 1};
```
CaseyCarter: s/return/  return/

gnzlbg: fixed

test/view/iota.cpp
```diff
@@ -50,6 +50,22 @@ void test_iota_minus() {
...
+
+    auto rng = view::ints(from, to);
+    CHECK(ranges::size(rng) == std::size_t(5));
+    ::check_equal(rng, {U(max - 2), U(max - 1), max, U(0), U(1)});
+
+    auto rng2 = view::closed_ints(from, to);
+    CHECK(ranges::size(rng2) == std::size_t(6));
+    ::check_equal(rng2, {U(max - 2), U(max - 1), max, U(0), U(1), U(2)});
+}
+
```
ericniebler: This seems like a perversion to me. And now I feel bad because I didn't stop you before spending time on it. :-( I would prefer a narrow interface on `iota` and `ints`^D^D^D^D`indices` that requires `from < to`.

ericniebler:
I don't feel confident making _any_ change to `ints` or `closed_ints` since the semantics of edge cases are weird. And I don't feel motivated to make any changes since we've decided elsewhere to deprecate them. My recommendation is to just leave them alone and hurry up deprecating them.

test/view/iota.cpp
gnzlbg: While this is indeed a perversion, I think this should work for unsigned integers, and that we should assert for signed integers to prevent undefined behavior. 

I am preparing right now a PR that deprecates this and adds `view::indices/closed_indices` and for those and unsigned integers I use this behavior as well. 

Maybe fixing these for the two-argument ints version is not worth it since those who could benefit will almost immediately get a deprecated warning, but the fix is already here so... I also don't see a reason why we shouldn't apply it :/

gnzlbg:
So i've added `view::indices/closed_indices` and deprecated `closed_ints`.

I've preserved the "fixed" wrap around semantics and test for unsigned integers because "why not".

I've made the default `indices/closed_indices` integer type be `std::size_t` because that's the types of indices in the STL. Thoughts on this? Should I remove this behavior from `view::indices` all together?

include/range/v3/view/iota.hpp
```diff
@@ -163,6 +163,25 @@ namespace ranges
                     :  static_cast<D>(v0 - v1);
             }
 
+            template<typename Val, CONCEPT_REQUIRES_(SignedIntegral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                RANGES_EXPECT(from <= to);
+                RANGES_EXPECT((static_cast<D>(to) - static_cast<D>(from))
+                              < std::numeric_limits<D>::max());
```
CaseyCarter: What does this expectation accomplish? If `to - from == max`, it's representable. `to - from <= max` always holds.

CaseyCarter: If it's an attempt to guard against `to - from` overflowing `D`, which is commendable, it's not working.

include/range/v3/view/iota.hpp
```diff
@@ -163,6 +163,25 @@ namespace ranges
...
+                return static_cast<D>(to) - static_cast<D>(from);
+            }
+
+            template<typename Val, CONCEPT_REQUIRES_(UnsignedIntegral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                RANGES_EXPECT(static_cast<D>(to - from) < std::numeric_limits<D>::max());
```
CaseyCarter: Ditto.

test/view/iota.cpp
```diff
@@ -34,20 +35,40 @@ struct Int
...
+    CHECK(detail::iota_minus_(I(max - I(1)), I(max - I(1))) == D(0));
+    CHECK(detail::iota_minus_(I(max - I(1)), I(max)) == D(-1));
+    CHECK(detail::iota_minus_(I(max), I(max - I(1))) == D(1));
+    CHECK(detail::iota_minus_(I(max), I(max)) == D(0));
+}
+
+template <typename U>
+void test_wrap_around() {
+    using namespace ranges;
+    auto max = std::numeric_limits<U>::max();
```
CaseyCarter: I would get rid of `max`, and use `U(-1)`, `U(-2)`, and `U(-3)` in place of `max`, `U(max - 1)`, and `U(max - 2)`. This function then behaves identically for both signed and unsigned types.

include/range/v3/view/indices.hpp
```diff
@@ -0,0 +1,123 @@
...
+            /// Half-open range of indices: [from, to).
+            struct indices_fn
+              : iota_view<std::size_t>
+            {
+                indices_fn() = default;
+
+                template<typename Val>
+                meta::if_c<
+                    (bool)Integral<Val>(),
+                    detail::take_exactly_view_<iota_view<Val>, true>>
```
CaseyCarter: Why is this not `take_exactly_view<iota_view<Val>>`? (and on 80)

include/range/v3/view/indices.hpp
```diff
@@ -0,0 +1,123 @@
...
+        namespace view
+        {
+            /// Half-open range of indices: [from, to).
+            struct indices_fn
+              : iota_view<std::size_t>
+            {
+                indices_fn() = default;
+
+                template<typename Val>
+                meta::if_c<
```
CaseyCarter: Why is the style of constraint different for this function and the function on line 50?

include/range/v3/view/iota.hpp
```diff
@@ -524,16 +543,17 @@ namespace ranges
...
 
             struct closed_ints_fn
             {
                 template<typename Val,
                     CONCEPT_REQUIRES_(Integral<Val>())>
+                RANGES_DEPRECATED("view::closed_ints is deprecated; use view::closed_indices instead!")
                 detail::take_exactly_view_<iota_view<Val>, true> operator()(Val from, Val to) const
                 {
-                    return {iota_view<Val>{from}, detail::iota_minus_(to, from) + 1};
+                    return {iota_view<Val>{from}, detail::ints_distance_(from, to) + 1};
```
CaseyCarter: Is this why `ints_distance_` forbids `to - from == max`? If so, that precondition should be here instead of there.

test/view/iota.cpp
```diff
@@ -76,20 +97,20 @@ int main()
     models<concepts::BoundedView>(aux::copy(shorts));
     static_assert(Same<int, range_difference_type_t<decltype(shorts)>>(), "");
 
-    auto uints = view::closed_ints(
+    auto uints = view::inclusive_indices(
         std::numeric_limits<std::uint32_t>::min(),
```
CaseyCarter: s/(u)?int([0-9]+)_t/\1int_least\2_t/g

test/view/iota.cpp
```diff
@@ -76,20 +97,20 @@ int main()
...
-        std::numeric_limits<std::uint32_t>::max());
+        std::numeric_limits<std::uint32_t>::max() - 1);
     models<concepts::BoundedView>(aux::copy(uints));
     static_assert(Same<std::int64_t, range_difference_type_t<decltype(uints)>>(), "");
```
CaseyCarter: s/int64_t/intmax_t/

test/view/iota.cpp
```diff
@@ -76,20 +97,20 @@ int main()
...
     static_assert(Same<std::uint64_t, range_size_type_t<decltype(uints)>>(), "");
```
CaseyCarter: s/uint64_t/uintmax_t/

include/range/v3/view/indices.hpp
```diff
@@ -0,0 +1,123 @@
...
+namespace ranges
+{
+    inline namespace v3
+    {
+
+        namespace view
+        {
+            /// Half-open range of indices: [from, to).
+            struct indices_fn
+              : iota_view<std::size_t>
```
gnzlbg: If we are going to use `difference_type` for indexing in the STL2, we should use `std::ptrdiff_t` here.

include/range/v3/view/indices.hpp
gnzlbg: @ericniebler ?

I just took this from `view::ints`, and since I did not know _why_ they were constrained that way instead of with `CONCEPT_REQUIRES_` I decided not to change it, but I see no reason against using `CONCEPT_REQUIRES_` here.

include/range/v3/view/iota.hpp
gnzlbg: See my comment above. While this check is probably wrong, `to - from < std::numeric_limits<range_size_t>::max()` or otherwise computing the size will overflow and give `0`, so I think fixing the check here makes sense. Otherwise we end up with ranges of size `0` that are iterable ...

include/range/v3/view/iota.hpp
gnzlbg: So this check uses `<` instead of `<=` to catch when the range size which is `(to - from) + 1` for closed ranges would overflow. That is, if `to - from == max` everything is still fine, but when one attempts to compute the size that will overflow for _unsigned integers_.

However, we are dealing with signed integers here which have an unsigned range size that is at least as wide, so doing that should never overflow, so this check is moot and can be removed. 

> If it's an attempt to guard against to - from overflowing D, which is commendable, it's not working.

I don't think that can ever work. If `to - from` overflows its UB, I don't see a way to detect this.

include/range/v3/view/iota.hpp
gnzlbg: Yep. I remember now why I used difference type, it's because `take_exactly` takes a difference type, so that one is the one in which overflow must be prevented. 

I think I should write a comment because I did it 2 days ago and already forgot why. . . sigh.

test/view/iota.cpp
gnzlbg: I am only using this function below to test the unsigned types and thought `max` was more clear than `U(-1)`.

include/range/v3/view/indices.hpp
ericniebler: > Why is the style of constraint different for this function and the function on line 50?
  
I dunno. :-|

include/range/v3/view/indices.hpp
ericniebler: I dunno.

include/range/v3/view/indices.hpp
ericniebler: Oh, I remember. I needed to fall back on `enable_if`-style constraints when I needed the ability to forward-declare a constrained function.

ericniebler:
> I've preserved the "fixed" wrap around semantics and test for unsigned integers because "why not".

Why not: Because users are far, far, *far* more likely to accidentally get wrap-around behavior in error than to actually want unsigned wrap-around. I would much prefer to trap these bugs than to enable a use case of highly dubious value.

test/view/iota.cpp
CaseyCarter: The fact that `[U(-3), U(2))` is a valid range of exactly 5 elements whether or not `U` is a signed or unsigned integral type is the primary motivation for not requiring `to < from` for ranges of unsigned types.

ericniebler: https://github.com/ericniebler/range-v3/pull/593#issuecomment-287879747

include/range/v3/view/indices.hpp
CaseyCarter: > That said, I'm a bit uncomfortable with drifting from unsigned to signed here and there, in dribs and drabs.

I have an eventual goal in mind. I've got a rough branch that makes `ranges::size` always return a signed type, and changes all of the `size_type`s of the views to be signed types. I've got the beginnings of a semantic wrapper `nonnegative<T>` that "knows" that the stored `T` value is not negative. What I haven't done yet is wrap all of that up nicely to see how it works together and interacts with the standard library.

I think your "Dribs and drabs" argument has sold me. Changing this component to use a signed type as well is not a significant burden, and I'd much rather present a holistic design than try to explain why a single component is trying to swim against the current.

gnzlbg: I've changed this to use `std::ptrdiff_t` instead of `std::size_t` as default integer type.

include/range/v3/view/indices.hpp
gnzlbg: @CaseyCarter @ericniebler because of [this](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/take_exactly.hpp#L43), and [STL2 issue 63](https://github.com/ericniebler/stl2/issues/63).

include/range/v3/view/indices.hpp
gnzlbg: I've switched this to `CONCEPT_REQUIRES_` for view::indices/closed_indices.

include/range/v3/view/iota.hpp
gnzlbg: So I've removed the check in the signed version, and hopefully fixed this check.

test/view/iota.cpp
gnzlbg: fixed

test/view/iota.cpp
gnzlbg: fixed

gnzlbg:
@ericniebler I agree that this is often a bug. The only concern I have is: "what about users that do want wrap around behavior"? While I don't think that we should provide a solution for this right now, I do think that we should at least discuss possible paths forward to supporting those, and try to make sure that those can be pursued by whoever is interested without breaking backwards compatibility (within reasonable margins, range-v3 is not the C++ standard). 

Here is my mental sketch of how somebody interested _could_ extend `view::indices` to support wrap around behavior. 

- 0. `view::indices` currently only supports native integral types (`std::is_integral<T>`), and does not allow wrap around (as of this commit). This would be the starting point after this PR.
- 1. Fix part of #591. That is, add an `IntegralNumber` concept that allows "user-defined" integer types with same complexities and semantics as `std::is_integral" native types. 
- 2. Add an user-defined `wrapping<T>` library type that models `IntegralNumber` if T models `IntegralNumber` and provides wrapping integer arithmetic independently of signed-ness.
- 3. Extend in a backwards-compatible way `view::indices`/`ints` to `IntegralNumber` types, that use whatever wrap-around behavior user-defined integral numbers provide. This will require dancing around `std::is_integral` types also modeling `IntegralNumber`, but is doable in a backwards compatible way. 

That is, those wanting wrap around behavior would just need to pass their wrapping types to `view::indices` or write `view::indices(wrapping(a), wrapping(b))`.

include/range/v3/view/iota.hpp
```diff
@@ -163,6 +163,47 @@ namespace ranges
...
+                    "iota_difference_type must be at least as wide as the signed integer type; "
+                    "otherwise the expression below might overflow when to - from would not.");
+                return static_cast<D>(to) - static_cast<D>(from);  
+            }
+
+            template<typename Val, CONCEPT_REQUIRES_(UnsignedIntegral<Val>())>
+            RANGES_CXX14_CONSTEXPR
+            iota_difference_t<Val> ints_distance_(Val from, Val to) noexcept {
+                using D = iota_difference_t<Val>;
+                using UD = std::make_unsigned_t<D>;
```
ericniebler:
```
/Users/travis/build/ericniebler/range-v3/include/range/v3/view/iota.hpp:181:33: error: no template named 'make_unsigned_t' in namespace 'std'; did you mean 'make_unsigned'?
                using UD = std::make_unsigned_t<D>;
                           ~~~~~^~~~~~~~~~~~~~~
                                make_unsigned
```

include/range/v3/view/indices.hpp
```diff
@@ -0,0 +1,117 @@
...
+                (
+                    detail::take_exactly_view_<iota_view<Val>, true>
+                        {iota_view<Val>{from}, detail::ints_distance_(from, to)}
+                )
+
+                template<typename Val, CONCEPT_REQUIRES_(Integral<Val>())>
+                auto operator()(Val to) const
+                RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+                (
+                    (*this)(Val(), to)
```
ericniebler: Older versions of gcc will not suffer to have `this` appear in trailing return types.

include/range/v3/view/indices.hpp
```diff
@@ -0,0 +1,117 @@
...
+            RANGES_INLINE_VARIABLE(indices_fn, indices)
+
+            /// \relates inclusive_indices_fn
+            /// \ingroup group-views
+            RANGES_INLINE_VARIABLE(inclusive_indices_fn, inclusive_indices)
+
+        }  // namespace view
+    }
+}
+
```
ericniebler: You need to do the `SATISFY_BOOST_RANGE` dance.

include/range/v3/view/iota.hpp
ericniebler: If we're always doing `n+1` in `closed_ints` and `inclusive_indices`, then the need for the view goes away. Users can do `+1` themselves. What they *can't* do it get a closed range that covers the entire range of integers without fear of overflow.

If we decide that's not a valid thing to want, I suggest we deprecate `closed_ints` and `inclusive_indices`. (And I disapprove of the inconsistent naming, FWIW.) If it *is* a valid thing to want, then this PR is wrong.

include/range/v3/view/indices.hpp
CaseyCarter: [Looking in range_fwd.hpp](https://github.com/CaseyCarter/range-v3/blob/master/include/range/v3/range_fwd.hpp#L658), `take_exactly_view<iota_view<Val>>` is still defined as `detail::take_exactly_view_<iota_view<Val>, is_random_access_bounded_<Rng>::value>>` and [`is_random_access_bounded_` is still `RandomAccess` and `BoundedRange`](https://github.com/CaseyCarter/range-v3/blob/master/include/range/v3/view/take_exactly.hpp#L39).

Until `take_exactly_view` changes, there's no reason to break the abstraction here.

CaseyCarter:
> I agree that this is often a bug. The only concern I have is: "what about users that do want wrap around behavior"? 

In a future where people use unsigned integers because they *want* modulo behavior, and not because they want "nonnegative," we would re-enable the proper behavior here.

include/range/v3/view/indices.hpp
gnzlbg: IIUC `indices` is not a "view", it just returns `iota | take_exactly` and `iota` already does this, so that should be enough right?

include/range/v3/view/iota.hpp
gnzlbg: fixed

include/range/v3/view/iota.hpp
gnzlbg: @ericniebler I've split `ints_distance_` into `ints_open_distance_` and `ints_closed_distance`. 

Its not about "jus" doing `+1`, but rather, about making sure that this `+1` doesn't overflow the signed difference_type that `view::take_exactly` takes as argument. 

So for open ranges, one checks that:
- `from <= to` (for both signed and unsigned integers),
- `Unsigned(to - from) <= Unsigned(numeric_limits<difference_type>::max())` for unsigned integers only, otherwise overflowing the signed difference type is undefined behavior.

For closed ranges one has the same checks as for open ranges, plus:
- `open_distance < numeric_limits<difference_type>::max()` (for both)
since otherwise `open_distance + difference_type(1) will overflow, introducing undefined behavior. 

@CaseyCarter, does this makes sense? (or more sense as before?).

> What they can't do is get a closed range that covers the entire range of integers without fear of overflow.

For unsigned integers they can only cover a sub-range of (half - 1) size. I think its worth it to assert on this to prevent undefined behavior. 

> And I disapprove of the inconsistent naming, FWIW.

I've renamed `inclusive_indices` to `closed_indices` "for consistency", but I think that `inclusive` is a better name, and we are deprecating `closed_ints` anyways, so if we remove that at some point, we will be left with `closed_indices` only. I'd rather not make a breaking change there again to rename it to `inclusive_indices` if we happen to like that better. Idk, what do you guys like more? Are there better names?

include/range/v3/view/indices.hpp
gnzlbg: IIRC this was actually a clang bug right? This shouldn't be allowed. I'll fix it.

gnzlbg: Fixed and rebased.

gnzlbg:
It seems that these are not true on MacOS :'( whyyyyyy!

```shell
/Users/travis/build/gnzlbg/range-v3/test/view/iota.cpp:122:5: error: static_assert failed ""
    static_assert(Same<std::intmax_t, range_difference_type_t<decltype(uints)>>(), "");
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/gnzlbg/range-v3/test/view/iota.cpp:123:5: error: static_assert failed ""
    static_assert(Same<std::uintmax_t, range_size_type_t<decltype(uints)>>(), "");
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/gnzlbg/range-v3/test/view/iota.cpp:129:5: error: static_assert failed ""
    static_assert(Same<std::intmax_t, range_difference_type_t<decltype(ints)>>(), "");
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/gnzlbg/range-v3/test/view/iota.cpp:130:5: error: static_assert failed ""
    static_assert(Same<std::uintmax_t, range_size_type_t<decltype(ints)>>(), "");
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4 errors generated.
```

gnzlbg:
So, on MacOSX:

- `std::uint64_t == unsigned long long`
- `std::uintmax_t == unsigned long`
- `std::size_t == unsigned long`

and 

- `std::int64_t == long long`
- `std::intmax_t ==  long`
- `std::ptrdiff_t ==  long`

For `view::closed_ints`, however:

- `range_size_type_t<decltype(ints)> == unsigned long long`
- `range_difference_type_t<decltype(ints)> == long long`

because the difference type is based on `int_fastXX_t`, which on MacOSX is:

- `std::int_fast32_t == int`.
- `std::int_fast64_t == long long`. 

So... from `int_fast64_t` one gets `long long` and `unsigned long long` as difference and size types for `view::iota`, but `std::intmax_t` and `std::uintmax_t` are `long` and `unsigned long`. 

I think that the right thing to do is to replace `std::intmax_t` and `std::uintmax_t` with `std::int_fastXX_t` since that is actually what `view::iota` is doing.

CaseyCarter:
> So, on MacOSX:
> ```c++
> std::uint64_t == unsigned long long
> std::uintmax_t == unsigned long
> std::size_t == unsigned long
> ```

It is insane - albeit conforming - that `intmax_t` has lower integer conversion rank than `long long`, even if `long` and `long long` do have the same object representation. I agree with your solution.

ericniebler:
This one looks ready to go. I'll merge soon unless I hear shouts.

gnzlbg:
@ericniebler there is a travis failure in the boost build bot, but i think this has always been the case (it surfaces now because clang 5.0 was broken for a week). For some reason Boost is failing to link, and I've debugged enough to know that it is not modules related (i can reproduce it without modules enabled on that build bot). I've fixed some parts of it in the cmake refactor PR, and plan to fully fix it there. That PR should deliver a green build on everything.

PR #594, opened 2017-03-16
Updated the definition of OutputIterator to satisfy output iterators that have a void result type for '*it++ = x'.

tahonermann:


ericniebler:
Nice catch!

PR #595, opened 2017-03-17
Fix assertion in view::cycle

FloopCZ:
Fix constructor in view::cycle, which uses a moved-from variable


gnzlbg:
LGTM.

ericniebler:
Yikes! Thanks.

PR #599, opened 2017-03-17
[docs] Document actions in front-page.

gnzlbg:
This adds a list of the available actions to the front-page.


ericniebler:
Looks like `erase`, `insert`, `push_front`, `push_back`, and `unique` are missing links to the doxygen-generated docs. Why?

gnzlbg:
I can't link to the inline variables via doxygen, and the `_fn` structs are
in an `adl`-barrier namespace that is ignored by doxygen so I cannot link
to those either. So... I decided to left them unlinked to anything for now,
not that the inline variables / _fn struct offer any extra documentation
anyways.

A temporary solution could be to link to the files directly? I don't know
if doxygen can do that but it might work.

PR #600, opened 2017-03-17
[cmake/travis] refactor/improve CMake; add MSan/LSan builds in travis

### LINKED ISSUE #333: view.take_while: runtime error: member call on null pointer of type 'struct __lambda8'
We should investigate [this ASAN message](https://travis-ci.org/ericniebler/range-v3/jobs/124542559#L5024) from Travis:

<pre>
182: Test command: /home/travis/build/ericniebler/range-v3/build/test/view/view.take_while
182: Test timeout computed to be: 1500
182: /home/travis/build/ericniebler/range-v3/test/view/take_while.cpp:56:76: runtime error: member call on null pointer of type 'struct __lambda8'
182: /home/travis/build/ericniebler/range-v3/test/view/take_while.cpp:56:76: runtime error: member call on null pointer of type 'struct __lambda8'
</pre>

### LINKED ISSUE #451: Build warnings

### LINKED ISSUE #640: travis-ci tests older clangs with new libc++
The `install_libcxx.sh` script that travis-ci uses to install libc++ always fetches tip-of-trunk libc++ sources and builds from that. That's regardless of what version of clang is being used. Trouble is, that's not how our users are likely to do it. Clang-3.6 (for instance) was released with a specific version of libc++, and testing with any other version is likely to either hide real problems or turn up spurious ones.

Find some way to install a version of libc++ that is appropriate for the version of clang that is being used.

ATTN: @gnzlbg

gnzlbg:
- Adds MemorySanitizer and LeakSanitizer (under ASan) in travis:
  - Builds libc++ with MemorySanitizer and AddressSanitizer as required.
  - Enables the leak sanitizer on AddressSanitizer builds under Linux.
  - Enables the unsigned-integer overflow check in the
    UndefinedBehaviorSanitizer:
    - Avoids overflow in view.unique tests (not a bug).
    - Add RANGES_INTENDED_MODULO_ARITHMETIC macro to explicitly annotate
      when modulo arithmetic is intended.
    - Explicitly annotates those parts of range-v3 in which modulo arithmetic is
      intended (mostly in utility/random.hpp).

- Refactor the root `CMakeLists.txt` into:
  - `cmake/readme.md`: brief explanation of what goes where.
  - `cmake/ranges_options.cmake`: library configuration options.
  - `cmake/ranges_env.cmake`: detects environment (os, compiler, build type...).
  - `cmake/ranges_flags.cmake`: sets up supported compiler flags.

- AppVeyor and Clang/C2 (by @CaseyCarter):
  - Add cmake to PATH in appveyor.yml.
  - Disable -fstruct-vtable-pointers on Clang/C2.

- Add `RANGES_NATIVE` option (defaults to `On`) to allow
  disabling `-march/mtune=native` independently of the build-type.

- Disables std::auto_ptr in boost (not available in C++1z).

- Add travis build bots for clang-4.0 in C++11 and C++14.

- Tests clang versions with the corresponding libc++ version (they were always
  tested with libc++ trunk).

Closes #333: the ASAN errors reported there cannot be reproduced anymore.
Closes #451: -Wnoexcept should be part of -Weverything.
Closes #640: clang versions are tested with appropriate libc++ version.

Thanks @CaseyCarter for the comments, help, and all the Clang/C2 and
AppVeyor fixes.


gnzlbg:
Do not merge this yet, it enables MSan and LSan in some travis jobs. I am just submitting this PR already so that we can all see what fails where and send separate PRs to fix those issues _before_ we merge this PR.

cmake/ranges_flags.cmake
```diff
@@ -0,0 +1,173 @@
...
+    # frame pointer, so leave it enabled when compiling with either of them:
+    ranges_append_flag(RANGES_HAS_OMIT_FRAME_POINTER -fomit-frame-pointer)
+  endif()
+  
+  ranges_append_flag(RANGES_HAS_MARCH_NATIVE "-march=native")
+  ranges_append_flag(RANGES_HAS_MTUNE_NATIVE "-mtune=native")
+  ranges_append_flag(RANGES_HAS_OFAST -Ofast)
+
+  ranges_append_flag(RANGES_HAS_STRICT_ALIASING -fstrict-aliasing)
+  ranges_append_flag(RANGES_HAS_STRICT_VTABKE_POINTERS -fstrict-vtable-pointers)
```
CaseyCarter: `VTABKE`

CaseyCarter:
> * enables the unsigned-integer overflow check in the UndefinedBehaviorSanitizer
>   * this requires explicitly annotating in the code when unsigned integer overflow is actually intended.

This is fundamentally broken: unsigned integers do not "overflow" in C++.

gnzlbg:
> This is fundamentally broken: unsigned integers do not "overflow" in C++.

Sure, but when somewhere in any of the codebases I work on an unsigned integer overflows, 99% of the time it is a bug. The other 1% of the time it doesn't hurt that much to say: "this operation is intended to overflow". Supporting this in libc++ [took 3 explicit annotations](https://github.com/llvm-mirror/libcxx/search?utf8=%E2%9C%93&q=DISABLE+UBSAN+INTEGER&type=Code), I don't expect it to be more complicated than that in range-v3.  I don't recall whether it found some cases in C++ were unsigned integers were overflowing by accident though. I don't think so but @EricWF might remember more about how hard it was to change this.

CaseyCarter:
> Sure, but when somewhere in any of the codebases I work on an unsigned integer overflows, 99% of the time it is a bug. The other 1% of the time it doesn't hurt that much to say: "this operation is intended to overflow".

It doesn't makes sense to say "this operation is intended to overflow" for a type that does not have operations that overflow. The bug is choosing an unsigned type to represent a non-negative value and then being surprised that the type has modulo behavior. I am not convinced that we do the community a service by continuing to perpetuate this fallacy.

That said, I realize that there are billions of lines of code that are suffering the consequences of this mistake, and that range-v3 has to interoperate with that code. I'll get off my soapbox and save the rant for STL2.

gnzlbg:
I have one failure in the MSan build that's taking too long because I forgot to disable valgrind for MSan runs. I'll fix that tomorrow along with the other issues.

---

@CaseyCarter 

> The bug is choosing an unsigned type to represent a non-negative value and then being surprised that the type has modulo behavior. 

Even when working only with signed integers, an innocent call to `.size()` gives birth to an unsigned integer in an signed world, in which implicit conversions can silently change the arithmetic of complex expressions into pure chaos. 

UBSan will point you to every expression in which this happens if you are willing to open your heart :D 

Kidding aside, I have to admit that I love this check. This check has found more bugs at work than all checks of all the sanitizers together. Ironically, it is the only check in the UndefinedBehaviorSanitizer that does not check for undefined behavior.  

It also lets you implement something like Rust's `wrapping<T>` in C++. In Rust, "integer arithmetic is orthogonal to representable values", so all native integers trap on overflow/underflow. If you want wrapping or saturating arithmetic, you _need_ to use a `Wrapping<T>` or `Saturating<T>` type that makes it explicit. I wish we could have something like that in the STL2, so that `std::size_t` could still be unsigned (as in positive numbers only), but `Trapping<T>`, as in, you get an assertion when it overflows.

ericniebler:
> an innocent call to `.size()` gives birth to an unsigned integer in an signed world 

The fact that `.size()` typically returns an unsigned integer is a design mistake that (I hope) STL2 will correct.

> If you want wrapping or saturating arithmetic, you need to use a `Wrapping<T>` or `Saturating<T>` type that makes it explicit. I wish we could have something like that in the STL2, so that `std::size_t` could still be unsigned (as in positive numbers only), but `Trapping<T>`, as in, you get an assertion when it overflows.

That said, I like the idea of making STL2 work with integral-like class types that are runtime-checked for things like overflow and underflow, and that disallow narrowing conversions. That is a good argument for an `IntegralLike` concept.

cmake/ranges_flags.cmake
gnzlbg: fixed, I've also enabled the nullability checks, and make all checks terminate the program by default so that we see the builds failing on travis.

gnzlbg:
It seems that travis is having some issues with clang 5.0, all builds for it on all PRs are broken :/ 
(maybe restarting them today/tomorrow magically fixes them)

About the appveyor fail:

```shell
virtual __thiscall std::exception::~exception(void)': Intrinsic not yet implemented:
    %4 = call i8* @llvm.invariant.group.barrier(i8* %3), !dbg !960
```

It seems that I broke this, but IIUC the only piece of MSVC functionality in the old cmake file is:

```cmake
    set(CMAKE_CXX_STANDARD 14)
    set(RANGES_CXX_STD 14)
```

CaseyCarter:
> It seems that travis is having some issues with clang 5.0, all builds for it on all PRs are broken :/

Yeah, llvm.org/apt has become transiently flaky again.

> About the appveyor fail

I've gotten similar crazy errors like this in the past by passing perfectly reasonable clang options to clang/c2 that clang/c2 doesn't "officially" support / know how to translate into the C2 IL representation. (Way back when, "-Ofast" would blow up the compiler.) I strongly suspect the new cmake setup is passing such an option.

gnzlbg:
> I strongly suspect the new cmake setup is passing such an option.

Thanks @CaseyCarter , I'll start looking into this. Maybe I am lucky and the culprit is one of the few `set(CMAKE_CXX_FLAGS ...)` options. All other options are only passed to the compiler if it doesn't fail to compile a trivial example program in the new setup, but... the error seems to be like something that the trivial example that CMake uses might not trigger... In that case its going to be harder to track it down :(

CaseyCarter:
> Maybe I am lucky and the culprit is one of the few set(CMAKE_CXX_FLAGS ...) options.

*sigh* It's apparently **still** `-Ofast`, which turns on generation of LLVM intrinsics that Clang/C2 doesn't yet support.

gnzlbg:
@CaseyCarter Oh ! I'll add a case for Clang/C2 and release builds to turn it into O3. I've fixed some minor bugs with `-ggdb` on linux and it seems that the MSAN check is failing (and thus not enabled on travis) because CMake passes the "examples" only the CMAKE_CXX_FLAGS but not the CMAKE_EXE_LINKER_FLAGS and clang fails to compile with msan enabled without the proper linker flags :/

CaseyCarter:
>  I'll add a case for Clang/C2 and release builds to turn it into O3.

See https://github.com/gnzlbg/range-v3/pull/2.

gnzlbg:
So MSan runs fine with the last commit on Linux passing the following compiler and linker flags to cmake:

```cmake
-DCMAKE_CXX_COMPILER_FLAGS=" -fsanitize=memory -isystem $LIBCXX_MSAN_INCLUDE " \
-DCMAKE_EXE_LINKER_FLAGS=" -stdlib=libc++ -L$LIBCXX_MSAN_LIB -Wl,-rpath,$LIBCXX_MSAN_LIB -lc++abi "
```

I get some test fails due to "read from uninitialized-memory" errors, don't know if they are spurious or not: 

> 89% tests passed, 21 tests failed out of 196

ericniebler:
> I get some test fails due to "read from uninitialized-memory" errors, don't know if they are spurious or not:

MSan and ASan have very low false positive rates. I would be surprised if they are spurious. Can you post a sample backtrace?

gnzlbg:
This branch also fixes a bug in the current CMakeFile.txt, that passes -march=native/mtune=native to the compiler even in debug builds. The fix is to pass these to the compiler only in release builds.

P.S.: I'll start posting the MSan issues later today. They look all to be different issues, and some of them are quite long. If one cannot really debug them I don't know how much will the traces help, and if one can debug them, one can get the traces by just letting the tests run :/

CaseyCarter:
> This branch also fixes a bug in the current CMakeFile.txt, that passes -march=native/mtune=native to the compiler even in debug builds. The fix is to pass these to the compiler only in release builds.

That's not a bug; it's a feature. Instruction selection should not vary between debug and release.

gnzlbg:
Then it should be conditional I guess because `-march=native -mtune=native` also means that the binaries are not portable between CPUs of the same architecture.

CaseyCarter:
I cannot imagine why I would want portability for unit test binaries, but I'll trust that you have a good reason for doing so. I wouldn't be opposed to having `-march` off by default, or both off by default, or even to removing them entirely from the CMakeLists. I'm only opposed to injecting differences between debug builds and release builds.

gnzlbg:
> I'm only opposed to injecting differences between debug builds and release builds.

That's a good point, I'll add an option `RANGES_NATIVE` to enable these flags, and will enable them in some of the build bots on travis on clang and gcc in both debug and release (probably for the 1z builds only, and for the sanitizer builds since this flags enable some times some vectorization that triggers alignment issues that the sanitizers do catch).

EricWF:
>  Supporting this in libc++ took 3 explicit annotations, I don't expect it to be more complicated than that in range-v3. @EricWF might remember more about how hard it was to change this.

Libc++ isn't completely clean yet. There are still a couple places we have unannotated unsigned integer overflow. However finding these instances and annotating them was pretty easy. It also made libc++'s code clearer by explicitly calling out that the overflow is expected.

gnzlbg:
@CaseyCarter I've added a RANGES_NATIVE option (defaults to `On`, so the behavior doesn't change), that allows disabling (or enabling) it for all builds.

include/range/v3/detail/config.hpp
```diff
@@ -389,6 +389,14 @@ namespace ranges
 #define RANGES_RE_ENABLE_WARNINGS
 #endif
 
+
+#if __has_attribute(no_sanitize)
+#define RANGES_INTENDED_UNSIGNED_OVERFLOW \
```
CaseyCarter: I'm strongly opposed to any use of the word "overflow" to describe the behavior of unsigned integers. Framing the problem incorrectly creates confusion. I suggest naming this `RANGES_REALLY_UNSIGNED`.

EricWF: The pedant in me needs to point out that the word overflow describes exactly what happens. Exceeding the maximum representable value is an overflow. The difference is whether the overflow has defined behavior or not.

Just my two cents.

cmake/ranges_flags.cmake
```diff
@@ -0,0 +1,184 @@
...
+endif()
+
+if(RANGES_CXX_COMPILER_GCC AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.0"))
+  ranges_append_flag(RANGES_HAS_WNO_MISSING_FIELD_INITIALIZERS -Wno-missing-field-initializers)
+endif()
+
+
+# Workaround missing __extern_always_inline in clang under linux
+if(RANGES_ENV_LINUX AND RANGES_CXX_COMPILER_CLANG)
+  ranges_append_flag(RANGES_HAS___EXTERN_ALWAYS_INLINE -D__extern_always_inline=inline)
```
EricWF: This seems suspicious... Can you clarify what's going on here?

cmake/ranges_flags.cmake
```diff
@@ -0,0 +1,184 @@
...
+# Template diagnostic flags
+ranges_append_flag(RANGES_HAS_FDIAGNOSTIC_SHOW_TEMPLATE_TREE -fdiagnostics-show-template-tree)
+ranges_append_flag(RANGES_HAS_FTEMPLATE_BACKTRACE_LIMIT "-ftemplate-backtrace-limit=0")
+
+# Clang modules support
+if (RANGES_MODULES)
+  ranges_append_flag(RANGES_HAS_MODULES -fmodules)
+  ranges_append_flag(RANGES_HAS_MODULE_MAP_FILE "-fmodule-map-file=${PROJECT_SOURCE_DIR}/include/module.modulemap")
+  ranges_append_flag(RANGES_HAS_MODULE_CACHE_PATH "-fmodules-cache-path=${PROJECT_BINARY_DIR}/module.cache")
+  if (RANGES_LIBCXX_MODULE)
```
EricWF: I think that you want to use both implicit module maps and `-Xclang -fmodules-local-submodule-visibility` when not on Apple. Have you tested that configuration?

include/range/v3/detail/config.hpp
CaseyCarter: The pedant in you is wrong: unsigned types in C++ are closed under all operations; there are no operations that exceed the maximum representable value. Again, this is fundamentally what modulo behavior means. We are tagging the functions that actually want modulo behavior for unsigned integers, and leaving untagged those functions that wrongly use unsigned types to mean nonnegative.

EricWF: _@CaseyCarter critiques @EricWF's pedanticness. It's super effective._

From the wiki for [Integer Overflow](https://en.wikipedia.org/wiki/Integer_overflow):
> In computer programming, an integer overflow occurs when an arithmetic operation attempts to create a numeric value that is outside of the range that can be represented with a given number of bits.
> ...
> The most common result of an overflow is that the least significant representable bits of the result are stored the result is said to wrap around the maximum (i.e. modulo power of two).

It would appear that commonly accepted definitions for Integer overflow consider the modular arithmetic case. But at this point we're just arguing semantics, and I agree with your overall goal of further separating the lingo for well-defined overflow and undefined overflow.

gnzlbg: @CaseyCarter What about `RANGES_INTENDED_WRAP_AROUND` ? (or modulo 2 arithmetic?)

cmake/ranges_flags.cmake
gnzlbg: IIRC some gcc 4.9 versions of libstdc++ (or libc?) use an annotation `__extern_always_inline` that is not supported by clang (or older clang versions). So... we just replace it with `inline` and cross fingers.

EricWF: That sounds familiar. It would be really nice to document the newest Clang version that triggers this, because I think it's getting close to being no longer relevant.

cmake/ranges_flags.cmake
gnzlbg: IIUC `implicit-module-maps` are enabled by default always, and below I add a `-fno-implicit-module-maps` on MacOSX. I've tested this on linux without `-Xclang -fmodules-local-submodule-visibility` and "it works for me", but I will try to see if adding it changes anything, and if it doesn't, I can add it by default.

cmake/ranges_flags.cmake
gnzlbg: I can remove it and do a travis run, it should tell us if any clang version [3.6,5.0] works fine without this.

cmake/ranges_flags.cmake
EricWF: I don't understand the `-fno-implicit-module-maps` os Mac OS X. I'm assuming the module maps in the OS X SDK give you trouble? If not I would suggest taking advantage of the provided libc and system library module maps. 

PS. Clang and LLVM all require `-fmodules-local-submodule-visibility` on Linux in order to work. Adding it for ranges-v3 should probably be the prefered configuration.

include/range/v3/detail/config.hpp
CaseyCarter: > But at this point we're just arguing semantics

Yes, although I would argue that `UINT_MAX + 1` does not "attempt to create a numeric value that is outside the range that can be represented;" `UINT_MAX + 1` "attempts to create" `0` and succeeds ;)

I apologize for nitpicking at semantics here, but I think framing the problem correctly helps people to better think about how to fix it. The problem is not that unsigned integers can wrap, despite the "unsigned overflow" frame chosen by ubsan. The problem is that a great many C++ users - including the designers of the standard library - have repeatedly chosen to use unsigned types to mean non-negative as if the inability to represent negative numbers is the primary difference between signed and unsigned integers and not the fact that unsigned integers have modulo behavior.

By framing the problem incorrectly, we encourage people to continue using unsigned integers to represent non-negative quantities and to think of "overflow" as a danger they must defend against. We essentially make the problem worse.

So yes, I would like `RANGES_REALLY_UNSIGNED` or `RANGES_INTENTIONALLY_UNSIGNED` or `RANGES_TELL_UBSAN_TO_SHUT_UP` to distinguish functions that use unsigned types on purpose from those that improperly use unsigned to mean non-negative. (I suppose I can live with `RANGES_INTENDED_WRAP_AROUND`.) But if you really like `RANGES_INTENDED_UNSIGNED_OVERFLOW` then just be patient, I don't think @ericniebler is as militant about signed-vs-unsigned as I am. He may tell you to keep it ;)

gnzlbg: > I apologize for nitpicking at semantics here

No need to apologize, I think this discussion is worth having. 

> have repeatedly chosen to use unsigned types to mean non-negative 

I wouldn't really blame them that much. They are called `unsigned` after all. Its deviously misleading.

> I suppose I can live with `RANGES_INTENDED_WRAP_AROUND`.

I think I would go for `RANGES_INTENDED_UNSIGNED_WRAP_AROUND`, since we don't intend signed integers to wrap around.  `RANGES_INTENDED_UNSIGNED_OVERFLOW` is a bad name though, I don't want to keep that.

cmake/ranges_flags.cmake
gnzlbg: >  I'm assuming the module maps in the OS X SDK give you trouble?

See this: https://bugs.llvm.org//show_bug.cgi?id=32192

It would be cool if somebody could attempt to reproduce my minimal example there on MacOSX. I cannot believe that this is true. I must have made a mistake somewhere.

gnzlbg:
So I need to:

- [ ] work around the lack of `__has_attribute` in GCC 4.9,
- [ ] fix Boost usage of `std::auto_ptr` which was removed from libc++,
- [ ] once the other branches are merged, rebase the build system changes they include

include/range/v3/detail/config.hpp
ericniebler: > I don't think @ericniebler is as militant about signed-vs-unsigned as I am

`$ericniebler = $CaseyCarter =~ s/dogmatic/pragmatic/gr;`

;-)

`RANGES_INTENDED_MODULAR_ARITHMETIC`?

CaseyCarter:
```perl
$ericniebler = $CaseyCarter =~ s/dogmatic/pragmatic/gr;
```

You are mocking me in Perl; I think I came out of this ahead ;)

cmake/ranges_flags.cmake
```diff
@@ -0,0 +1,184 @@
...
+ranges_append_flag(RANGES_HAS_WNO_CXX98_COMPAT -Wno-c++98-compat)
+ranges_append_flag(RANGES_HAS_WNO_CXX98_COMPAT_PEDANTIC -Wno-c++98-compat-pedantic)
+ranges_append_flag(RANGES_HAS_WNO_WEAK_VTABLES -Wno-weak-vtables)
+ranges_append_flag(RANGES_HAS_WNO_PADDED -Wno-padded)
+ranges_append_flag(RANGES_HAS_WNO_MISSING_PROTOTYPES -Wno-missing-prototypes)
+ranges_append_flag(RANGES_HAS_WNO_MISSING_VARIABLE_DECLARATIONS -Wno-missing-variable-declarations)
+ranges_append_flag(RANGES_HAS_WNO_DOCUMENTATION -Wno-documentation)
+ranges_append_flag(RANGES_HAS_WNO_DOCUMENTATION_UNKNOWN_COMMAND -Wno-documentation-unknown-command)
+ranges_append_flag(RANGES_HAS_WNO_OLD_STYLE_CAST -Wno-old-style-cast)
+ranges_append_flag(RANGES_HAS_WNO_OLD_STYLE_CAST -Wno-shadow)
```
CaseyCarter: s/RANGES_HAS_WNO_OLD_STYLE_CAST/RANGES_HAS_WNO_SHADOW/

cmake/ranges_flags.cmake
EricWF: I'll attempt to reproduce this weekend.

cmake/ranges_flags.cmake
```diff
@@ -0,0 +1,184 @@
...
+
+  if (RANGES_CXX_COMPILER_CLANGC2)
+    ranges_append_flag(RANGES_HAS_O2 -O2)
+  else()
+    ranges_append_flag(RANGES_HAS_OFAST -Ofast)
+  endif()
+
+  ranges_append_flag(RANGES_HAS_STRICT_ALIASING -fstrict-aliasing)
+  ranges_append_flag(RANGES_HAS_STRICT_VTABKE_POINTERS -fstrict-vtable-pointers)
+  ranges_append_flag(RANGES_HAS_DEFINE_FAST_MATH "-D__FAST_MATH__")
```
CaseyCarter: Why are we defining this implementation-reserved name instead of passing the `-ffast-math` option?

gnzlbg: IIRC some clang versions did not define this but libstdc++ uses it.

This was assuming that `-ffast-math` was enabled in release, but now it isn't enabled anymore for clangC2, so this is wrong.

ericniebler:
Needs merge conflict resolution.

ericniebler:
Merge and rebase, please?

cmake/ranges_flags.cmake
gnzlbg: @EricWF I've removed it and travis passes, no idea where the problem was. It was there because something did not work back then. I'll just remove it.

cmake/ranges_flags.cmake
gnzlbg: @EricWF were you able to reproduce this?

ericniebler:
Needs rebase.

cmake/ranges_flags.cmake
gnzlbg: @EricWF I screwed up. I removed it from the CMake files but the `.travis.yml` script was still passing it to clang. You can see an example of the breakage [here](https://travis-ci.org/gnzlbg/range-v3/jobs/221703915#L1838) and also all the other builds that break [here]().

Example:

```shell
/usr/include/x86_64-linux-gnu/bits/math-finite.h:177:1: error: unknown type name
      '__extern_always_inline'
__extern_always_inline double __NTH (lgamma (double __d))
^
```

Clang versions affected: 3.6, 3.7, 3.8, 3.9. It works fine in clang-trunk, clang 4.0 is not tested yet (I'll add build bots for it).

gnzlbg:
So I've rebased this, added build bots for clang-4.0, and changed the clang tests to use the appropriate libc++ version. Everything should be green except for the clang-3.6 with libc++-3.6 tests, which fail to compile (see issue #640 ).

ericniebler:
ASAN doesn't seem to like libc++'s `stringstream`. Possible dupe of https://github.com/google/sanitizers/issues/557?

gnzlbg:
The older libcxx versions where not annottated for UBSan so I need to
disable it for the versions that break.

ericniebler:
ASAN failures for act.insert, act.push_back, and act.push_front. Interesting, or is this another case where older libc++ needed annotations?

Also, ASAN linker problems for the later releases in the clang-3 series. :-P

gnzlbg:
@ericniebler I've decided to disable all sanitizers for libc++ < 4.0. Clang trunk/libc++ trunk (and in the near future clang-4.0/libc++4.0) are tested with ASan, UBSan, MSan, and LSan, that should be enough. We can always enable ASan in older clang/libc++/libc++abi combinations later. 

There are still some tests failing in clang-3.8 and clang-3.9 which did not fail with libc+-trunk, I have to look into those.

@EricWF thanks for improving libc++/libc++abi's build system in LLVM >= 4.0. It makes installing them w/o sanitizers a breeze! If you have any tips on how to improve sanitizer coverage for older clang versions please ping me on slack.

gnzlbg:
@ericniebler since valgrind fails in the same tests as ASan I guess the failures we are seeing here are real. 

It seems that clang-3.8 + libc++-3.8 and clang-3.9 + libc++-3.9 have some sort of memory issue :/
I sadly won't have much time to investigate this further till july, but I can update this branch to the latest master in the mean time.

ericniebler:
I've looked over range-v3's insert actions and haven't found anything untoward. It's likely this isn't our bug.

EricWF:
I didn't get libc++ ubsan clean until 4.0. Maybe that's what you're experiencing?

gnzlbg:
@EricWF some of the errors look like:

```shell
29: MemCheck command: /usr/bin/valgrind "--log-file=/home/travis/build/ericniebler/range-v3/build/Testing/Temporary/MemoryChecker.29.log" "--trace-children=yes" "--leak-check=full" "/home/travis/build/ericniebler/range-v3/build/test/action/act.stable_sort"
29: Test timeout computed to be: 1500
28: > ERROR: CHECK failed "rgs[0] == "This""
28: > 	/home/travis/build/ericniebler/range-v3/test/action/split.cpp(47)
28: > 	EXPECTED: This
28: > 	ACTUAL: his
```

or 

```shell
154: MemCheck command: /usr/bin/valgrind "--log-file=/home/travis/build/ericniebler/range-v3/build/Testing/Temporary/MemoryChecker.154.log" "--trace-children=yes" "--leak-check=full" "/home/travis/build/ericniebler/range-v3/build/test/utility/utility.variant"
154: Test timeout computed to be: 1500
154: > ERROR: CHECK failed "get<1>(v) == "hello world""
154: > 	/home/travis/build/ericniebler/range-v3/test/utility/variant.cpp(91)
154: > 	EXPECTED: hello world
154: > 	ACTUAL: ello world
154: 
154: 
154: 
154: 
154: 
154: 
154: 
154: 
154: 
154: 
154: 
154: ï¿½t
154: 
154: 
154: .N=ï¿½
154: 
154: 
154: 
154: ^ï¿½
154: 
154: 
154: 
154: 
154: Xc
154: 
```

CaseyCarter:
I've reproduced the errors locally with clang 3.9. `std::string` seems to be the primary culprit; this program:
```c++
#include <iostream>
#include <string>

int main() {
  std::string s{"Hello, World!\n"};
  std::cout << s;
}
```
outputs (piped through `hexdump -c`):
<pre>
0000000   e   l   l   o   ,       w   o   r   l   d   !  \n  \0  \0  \0
0000010  \0  \0  \0  \0  \0  \0 016   $ 032   @  \0  \0  \0  \0  \0   @
0000020
0000024
</pre>
If I modify `install_libcxx.sh` to build libc++ without `-DLIBCXX_ABI_UNSTABLE=ON`, it instead outputs:
<pre>
0000000   H   e   l   l   o   ,       w   o   r   l   d   !  \n
000000e
</pre>

I think the "fix" is to build without `LIBCXX_ABI_UNSTABLE` for versions less than 4ish; doing so works for me locally with 3.9. If my test flight succeeds (https://travis-ci.org/CaseyCarter/range-v3/builds/267617654), I'll cleanup my rebased branch with these changes and we can finally get this change merged.

ericniebler:
> I think the "fix" is to build without `LIBCXX_ABI_UNSTABLE` for versions less than 4ish 

Seems like a hack that is hiding a real problem. Is some range-v3 code making a bad assumption about the behavior of `std::string`, or is this a libc++ bug? Either way, it should be tracked down and squashed.

CaseyCarter:
> Seems like a hack that is hiding a real problem. Is some range-v3 code making a bad assumption about the behavior of `std::string`, or is this a libc++ bug? Either way, it should be tracked down and squashed.

It's definitely a libc++ bug: notice that the program I used to repro the bug doesn't include any part of range-v3. Given that it doesn't repro for the trunk builds on Travis, I assume that the bug has been fixed. @EricWF will correct me if I'm wrong, but I suspect that `LIBCXX_ABI_UNSTABLE` wasn't something that was actively supported for external use in the 3.8-3.9 era and that it's not a problem that it was broken.

CaseyCarter:
> @EricWF will correct me if I'm wrong, but I suspect that `LIBCXX_ABI_UNSTABLE` wasn't actively supported for external use in the 3.8-3.9 era and that it's not a problem that it was broken.

Clarification: `LIBCXX_ABI_UNSTABLE` causes `_LIBCPP_ABI_UNSTABLE` to be defined during the build. When `_LIBCPP_ABI_UNSTABLE` is defined, various components - notably `string` - use different/incompatible layouts than in the typical stable ABI configuration. The cmake install commands *which we do not use for versions < 4* create a `__config` header that properly defines `_LIBCPP_API_UNSTABLE` to agree with the compiled bits.

The fix is to either (a) not specify `LIBCXX_ABI_UNSTABLE` for older libc++ versions, or (b) hand-roll the `__config` header. My money is on (a), especially given that that stable ABI is what we should really be testing against since that's what users will be using.

CaseyCarter:
@gnzlbg I've rebased this onto master, cleaned up the clang 3.8/3.9 failures, and thrown in a couple of small Travis changes on my [`cmake_refactor` branch](https://github.com/CaseyCarter/range-v3/tree/cmake_refactor). Would you please grab that commit (there's only one on the branch) and force push it into this PR?

gnzlbg:
I'm traveling today and the internet connection im getting is pretty bad.
It looks like it will get worse so feel free to open a new PR and close
mine. Otherwise I'll do that ASAP but that might be Monday...

I agree with disabling the unstable ABI for longer versions but @EricWF is
there a reason to use it on newer versions?

gnzlbg:
@CaseyCarter  I've forced-push and resolved the conflicts.

ericniebler:
Huzzah! Thanks for the team effort.

PR #603, opened 2017-03-20
Fix some size + single-pass bugs:

CaseyCarter:
* `iter_size` can work fine with *references* to single-pass iterators

* `view_interface` should not trash single-pass ranges by calling `iter_size(begin(), end())`

* `size` should not trash single-pass ranges by calling `iter_size(begin(), end())`

### DIFF
--- a/include/range/v3/size.hpp
+++ b/include/range/v3/size.hpp
@@ -65,7 +65,8 @@ namespace ranges
                     size(r)
                 )
 
-                template<typename R>
+                template<typename R, typename I = decltype(ranges::cbegin(std::declval<R const&>())),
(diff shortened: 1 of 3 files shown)


include/range/v3/size.hpp
```diff
@@ -65,7 +65,8 @@ namespace ranges
                     size(r)
                 )
 
-                template<typename R>
+                template<typename R,
+                    CONCEPT_REQUIRES_(ForwardRange<R const>())>
```
ericniebler: You can't use the Range concepts here. `size` is used to define them.

CaseyCarter:
Build errors are clang-5 transient.

PR #604, opened 2017-03-20
Implement view::cartesian_product

CaseyCarter:
* Expose `common_tuple::base`

* Move `detail:deref_fun` into functional.hpp and rename to `dereference` so `cartesian_product can reuse it

* Drive-by:
  * Polish constraints in `to_container`
  * sort view.hpp
  * Pass function objects by reference internally in the tuple algorithms

### DIFF
--- a/include/range/v3/utility/functional.hpp
+++ b/include/range/v3/utility/functional.hpp
@@ -181,6 +181,17 @@ namespace ranges
           : coerce<T>
         {};
 
+        struct dereference_fn
+        {
+            template<typename I>
+            constexpr auto operator()(I &i) const
+            RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+            (
+                *i
+            )
+        };
+        RANGES_INLINE_VARIABLE(dereference_fn, dereference)
+
         /// \addtogroup group-concepts
         /// @{
         namespace concepts
--- a/include/range/v3/view/cartesian_product.hpp
+++ b/include/range/v3/view/cartesian_product.hpp
@@ -49,26 +49,12 @@ namespace ranges
             struct cartesian_size_fn
             {
                 template<typename Rng, CONCEPT_REQUIRES_(SizedRange<Rng>())>
-                auto operator()(std::size_t s, Rng && rng)
+                auto operator()(std::size_t s, Rng &&rng)
                 RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
                 (
                     s * static_cast<std::size_t>(ranges::size(rng))
                 )
             };
-
-            template<typename View,
-                CONCEPT_REQUIRES_(BoundedRange<View>())>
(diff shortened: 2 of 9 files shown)


include/range/v3/utility/functional.hpp
```diff
@@ -181,6 +181,17 @@ namespace ranges
           : coerce<T>
         {};
 
+        struct dereference_fn
+        {
+            template<typename I>
+            constexpr auto operator()(I &i) const
```
ericniebler: If we're exposing this, do we really want to disallow rvalues?

include/range/v3/view/cartesian_product.hpp
```diff
@@ -0,0 +1,377 @@
...
+            {
+                return ranges::end(v);
+            }
+            template<typename View,
+                CONCEPT_REQUIRES_(!BoundedRange<View>() &&
+                    RandomAccessRange<View>() && SizedRange<View>())>
+            static iterator_t<View> last_iter(View& v)
+            {
+                return ranges::begin(v) + ranges::size(v);
+            }
```
ericniebler: Isn't this what `next(begin(v), end(v))` does?

include/range/v3/view/cartesian_product.hpp
```diff
@@ -0,0 +1,377 @@
...
+            CONCEPT_REQUIRES(CanSize<false>() && !CanSize<true>())
+            std::size_t size()
+            {
+                return tuple_foldl(views_, std::size_t{1},
+                    detail::cartesian_size_fn{});
+            }
+        };
+
+        namespace view {
```
ericniebler:
```c++
namespace view
{
```

include/range/v3/utility/functional.hpp
CaseyCarter: I don't want to add more functions that return possibly-dangling references obtained from temporary iterators. (https://github.com/ericniebler/stl2/issues/156)

include/range/v3/view/cartesian_product.hpp
CaseyCarter: Yes; the only difference is that `last_iter` avoids calling `begin` on `BoundedRange`s, and is always O(1).

`last_iter` is only used in `cursor::prev_`, which is only called by `cursor::prev`, which requires `CanBidi`, which basically implies that `ranges::next(begin(v), end(v))` can use one of the O(1) implementations of `ranges::advance`. So I can dump `last_iter` and replace it with a call to `next`.

include/range/v3/view/cartesian_product.hpp
CaseyCarter: Grrr, I have some misplaced reference qualifiers as well.

CaseyCarter:
Comments addressed; please squash on merge (too lazy to write an actual commit log)

CaseyCarter:
Travis failures are clang-5 transient.

PR #619, opened 2017-03-28
test constexpr ranges::swap of initializer_list now that ranges::swap uses poison pill

ericniebler:


ericniebler:
No, I had to take that hack out. There were bizarre interactions having to do with the fact that in some places `noexcept` clauses were causing hard errors. I don't remember the details.

CaseyCarter:
> Won't `ranges::swap` call `std::swap` if it is constrained?

Only if `std::swap` is more-specialized than the poison pill. I suppose an implementation could overload `std::swap` for `initializer_list`s?

ericniebler:
> Only if std::swap is more-specialized than the poison pill.

That's the only time `ranges::swap` calls `std::swap`.

> I suppose an implementation could overload std::swap for initializer_lists?

Unlikely. Anyway, this PR is just an experiment. I'll merge it if it passes the Travis sniff test.

CaseyCarter:
> Unlikely.

Yeah, agreed. I do believe it would be conforming to do so, but can imagine why an implementation would bother.

PR #629, opened 2017-04-04
unbreak Readable concept for types that return move-only types by value

ericniebler:


ericniebler:
See ericniebler/stl2#399

PR #631, opened 2017-04-06
view_predicate should be true if either T or const T does not satisfy…

CaseyCarter:
… Range

per [#ranges discussion](https://cpplang.slack.com/archives/C4Q3A3XB8/p1491472624545910).


CaseyCarter:
Where the test coverage at?

CaseyCarter:
> Where the test coverage at?

History rewritten with test coverage. In this timeline, I never made the mistake.

Issue #567, opened 2017-02-27
view::chunk does not work after view::join


FloopCZ:
Consider the following code:
```c++
#include <vector>
#include <range/v3/view.hpp>
using namespace ranges;

int main()
{
  std::vector<std::vector<int>> data{{1, 2, 3}, {4, 5, 6}};
  auto rng = data | view::join | view::chunk(3);
}
```
The compilation fails with the following static assertion: "The first argument to view::chunk must be a model of the ForwardRange concept". Does view::join somehow violate ForwardRange concept?

FloopCZ:
I have examined the code and now I see the problem. view::join implements only the InputRange concept, because its cursors share a single "outer" and a single "inner" range stored inside the view object. If it should implement ForwardRange (multiple cursors allowed), each cursor would need its own "inner" range.

view::chunk requires ForwardRange argument, since in each increment, it performs two steps: take(n), advance(n). That's why the provided range has to support multiple cursors.

I see two solutions:
1. Store the "inner" range of the view::join inside the cursor (instead of inside the view object). For the cursor to be copyable and movable including its member iterators, the "inner" range would have to be a [view::shared](https://github.com/ericniebler/range-v3/pull/557) (so that the iterators would still be valid after copy/move).
2. Make a variant of view::chunk for InputRanges. That would require the chunk cursor to buffer the whole chunks, let's say in std::vector, and return the vector.

The first option is bad, since view::join would require dynamic heap allocation. The second option is even worse, since view::chunk iterator would contain the entire chunk. Can anyone see a better option?

Victor-Savu:
@FloopCZ I wonder if solution 2 could be improved by providing an action (e.g. `action::buffer`) which would convert the InputRange into a ForwardRange by doing the buffering on-demand. I am not familiar with the internals (I would have to look at it this evening after work) but I can imagine that the buffering behavior can be extracted out of the implementation for your proposed second solution and put in a general buffering action. This way, heap allocation and buffering would happen, but it is at least explicit and keeps view::chunk lightweight.
I am not claiming this would be a good solution for this particular problem (not even that it makes option 2 any better than option 1), but if there is a feasible implementation it might solve a more general class of problems where implicit buffering is to be avoided.

Food for thought: is there a class of actions that can transform an InputRange into a ForwardRange? (i.e. a class which action::buffer would be a part of)

FloopCZ:
[...]

PR #635, opened 2017-04-08
Implement chunk_view for input ranges

CaseyCarter:
Partially addresses #567.

### DIFF
--- a/include/range/v3/size.hpp
+++ b/include/range/v3/size.hpp
@@ -38,7 +38,7 @@ namespace ranges
             {
             private:
                 template<typename R, std::size_t N>
-                static constexpr std::size_t impl_(R const (&)[N], int) noexcept
+                static constexpr std::size_t impl_(R (&)[N], int) noexcept
                 {
                     return N;
                 }
@@ -46,9 +46,9 @@ namespace ranges
                 // Prefer member if it returns Integral.
                 template<typename R,
                     typename = meta::if_c<!disable_sized_range<R>()>,
-                    typename N = decltype(aux::copy(std::declval<R const &>().size())),
+                    typename N = decltype(aux::copy(std::declval<R &>().size())),
                     CONCEPT_REQUIRES_(Integral<N>())>
-                static constexpr N impl_(R const &r, int)
+                static constexpr N impl_(R &r, int)
                 RANGES_AUTO_RETURN_NOEXCEPT
                 (
                     r.size()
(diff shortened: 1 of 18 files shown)


include/range/v3/view/chunk.hpp
```diff
@@ -62,74 +70,104 @@ namespace ranges
...
+                RANGES_CXX14_CONSTEXPR
                 range_difference_type_t<Rng> distance_to(iterator_t<Rng> const &here,
                     iterator_t<Rng> const &there, adaptor const &that) const
                 {
-                    // This assertion is true for all range types except cyclic ranges:
-                    //RANGES_EXPECT(0 == ((there - here) + that.offset() - offset()) % n_);
-                    return ((there - here) + that.offset() - offset()) / n_;
+                    auto const delta = (there - here) + (that.offset() - offset());
+                    // This can fail for cyclic base ranges when the chunk size does not divide the
+                    // cycle length. Such iterator pairs are NOT in the domain of -.
```
ericniebler: Where do we document the domain of `op-` for iterators of cyclic ranges?

include/range/v3/view/chunk.hpp
```diff
@@ -14,114 +14,360 @@
...
+                    offset() = 0;
+                }
+                CONCEPT_REQUIRES(detail::CanSizedSentinel<Rng>())
+                RANGES_CXX14_CONSTEXPR
+                range_difference_type_t<Rng> distance_to(iterator_t<Rng> const &here,
+                    iterator_t<Rng> const &there, adaptor const &that) const
+                {
+                    auto const delta = (there - here) + (that.offset() - offset());
+                    // This can fail for cyclic base ranges when the chunk size does not divide the
+                    // cycle length. Such iterator pairs are NOT in the domain of -.
```
ericniebler: Where do we document the domain of `operator-` for the iterators of cyclic ranges?

include/range/v3/size.hpp
```diff
@@ -38,17 +38,17 @@ namespace ranges
             {
             private:
                 template<typename R, std::size_t N>
-                static constexpr std::size_t impl_(R const (&)[N], int) noexcept
+                static constexpr std::size_t impl_(R (&)[N], int) noexcept
```
CaseyCarter: I `s/const //g`-ed the entire definition. It obviously makes no difference to this overload, but does maintain a consistent appearance.

include/range/v3/view/chunk.hpp
CaseyCarter: `SizedSentinel` specifies that the difference must be well-defined when it is representable by the difference type; this view is non-conforming. Luckily it explodes the program instead of allowing non-conformance to be observed. I'm not terribly happy with that as a long-term solution, but I do think it is categorically better than the prior garbage-returning behavior.

It feels like overkill to remove difference support from `chunk_view` altogether; solving this permanently needs more thought.

include/range/v3/view/chunk.hpp
```diff
@@ -14,114 +14,325 @@
...
+                Rng,                          // base
+                range_difference_type_t<Rng>, // n
+                range_difference_type_t<Rng>, // remainder
+                iter_cache_t                  // it
+            > data_{};
+
+            RANGES_CXX14_CONSTEXPR Rng &base() noexcept { return ranges::get<0>(data_); }
+            constexpr Rng const &base() const noexcept { return ranges::get<0>(data_); }
+            RANGES_CXX14_CONSTEXPR range_difference_type_t<Rng> &n() noexcept { return ranges::get<1>(data_); }
+            constexpr range_difference_type_t<Rng> const &n() const noexcept { return ranges::get<1>(data_); }
```
CaseyCarter: It is, by a bunch. Fixing.

PR #645, opened 2017-04-16
Fix MSVC debugging issues

khlebnikov:
For some reason, MSVC 2017 (clang/C2) failed to find header files during debugging (displaying the "Source not found..." page). Adding the headers to the solution fixes the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,11 +87,16 @@ range_v3_append_flag(RANGE_V3_HAS_MTUNE_NATIVE "-mtune=native")
 file(GLOB_RECURSE RANGE_V3_PUBLIC_HEADERS
                   RELATIVE "${CMAKE_SOURCE_DIR}/include"
                   "${CMAKE_SOURCE_DIR}/include/*.hpp")
+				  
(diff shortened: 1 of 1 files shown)


ericniebler:
I can't test this, and I'm hardly a cmake expert. @gnzblg, @CaseyCarter, does this look good to you?

gnzlbg:
@khlebnikov how are you including range-v3 in your project? As a CMake [`INTERFACE`](https://cmake.org/cmake/help/v3.0/manual/cmake-buildsystem.7.html#interface-libraries) ?

khlebnikov:
@gnzlbg Not really, I just use range-v3 as the primary project and "debug" the tests to see which overloads are being chosen.

CaseyCarter:
This seems sane: it adds dependencies on the headers to the phony "header" target. The only effect should be to have them show up in IDE project files (according to [the documentation of `add_custom_target`](https://cmake.org/cmake/help/v3.0/command/add_custom_target.html), anyway).

ericniebler:
Thanks, @khlebnikov!

PR #647, opened 2017-04-19
for_each_n algorithm

khlebnikov:


include/range/v3/algorithm/for_each_n.hpp
```diff
@@ -0,0 +1,67 @@
...
+    inline namespace v3
+    {
+        /// \addtogroup group-algorithms
+        /// @{
+        struct for_each_n_fn
+        {
+            template<typename I, typename F, typename P = ident,
+                CONCEPT_REQUIRES_(InputIterator<I>() &&
+                    IndirectInvocable<F, projected<I, P>>())>
+            tagged_pair<tag::in(I), tag::fun(F)>
```
CaseyCarter: `std::for_each_n` does not return the function object, the range version should also not do so. Callers who want to recover the state of a mutable function object can simply pass a `reference_wrapper`.

include/range/v3/algorithm/for_each_n.hpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    inline namespace v3
+    {
+        /// \addtogroup group-algorithms
+        /// @{
+        struct for_each_n_fn
+        {
+            template<typename I, typename F, typename P = ident,
+                CONCEPT_REQUIRES_(InputIterator<I>() &&
+                    IndirectInvocable<F, projected<I, P>>())>
```
CaseyCarter: This requires `F` to be `CopyConstructible`, whereas the standard only requires `MoveConstructible`. Yes, this is the only occurrence of a function object argument to a standard library algorithm that is not required to be `CopyConstructible`. Yes, I think it is silly to relax this single requirement and create that inconsistency. We don't have a great way to express the relaxed requirement: I would inline the expansion of `IndirectInvocable<F, projected<I, P>>` and replace the `CopyConstructible<F>` term with `MoveConstructible<F>`.

include/range/v3/algorithm/for_each_n.hpp
```diff
@@ -0,0 +1,67 @@
...
+                    invoke(fun, invoke(proj, *b));
+                return {recounted(begin, b, norig), detail::move(fun)};
+            }
+
+            template<typename Rng, typename F, typename P = ident,
+                CONCEPT_REQUIRES_(InputRange<Rng>() &&
+                    IndirectInvocable<F, projected<iterator_t<Rng>, P>>())>
+            tagged_pair<tag::in(safe_iterator_t<Rng>), tag::fun(F)>
+            operator()(Rng &&rng, range_difference_type_t<Rng> n, F fun, P proj = P{}) const
+            {
```
CaseyCarter:
```c++
if (SizedRange<Rng>())
    RANGES_EXPECT(n <= distance(rng));
```

include/range/v3/algorithm/for_each_n.hpp
khlebnikov: Yes, makes sense. So if I understand the motivation correctly - we should just basically replicate the interface provided by the STL. This makes perfect sense. Question though - the STL `for_each` doesn't return the iterator - it returns only the functor. Why does range `for_each` return both?

include/range/v3/algorithm/for_each_n.hpp
khlebnikov: Looking at n4659, regular `for_each` also requires for `F` to be only `MoveConstructible` - should this be fixed as well?

include/range/v3/algorithm/for_each_n.hpp
khlebnikov: (In the meanwhile I'll continue reading the range proposal - maybe the answer is there :) )

include/range/v3/algorithm/for_each_n.hpp
CaseyCarter: Blech, I suppose so. How'd they sneak that change into 17 without me noticing!??

include/range/v3/algorithm/for_each_n.hpp
CaseyCarter: The final iterator position is useful information calculated by the algorithm, so we return it. With input iterators, the caller has no other way to recover the final iterator value.

khlebnikov: Please bear with me - just trying to fully understand the logic behind the design decisions so there's less time required from you for reviewing future pull requests from me :)

Here's how I understand it:

- STL `for_each` doesn't return the iterator because it takes two iterators of the same type as `first` and `last`, so the user has the access to `last` anyway and the algorithm wouldn't provide additional info by returning the final iterator
- Range `for_each` may take a sentinel, so the final iterator might contain some more information than the user has. But because STL returns the function object - range `for_each` returns both with a tagged pair. But because the STL `for_each_n` returns the iterator - so does range `for_each_n` and there's actually no need to return the function object (for reasons you mentioned).

Any misunderstandings from my side?

include/range/v3/algorithm/for_each_n.hpp
CaseyCarter: > How'd they sneak that change into 17 without me noticing!??

They backdoored it into 17 by including [the resolution of DR 1110](http://cplusplus.github.io/LWG/lwg-defects.html#1110) in C++11.

include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -497,55 +497,56 @@ namespace ranges
...
+                    iter_map_reduce_fn_<
+                        meta::bind_front<meta::quote<concepts::Invocable::result_t>, C&>,
+                        meta::quote<CommonReference>>,
+                    Is...>>;
         }
 
         template<typename C, typename ...Is>
         using IndirectInvocable = meta::and_<
+            detail::common_result_indirect_invocable_<C, Is...>,
             CopyConstructible<uncvref_t<C>>,
```
ericniebler: I think it's a bug that the `IndirectInvocable` concepts are testing for `CopyConstructible<uncvref_t<C>>`. It should just be `CopyConstructible<C>`. The algorithms take callables by value everywhere, so these concepts should never be passed a qualified type.

include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -497,55 +497,56 @@ namespace ranges
...
-                    meta::bind_front<meta::quote<concepts::Invocable::result_t>, C&>,
-                    meta::quote<CommonReference>>,
-                Is...>>;
+            CopyConstructible<C>>;
 
         template<typename C, typename ...Is>
-        using IndirectRegularInvocable = IndirectInvocable<C, Is...>;
+        using MoveIndirectInvocable = meta::and_<
+            detail::common_result_indirect_invocable_<C, Is...>,
+            MoveConstructible<uncvref_t<C>>,
```
ericniebler: Delete this.

include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -497,55 +497,56 @@ namespace ranges
...
+            MoveConstructible<C>>;
 
         template<typename C, typename ...Is>
         using IndirectRegularInvocable = IndirectInvocable<C, Is...>;
 
         template<typename C, typename ...Is>
         using IndirectPredicate = meta::and_<
+            detail::indirect_invocable_<Predicate, C, Is...>,
             CopyConstructible<uncvref_t<C>>,
```
ericniebler: Delete this.

include/range/v3/utility/iterator_concepts.hpp
```diff
@@ -497,55 +497,56 @@ namespace ranges
...
-                detail::iter_map_reduce_fn_<
-                    meta::bind_front<meta::quote<Predicate>, C&>,
-                    meta::quote<meta::strict_and>>,
-                Is...>>;
+            CopyConstructible<C>>;
 
         template<typename C, typename I0, typename I1 = I0>
         using IndirectRelation = meta::and_<
+            detail::indirect_invocable_<Relation, C, I0, I1>,
             CopyConstructible<uncvref_t<C>>,
```
ericniebler: Delete this.

include/range/v3/utility/iterator_concepts.hpp
khlebnikov: Yes, I looked at this from this perspective, but thought maybe I'm missing something - so I left it as it was. I'll fix it.

ericniebler:
Thanks!

PR #649, opened 2017-04-25
Reducing duplication with unreachable{} (numeric)

khlebnikov:
`unreachable`'s `constexpr` checks should have no effect on the code generation (and looking at optimized assembly in clang/cl - it indeed doesn't).

### DIFF
--- a/include/range/v3/numeric/adjacent_difference.hpp
+++ b/include/range/v3/numeric/adjacent_difference.hpp
@@ -26,6 +26,7 @@
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/iterator_concepts.hpp>
 #include <range/v3/utility/static_const.hpp>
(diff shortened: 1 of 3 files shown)


include/range/v3/numeric/adjacent_difference.hpp
```diff
@@ -81,6 +55,7 @@ namespace ranges
             operator()(I begin, S end, O result, S2 end_result, BOp bop = BOp{},
                        P proj = P{}) const
             {
+                // BUGBUG think about the use of coerce here.
                 using V = value_type_t<I>;
                 using X = concepts::Invocable::result_t<P, V>;
```
ericniebler: Pretty sure this should be `concepts::Invocable::result_t<P&, V>` here. Mind changing that?

khlebnikov: Yep, it should be `P&`. Fixed.

PR #650, opened 2017-04-27
RANGES_FORWARD

khlebnikov:
I think this conveys the intent better (while still maintaining "marginal compilation speed improvement"). Thoughts?


ericniebler:
No, thank you. I try to keep the use of macros to a bare minimum.

PR #653, opened 2017-04-27
Fix typo in view::chunk

FloopCZ:


include/range/v3/view/chunk.hpp
```diff
@@ -418,7 +418,7 @@ namespace ranges
                 void operator()(Rng &&, T) const
                 {
                     CONCEPT_ASSERT_MSG(InputRange<Rng>(),
-                        "The first argument to view::chunk must satisfy the ForwardRange concept");
+                        "The first argument to view::chunk must satisfy the InputRange concept");
```
CaseyCarter: Typical - I forget to change documentation even when it's literally on the next line inside a string literal. Thanks.

Issue #659, opened 2017-05-04
Build errors under GCC 7.1 


koalization:
I tried to build this wonderful library under the newly released version of GCC. Steps I followed:

1. `cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME/BUILDS/RANGE -DRANGE_V3_NO_HEADER_CHECK=1`
2. `make -j8`
 
While this works flawlessly with GCC 6.3.0 (both my own build on Ubuntu 16.04.2 and the one shipped with Debian Testing), it fails under GCC 7 (own builds of the stable version 7.1 on both Debian Testing/Ubuntu 16.04.2 and the pre-release version currently shipped with Fedora 26 (GCC 7.0.1), which is in turn a pre-release distribution). Here's the stderr:

```
/home/koalization/BUILDS/range-v3-master/test/getlines.cpp: In function 'int main()':
/home/koalization/BUILDS/range-v3-master/test/getlines.cpp:32:38: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     using Rng = decltype(getlines(sin));
                                      ^
/home/koalization/BUILDS/range-v3-master/test/getlines.cpp:32:38: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
cc1plus: all warnings being treated as errors
make[2]: *** [test/CMakeFiles/getlines.dir/build.make:63: test/CMakeFiles/getlines.dir/getlines.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:8968: test/CMakeFiles/getlines.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[  2%] Building CXX object test/CMakeFiles/multiple.dir/multiple2.cpp.o
[  4%] Linking CXX executable view_facade
[  4%] Built target view_facade
[  4%] Linking CXX executable bug474
[  4%] Built target bug474
[  4%] Linking CXX executable iterator_range
[  4%] Built target iterator_range
[  4%] Linking CXX executable index
[  4%] Built target index
[  4%] Linking CXX executable multiple
[  4%] Built target multiple
[  4%] Linking CXX executable view_adaptor
[  4%] Built target view_adaptor
[  4%] Linking CXX executable container_conversion
[  4%] Built target container_conversion
make: *** [Makefile:141: all] Error 2

```

[...]

PR #660, opened 2017-05-05
Build on GCC 7

CaseyCarter:
Partially addresses #659

* Disable `-Wnoexcept-type` globally

* Perma-workaround `-Wstrict-aliasing` false positive in getlines.cpp ([GCC PR 80633](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80633))

* Disable `-Wmaybe-uninitialized` in view/replace_if.cpp (I *think* this is a false positive, but it needs proper investigation.)


ericniebler:
Please open a task to remove or improve the workaround when possible.

CaseyCarter:
> Please open a task to remove or improve the workaround when possible.

I left #659 open for that reason, although it would be nice to retitle the issue.

PR #665, opened 2017-05-27
picky, picky, clang

ericniebler:


CaseyCarter:
I think we're still formally violating ODR since these objects with internal linkage are being used in the definitions of entities with external linkage without immediate lvalue-to-rvalue conversion. We should probably give them the `RANGES_INLINE_VARIABLE` treatment as well.

PR #667, opened 2017-06-08
Use to_address(p) instead of addressof(*p)

glenfe:
Before using placement new, addressof(*p) is not well-defined given that addressof takes T& and there's no object constructed in the storage yet.

### DIFF
--- a/include/range/v3/detail/variant.hpp
+++ b/include/range/v3/detail/variant.hpp
@@ -27,6 +27,7 @@
 #include <range/v3/utility/iterator_concepts.hpp>
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/functional.hpp>
+#include <range/v3/utility/to_address.hpp>
 
 namespace ranges
 {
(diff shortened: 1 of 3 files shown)


CaseyCarter:
> `addressof(*p)` is not well-defined given that addressof takes `T&` and there's no object constructed in the storage yet

[citation needed]

ericniebler:
This appears broken to me. Output iterators are not required to provide `operator->`. I don't see a way to salvage this.

glenfe:
@CaseyCarter Richard Smith pointed us to [CWG 453](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#453) for this (Jonathan Wakely drew my attention to the general issue and why addressof(*p) should not be used prior to construction).

(The other case they mentioned that addressof(*p) is insufficient for is the null pointer case, for which he pointed me to [CWG 232](http://www.open-std.org/Jtc1/sc22/wg21/docs/cwg_active.html#232) but that does not apply here I think.)

ericniebler:
The most recent proposed resolution for CWG 453 makes the original code well-formed. I'm not convinced the problem you're trying to solve is real, and I'm 100% certain that your proposed fix is unworkable.

glenfe:
@ericniebler I just saw your comment above, and yes, I understand now this isn't suitable here. Closing this issue.

PR #669, opened 2017-06-11
fix semiregular and optional over references

ericniebler:


include/range/v3/detail/optional.hpp
```diff
@@ -86,6 +86,49 @@ namespace ranges
             }
         };
 
+        template<typename T>
+        struct optional<T &>
+        {
+        private:
+            variant<meta::nil_, T &> data_;
```
CaseyCarter: This would be more efficiently implemented atop a `T*` than a variant, using `null` to represent the empty state.

ericniebler: Good point.

CaseyCarter: After which I *think* you get no benefit from specializing `semiregular` for references over simply using the base template.

ericniebler: No, because when instantiated with a reference type, `semiregular`'s two copy-assignment operators end up having the same signature --> hard error.

CaseyCarter: True: you'd have to disable one of them, and turn all the `move`s into `forward`s. The fifteen-line specialization is probably simpler.

PR #670, opened 2017-06-16
Batch of bugfixes

CaseyCarter:
Things I found while investigating move-only input views.

* Implement `meta::add_const_if` directly to avoid defer overhead. Add friendlier `meta::const_if` front-end for `add_const_if`. (Not a bugfix)

* Implement `empty` per the Ranges TS spec.

* `front` now requires a forward range. It's not technically *wrong* on input ranges, but all the uses in the library are bugs.

* `getline_` and `istream_range` can use the default `iter_move`; don't implement `move()` in their cursors. (Not a bugfix)

* `istream_range<T>` doesn't require `T` to be `MoveConstructible`, so `istream()` needs to construct its return object in-place.

* Implement `empty` for `iterator_range` since the default implementation in `view_interface` no longer works for input ranges (see below)

* remove unnecessary specialization of `view_predicate_<T[N]>`; array types don't model `View` since they're not `SemiRegular` (Not a bugfix)

`view_interface`:
* Implement distinct overload of `empty` for ranges with fixed cardinality
* Disable `operator bool`, `operator!`, and `empty` overloads that call `begin()` for `!ForwardView`
* `front()` requires `ForwardView`
* (minor syntax cleanups)

* Fix constraints for sizing `chunk_view`s that adapt input views

* `Readable` is no longer overconstraining for `indirect_fn` because it no longer requires `SemiRegular` (Not a bugfix)

* Don't call `empty` on input views in `intersperse_view`!

* Don't call `empty` and `front` on input views in `partial_sum_view`! (also, store the accumulator in a `semiregular_t` instead of `optional` while I'm in here.)

* `sample_view` should properly adapt mutable-only views

* simplify some flow control in `set_union_cursor` (Not a bugfix)

* Store `decay_t<Regex>` in `tokenize_view`, so it's actually a `View`

### DIFF
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -693,7 +693,7 @@ namespace meta
         /// we express it using \c defer as follows:
         ///
         /// \code
-        /// template<typename List>
+        /// template <typename List>
         /// using reverse = reverse_fold<List, list<>, lambda<_a, _b, defer<push_back,
         /// _a, _b>>>;
         /// \endcode
@@ -3204,7 +3204,7 @@ namespace meta
         /// A lexically scoped expression with local variables.
         ///
         /// \code
-        /// template<typename T, typename List>
+        /// template <typename T, typename List>
         /// using find_index_ = let<
         ///     var<_a, List>,
         ///     var<_b, lazy::find<_a, T>>,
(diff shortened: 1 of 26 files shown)


ericniebler:
Awesome, thanks!

PR #672, opened 2017-06-18
Implement coroutine generator support

CaseyCarter:
* Cmake support for detecting `-fcoroutines-ts` compiler flag and library support in `<experimental/coroutine>`
* Config system support for `RANGES_CXX_COROUTINES`
* Implement coroutine Range generator `ranges::experimental::generator`, and SizedRange generator `ranges::experimental::sized_generator`

### DIFF
--- a/include/range/v3/experimental/utility/generator.hpp
+++ b/include/range/v3/experimental/utility/generator.hpp
@@ -1,0 +1,353 @@
+// Range v3 library
+//
+//  Copyright Casey Carter 2017
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+#ifndef RANGES_V3_EXPERIMENTAL_UTILITY_GENERATOR_HPP
+#define RANGES_V3_EXPERIMENTAL_UTILITY_GENERATOR_HPP
+
+#include <range/v3/detail/config.hpp>
+#if RANGES_CXX_COROUTINES >= RANGES_CXX_COROUTINES_TS1
+#include <cstddef>
+#include <exception>
+#include <utility>
+#include <experimental/coroutine>
+#include <meta/meta.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/view_facade.hpp>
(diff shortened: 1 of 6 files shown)


ericniebler:
Wow! I need to work through the generator code to understand what it's doing. I would also be willing to accept patches to make ranges::optional be more like std::optional.

include/range/v3/experimental/utility/generator.hpp
```diff
@@ -0,0 +1,353 @@
...
+            } // namespace coroutine_owner_
+        } // namespace detail
+        /// \endcond
+
+        namespace experimental
+        {
+            // An owning coroutine_handle
+            template<typename Promise>
+            struct coroutine_owner
+            : std::experimental::coroutine_handle<Promise>
```
CaseyCarter: Dammit - VS Code broke all of these partial indent lines when I moved this stuff in an indent for the `experimental` namespace, and I missed it.

CaseyCarter:
Food for thought: I implemented `coro_fn` which wraps any viewable input range into a coroutine for testing purposes, but didn't provide that tool to users. Does anyone see a use case for such a tool?

Manu343726:
Do you know of any library implementing this generator with boost coroutine or coroutine2? Would you be interested in such a generator backend for this library?

CaseyCarter:
> Do you know of any library implementing this generator with boost coroutine or coroutine2? Would you be interested in such a generator backend for this library?

I don't know of a library implementing this generator atop any flavor of boost coroutine, nor have we had demand for such a thing. (I don't think people are aware that `ranges::experimental::generator` exists.) That said, if there's someone out there that would like to use generators - you really can't beat the ease of range definition - who doesn't have access to TS coroutines, I imagine such an backend would be useful to them.

Manu343726:
I'm asking because I would like to use ranges to filter visitation results, where the visitor is the usual tree recursive visitor:

``` cpp
visit_tree(root, [](const node& node)
{
    ...
});

// with coroutines:

void visit(push_t& yield, const node& root)
{
    visit(root, [&yield](const node& node)
    {
        yield(node);
    });
}

pull_t coro{[&root](push_t& yield)
{
    visit(yield, root);  
}};

auto nodes_named_a = coro | filter([](const node& node)
{
    return node.name == "a";
});
```

In case of boost coroutines, since the pull type implements begin() and end() maybe I could pass a coroutine directly to ranges (Like in the example above) and it would work. I'm not sure if that works directly.

PR #679, opened 2017-06-22
Allow move-only single-pass Views

CaseyCarter:
* Concept changes:
  * relax `View<T>` from `Range<T> && Semiregular<T>` to `Range<T> && Movable<T> && DefaultConstructible<T>`
  * add `Copyable<T>` requirement back into `ForwardView<T>` so that it is unchanged

* add `movesemiregular_t` type alias a la `semiregular_t`
  * does not wrap `T` if `Movable<T> && DefaultConstructible<T>` is satisfied
  * use in `istream_range`, `generate_view`, and `generate_n_view` to allow them to be move-only when `T` is not fully `Semiregular`

* `any_view<T, category::input>` is move-only:
  * `any_view_interface<T, category::input>` has no `clone()` member
  * `any_view_impl::clone` is not an `override` - and may have an ill-formed definition - for `category::input`. (Comment out all uses of `override` to avoid triggering Clang's "inconsistent override" warning.)

* In `view_interface`, perfect-forward `*this` into `operator[]` so that move-only views can use Pythonic slicing.

* Implement a move-only input view test tool: `debug_input_view`
  * catches repeated calls to `begin`, calls to `begin`/`end` on an invalidated view, and iterator operations after invalidation of the view or the iterator.
  * found most of the bugs fixed in my bugs_bugs_bugs PR.
  * Add test cases for each view adaptor that accepts input views

### DIFF
--- a/include/range/v3/utility/polymorphic_cast.hpp
+++ b/include/range/v3/utility/polymorphic_cast.hpp
@@ -5,6 +5,10 @@
 
 #ifndef RANGES_V3_UTILITY_POLYMORPHIC_CAST_HPP
 #define RANGES_V3_UTILITY_POLYMORPHIC_CAST_HPP
+
+#include <memory>
+#include <type_traits>
+#include <meta/meta.hpp>
 #include <range/v3/detail/config.hpp>
 
 namespace ranges
@@ -12,10 +16,19 @@ namespace ranges
     inline namespace v3
     {
         template<typename Target, typename Source>
-        inline Target polymorphic_downcast(Source* x)
+        meta::if_<std::is_pointer<Target>, Target>
+        polymorphic_downcast(Source *x) noexcept
         {
             RANGES_ASSERT(dynamic_cast<Target>(x) == x);
             return static_cast<Target>(x);
+        }
+        template<typename Target, typename Source>
+        meta::if_<std::is_reference<Target>, Target>
+        polymorphic_downcast(Source &&x) noexcept
+        {
+            using PTarget = meta::_t<std::add_pointer<Target>>;
+            auto ptr = polymorphic_downcast<PTarget>(std::addressof(x));
+            return static_cast<Target>(*ptr);
         }
     }
 }
--- a/include/range/v3/utility/polymorphic_cast.hpp
+++ b/include/range/v3/utility/polymorphic_cast.hpp
@@ -16,19 +16,26 @@ namespace ranges
     inline namespace v3
     {
         template<typename Target, typename Source>
-        meta::if_<std::is_pointer<Target>, Target>
-        polymorphic_downcast(Source *x) noexcept
+        auto polymorphic_downcast(Source *x) noexcept ->
+            meta::if_<std::is_pointer<Target>,
+                decltype((static_cast<Target>(x), dynamic_cast<Target>(x)))>
         {
-            RANGES_ASSERT(dynamic_cast<Target>(x) == x);
-            return static_cast<Target>(x);
+            auto result = static_cast<Target>(x);
+            RANGES_ASSERT(dynamic_cast<Target>(x) == result);
+            return result;
         }
         template<typename Target, typename Source>
-        meta::if_<std::is_reference<Target>, Target>
-        polymorphic_downcast(Source &&x) noexcept
+        auto polymorphic_downcast(Source &&x) noexcept ->
+            meta::if_<std::is_reference<Target>,
+                decltype((static_cast<Target>(std::declval<Source>()),
+                    dynamic_cast<Target>(std::declval<Source>())))>
         {
-            using PTarget = meta::_t<std::add_pointer<Target>>;
-            auto ptr = polymorphic_downcast<PTarget>(std::addressof(x));
-            return static_cast<Target>(*ptr);
+            auto &&result = static_cast<Target>(static_cast<Source &&>(x));
+#ifndef NDEBUG
+            auto &&dresult = dynamic_cast<Target>(static_cast<Source &&>(x));
+            RANGES_ASSERT(std::addressof(dresult) == std::addressof(result));
+#endif
+            return static_cast<Target>(result);
         }
     }
 }
--- a/include/range/v3/utility/semiregular.hpp
+++ b/include/range/v3/utility/semiregular.hpp
@@ -195,6 +195,13 @@ namespace ranges
         using semiregular_t =
             meta::if_<SemiRegular<T>, T, semiregular<T>>;
 
+        template<typename T>
+        using movesemiregular_t =
+            meta::if_c<
+                Movable<T>() && DefaultConstructible<T>(),
+                T,
+                semiregular<T>>;
+
         template<typename T, bool IsConst = false>
         using semiregular_ref_or_val_t =
             meta::if_<
--- a/include/range/v3/view/any_view.hpp
+++ b/include/range/v3/view/any_view.hpp
@@ -2,6 +2,7 @@
 // Range v3 library
 //
 //  Copyright Eric Niebler 2014
+//  Copyright Casey Carter 2017
 //
 //  Use, modification and distribution is subject to the
 //  Boost Software License, Version 1.0. (See accompanying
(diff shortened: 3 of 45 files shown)


CaseyCarter:
This is super experimental, but requires invasive changes to the core. Let me know if you want some/all of this behind a `RANGES_BLEEDING_EDGE` macro.  I'm especially concerned about the likelihood of users being broken by `any_view<T, category::input>` no longer being copyable.

ericniebler:
Suggest adding a `unique_any_view<T>` that is move-only+input, and leaving `any_view<T, category::input>` alone.

ericniebler:
Also, `any_view<T, category::input>` and its iterators shouldn't need `clone` members. Since copies are not required to be independent, they should just share their implementation(s) and save a ton of pointless allocations. See the `any_input_iterator` that I contributed to cmcstl2.

CaseyCarter:
> Suggest adding a `unique_any_view<T>` that is move-only+input, and leaving `any_view<T, category::input>` alone.

Done.

> Refactor `any_view` so we don't copy single-pass iterators.

Left to future work. Actually, I'll file an issue for it.

CaseyCarter:
> Also, `any_view<T, category::input>` and its iterators shouldn't need `clone` members. Since copies are not required to be independent, they should just share their implementation(s) and save a ton of pointless allocations.

Ok, I lied: I went ahead and implemented this. Input views now have a completely distinct `any_view` implementation that stores an adapted iterator and sentinel with the view. Cursors are simply pointers to the view erasure interface, and sentinels are `default_sentinel`.

I made a similar change to the forward+ implementation: the adapted view's sentinel is stored alongside it in the implementation object, any_sentinels simply hold pointers to the view erasure interface.

include/range/v3/view/any_view.hpp
```diff
@@ -415,66 +434,47 @@ namespace ranges
...
+        template<typename Ref>
+        using any_forward_view = any_view<Ref, category::forward>;
+
+        template<typename Ref>
+        using any_bidirectional_view = any_view<Ref, category::bidirectional>;
+
+        template<typename Ref>
+        using any_random_access_view = any_view<Ref, category::random_access>;
     }
 }
```
CaseyCarter: That's a lot of red and green to somehow end up at the same total line count.

include/range/v3/utility/polymorphic_cast.hpp
```diff
@@ -5,18 +5,31 @@
...
         {
             RANGES_ASSERT(dynamic_cast<Target>(x) == x);
             return static_cast<Target>(x);
         }
+        template<typename Target, typename Source>
+        meta::if_<std::is_reference<Target>, Target>
+        polymorphic_downcast(Source &&x) noexcept
+        {
+            using PTarget = meta::_t<std::add_pointer<Target>>;
+            auto ptr = polymorphic_downcast<PTarget>(std::addressof(x));
```
ericniebler: `Source` better be a reference for this to make any sense. `static_assert`?

include/range/v3/utility/semiregular.hpp
```diff
@@ -195,6 +195,13 @@ namespace ranges
         using semiregular_t =
             meta::if_<SemiRegular<T>, T, semiregular<T>>;
 
+        template<typename T>
+        using movesemiregular_t =
+            meta::if_c<
+                Movable<T>() && DefaultConstructible<T>(),
+                T,
+                semiregular<T>>;
```
ericniebler: I hate the name `movesemiregular`. I don't have a better suggestion.

include/range/v3/view/any_view.hpp
```diff
@@ -102,155 +176,106 @@ namespace ranges
...
-            public:
                 any_sentinel() = default;
-                template<typename Rng,
-                    CONCEPT_REQUIRES_(!Same<detail::decay_t<Rng>, any_sentinel>()),
-                    CONCEPT_REQUIRES_(InputRange<Rng>())>
-                any_sentinel(Rng &&rng, end_tag)
-                  : ptr_{new any_sentinel_impl<sentinel_t<Rng>, iterator_t<Rng>>{
-                        end(rng)}}
+                constexpr explicit any_sentinel(any_sentinel_interface const &view) noexcept
+                  : view_{&view}
```
ericniebler: "`view`" is a strange name for a sentinel.

include/range/v3/view/any_view.hpp
```diff
@@ -304,43 +328,53 @@ namespace ranges
 
             template<typename Ref, category Cat>
             struct any_view_interface
+              : any_sentinel_interface
```
ericniebler: Ah, a view IS-A sentinel in the OO sense. This is a bit nasty. I would prefer a HAS-A relationship here, but if it would uglify the code, then leave it.

include/range/v3/utility/polymorphic_cast.hpp
ericniebler: Well, I suppose someone might want to `polymorphic_downcast<Derived&&>(std::move(base))`. But I would prefer to write `std::move(polymorphic_downcast<Derived&>(base))`, and have a safer `polymorphic_downcast`.

include/range/v3/utility/semiregular.hpp
CaseyCarter: Same. Hate it utterly, have no better idea, was sure you would know what this name means.

include/range/v3/utility/polymorphic_cast.hpp
CaseyCarter: It's intentional that `polymorphic_downcast` can transform rvalues into rvalues; it's NOT intentional that it unsafely turns rvalues into lvalues. I'll fix this.

include/range/v3/view/any_view.hpp
CaseyCarter: `any_sentinel_interface` is an even-more-erased base for `any_view_interface` that allows a single `any_sentinel` type to be used with all forward+ `any_view`s. `any_cursor`s don't know where the `view` is, but `any_sentinel`s do. `any_cursor::equal(any_sentinel)` yanks the `any_sentinel_interface` out of the `any_sentinel` and passes an erased reference to the adapted iterator into `at_end`. `any_cursor::equal(any_sentinel)` roughly answers the question "does this cursor's wrapped iterator point at the end of that `any_sentinel`'s range."

The fact that I have to explain all of this drives home what I already knew: the name `any_sentinel_interface` is utterly and completely wrong.

include/range/v3/utility/semiregular.hpp
CaseyCarter: `Demiregular`? [`Demisemiquaver`](https://www.youtube.com/watch?v=sLm2jI8JH0c)?

include/range/v3/view/any_view.hpp
CaseyCarter: > `any_cursor`s don't know where the `view` is, but `any_sentinel`s do. 

Yes, this is ironically opposite the case for the input view implementation: `any_input_view_cursor` (which I should really rename to `any_input_cursor`) holds a pointer to the view which `default_sentinel` obviously does not.

include/range/v3/view/any_view.hpp
CaseyCarter: The first pass I made at this didn't store a sentinel in the `any_view_impl`: `at_end` just passed the view to `ranges::end` to get a sentinel on-demand to compare with its iterator argument. This was great, except for the fact that `at_end` MUST be `const` and not all views have `end() const`. This is when I realized there's only ever one sentinel value for an `any_view`, so I could simply store it inside the view's type erasure. (And the obvious lemma for input views - there's only ever one iterator - occurred to me soon after.)

I like that the sentinel and view "share" a type erasure instead of being separately erased. It simplifies things, and avoids having another vtable pointer. Plus I can bring back the "don't store sentinels at all" approach as an optimization for views with `end() const`. As a matter of fact, I'll do that now.

include/range/v3/utility/polymorphic_cast.hpp
CaseyCarter: >  it's supposed to mimic the behavior of `dynamic_cast`, after all

No: it's supposed to mimic the intersection of the behaviors of `dynamic_cast` and `static_cast`.

ericniebler:
Travis failures, any_view with gcc-4.9.

ericniebler:
Although I still have misgivings about move-only views, I think this is a good experiment to try.

CaseyCarter:
I think it's significant that I spent more time fixing bugs that I found (improper uses of input views) than updating uses to allow for move-only views. any_view was the only real sticking point.

PR #683, opened 2017-06-25
control use of <thread> with the RANGES_CXX_<feature> system

### LINKED ISSUE #682: Check for thread support
Currently utility/random.hpp has this:

```C++
#ifndef __MINGW32__
                    // Hash our thread id.  It seems to vary from run to run on OS X, not
                    // so much on Linux.
                    *it++ = randutils::hash(std::this_thread::get_id());
#endif
```

However, this doesn't take into account systems that don't have thread support (i.e, cross compiling for a homebrew kernel). Patching it out is trivial, but I think pedantically there should be a CMake check at least. Potentially with `try_compile` since libstdc++ for example has this macro define:

`#if defined(_GLIBCXX_HAS_GTHREADS) && defined(_GLIBCXX_USE_C99_STDINT_TR1)`

and doesn't define `thread` otherwise, which means checking for existence of `<thread>` and this platform-specific macro is not sufficient.

CaseyCarter:
Cmake uses `try_compile` to determine if `std::this_thread::get_id` is usable; disables `<thread>` usage if not by defining `RANGES_CXX_THREAD=0` on the command line.

CI scripts install *exactly* the version of cmake in `cmake_minimum_required` so we do not inadvertently use a newer feature without updating it.

Update the `RANGES_CXX_<feature>` database for C++17.

Fixes #682.


remyabel:
This is a *very minor* platform specific issue but thought I should point out that libstdc++, given the presence of glibc, *always* [returns `thread::id(1)`](https://github.com/gcc-mirror/gcc/blob/master/libstdc++-v3/include/std/thread#L338) or the hash 2292942703.

You could use `getpid()` or `syscall (SYS_getpid)` (`getpid` supposedly has caching issues but I haven't observed it) if you care about this granular level of platform support.

CaseyCarter:
> This is a very minor platform specific issue but thought I should point out that libstdc++, given the presence of glibc, always returns `thread::id(1)` or the hash 2292942703.

Only in single-threaded programs; it would not be conforming to do so in multithreaded programs. 

> You could use `getpid()` or `syscall (SYS_getpid)` (`getpid` supposedly has caching issues but I haven't observed it) if you care about this granular level of platform support.

The point of this code in random.hpp is to sample a lot of things that are cheap and may include a few bits of entropy in hopes of gathering enough total entropy to provide a good seed for the default uniform random bit generator just in case `std::random_device` is deterministic. It doesn't matter that some of the sources may have no entropy at all on some implementations.

ericniebler:
Thanks!

Issue #658, opened 2017-05-02
`coerce` in `partial_sum`


khlebnikov:
The `coerce` in `partial_sum` is used to [initialize the accumulator](https://github.com/ericniebler/range-v3/blob/e411a19a312542be98ec9f318ef5b335e0fdaf0a/include/range/v3/numeric/partial_sum.hpp#L60), but not [when it is updated](https://github.com/ericniebler/range-v3/blob/e411a19a312542be98ec9f318ef5b335e0fdaf0a/include/range/v3/numeric/partial_sum.hpp#L65). 

Is this a bug? If yes, I'll fix this along with some other changes to `partial_sum`.

CaseyCarter:
I think the uses of `coerce` are bogus, and the constraints are wrong:

```c++
template<typename I, typename O, typename BOp = plus, typename P = ident,
    typename X = concepts::Invocable::result_t<P&, reference_t<I>>,
    typename Y = concepts::Invocable::result_t<BOp&, uncvref_t<X>&, X>>
using PartialSummable = meta::strict_and<
    InputIterator<I>,
    OutputIterator<O, uncvref_t<X>&>,
    Invocable<P&, reference_t<I>>,
    Constructible<uncvref_t<X>, X>,
    Invocable<BOp&, uncvref_t<X>&, X>,
    Assignable<uncvref_t<X>&, Y>>;

template<typename I, typename S, typename O, typename S2,
    typename BOp = plus, typename P = ident,
    CONCEPT_REQUIRES_(Sentinel<S, I>() && Sentinel<S2, O>() &&
        PartialSummable<I, O, BOp, P>())>
tagged_pair<tag::in(I), tag::out(O)>
operator()(I begin, S end, O result, S2 end_result, BOp bop = BOp{}, P proj = P{}) const
{
    if(begin != end && result != end_result)
    {
        auto t(invoke(proj, *begin));
        *result = t;
        for(;;)
        {
            ++begin;
            ++result;
            if(begin == end || result == end_result) break;
            t = invoke(bop, t, invoke(proj, *begin));
            *result = t;
        }
    }
    return {detail::move(begin), detail::move(result)};
}
```

khlebnikov:
Yeah, the constraints looked weird to me as well (I'm working on `inclusive_scan` which is basically `partial_sum` in the no-init sequential version). Should I bundle these changes in the same pull request?

CaseyCarter:
[...]

PR #688, opened 2017-07-02
better constraints and implementation for partial_sum, refs #658

ericniebler:


include/range/v3/numeric/partial_sum.hpp
```diff
@@ -50,19 +48,21 @@ namespace ranges
             tagged_pair<tag::in(I), tag::out(O)>
             operator()(I begin, S end, O result, S2 end_result, BOp bop = BOp{}, P proj = P{}) const
```
CaseyCarter: Pre-existing: `S` and `S2` is weird. We should rename `S` to `S1` while we're in here.

include/range/v3/numeric/partial_sum.hpp
```diff
@@ -30,16 +30,14 @@ namespace ranges
     inline namespace v3
     {
         template<typename I, typename O, typename BOp = plus, typename P = ident,
-            typename V = value_type_t<I>,
-            typename X = concepts::Invocable::result_t<P&, V>,
-            typename Y = concepts::Invocable::result_t<BOp&, X, X>>
+            typename X = projected<projected<I, coerce<value_type_t<I>>>, P>,
+            typename Y = reference_t<X>>
         using PartialSummable = meta::strict_and<
+            SemiRegular<detail::decay_t<Y>>,
```
CaseyCarter: `SemiRegular` is ridiculously overconstrained here. This should be `Constructible<detail::decay_t<Y>, Y>`.

include/range/v3/numeric/partial_sum.hpp
```diff
@@ -30,16 +30,14 @@ namespace ranges
...
             InputIterator<I>,
-            OutputIterator<O, X>,
-            Invocable<P&, V>,
-            CopyConstructible<uncvref_t<X>>,
-            Invocable<BOp&, X, X>,
-            Assignable<uncvref_t<X>&, Y>>;
+            OutputIterator<O, detail::decay_t<Y>&>,
```
CaseyCarter: `detail::decay_t<Y> const &` - writing the accumulator to the output must not modify it.

include/range/v3/numeric/partial_sum.hpp
```diff
@@ -30,16 +30,14 @@ namespace ranges
...
+            IndirectRegularInvocable<BOp, detail::decay_t<Y>*, X>,
+            Assignable<detail::decay_t<Y>&, indirect_result_of_t<BOp&(detail::decay_t<Y>*, X)>>>;
```
CaseyCarter: `s/detail::decay_t<Y>*/detail::decay_t<Y>&/g`

ericniebler: I don't think so. This is `indirect_result_of`.

CaseyCarter: Ugh, sorry.

ericniebler:
Now I'm less certain about the use of `coerce`. The other algorithms don't coerce to the value type before applying the projection. This still needs more thought.

include/range/v3/numeric/partial_sum.hpp
ericniebler: Why?

CaseyCarter: I suppose because I would like the output sequence to depend only on the input sequence and the binary operation, and not depend on the choice of output iterator.

include/range/v3/numeric/partial_sum.hpp
```diff
@@ -50,19 +48,21 @@ namespace ranges
...
-
+                using X = projected<projected<I, coerce<value_type_t<I>>>, P>;
+                using Y = reference_t<X>;
+                coerce<value_type_t<I>> v;
                 if(begin != end && result != end_result)
                 {
-                    auto t(x(invoke(proj, v(*begin))));
+                    auto &&tmp1 = *begin;
+                    auto &&tmp2 = v((decltype(tmp1) &&) tmp1);
+                    detail::decay_t<Y> t(invoke(proj, (decltype(tmp2) &&) tmp2));
```
CaseyCarter: What does the `tmp1`/`tmp2` schtick buy us here? If `Y` is a proxy reference type the access to `t` in the loop (after incrementing `begin`) is UB regardless of whether or not we bind these references here.

ericniebler: In theory, passing an actual value (as opposed to a proxy) to a projection should yield another honest-to-god value, as opposed to a proxy reference. Part of my reasoning for requiring `SemiRegular<decay_t<Y>>` is to get closer to enforcing that. `Regular`-ish types should have independent copies.

include/range/v3/numeric/partial_sum.hpp
ericniebler: I want to say something here about `detail::decay_t<Y>` begin a type that can be mutated without bizarre, non-local side-effects. I suppose `SemiRegular` doesn't do that. Maybe it should.

khlebnikov: It seems like trying to constrain the behavior to avoid such 'evil' types is impossible. How would you check for lack of side effects? Perhaps it makes no sense for range library to try and prevent all the potential abuses?

include/range/v3/numeric/partial_sum.hpp
khlebnikov: I think that coerce is at all necessary because we need a type for "accumulator" in the numeric algorithms. The goal is not to allow the proxy reference to be used as this type. So it seems that using `coerce` to `value_type` of the iterator should prevent this - if the projection does not itself contain some kind of UB, it has no access to the original proxy and thus cannot return it back to us. 
Then, decaying the result should give us a safe accumulator type (again, barring 'evil' types which have some weird side-effects). With this logic, the dance with storing references to extend some temporary's lifetimes (this is the original intent, is it not?) indeed seems unnecessary.

khlebnikov: I guess all this logic was all there originally :)

include/range/v3/numeric/partial_sum.hpp
ericniebler: > Perhaps it makes no sense for range library to try and prevent all the potential abuses?

That isn't an argument against a range library trying to prevent _some_ abuses. @CaseyCarter, how strenuously are you objecting to `Semiregular` here? I like strong constraints.

include/range/v3/numeric/partial_sum.hpp
ericniebler: Meh. I don't feel strongly about it, and I don't particularly care enough to require implementers to `*out = as_const(acc);`. I'll change it if you feel strongly.

CaseyCarter: `OutputIterator<O, T const&>` activates the "implicit expression variants" wording and therefore requires `*out = acc;` to also be valid and not modify `acc`. (Irony: if writing through the output iterator might modify the accumulator, then an implementation must output a copy of the accumulator which would roughly require `SemiRegular` ;)

include/range/v3/numeric/partial_sum.hpp
CaseyCarter: Why not make it `Regular`? Or `Regular && StreamInsertable`? Constraining the arguments beyond the requirements of the algorithm is counterproductive. This implementation does four things with `detail::decay_t<Y>`:
* constructs an object of that type from a `Y`
* writes an lvalue through an `O`
* assigns an lvalue from the result of invoking an lvalue `BOp` with /*...*/
* destroys the object

I cannot envision a more efficient implementation of this algorithm that would make use of `SemiRegular<detail::decay_t<Y>>`.

ericniebler:
What we really want to say for `partial_sum` is that the binary operation and the result of the projection form a `Semigroup`. That would require `BOp` to be associative, which might be overconstraining. For instance, someone might want to use `partial_sum` with `plus<>` and `std::string`. I'm not sure how much I care about that. If users want the crappy non-associative `partial_sum`, they could always use the one in namespace `std`.

We really need to go through _all_ the numeric algorithms and constrain them with the appropriate mathematical concept from ring theory. Then we need appropriate refinements: `Semigroup`, `Monoid`, etc.

(And now I'm wondering if we actually need `IndirectSemigroup` and friends in order to accommodate proxy iterators. Blech.)

ericniebler:
I'm slowly convincing myself that the coercions are necessary to keep this algorithm meaningful, especially in the presence of proxy iterators and projections. I don't think the use of `coerce` in this algorithm introduces any inefficiencies for non-proxy iterators.

ericniebler:
I'm finally (mostly) happy with these constraints. I think it gives us a direction for the other numeric algorithms.

Issue #686, opened 2017-07-01
Feature request: Add `at` to view_interface template struct


apmccartney:
Random-access view_interface instantiations doesn't seem to provide a bounds-checking `at` method as vector does. Vector-api compatability is helpful when introducing range-v3 to an existing generic codebase. 

I was considering adding this capability and filing a pull request, but wanted to verify there wasn't a reason I ought not do so.

ericniebler:
Seems reasonable. Go for it!

Issue #680, opened 2017-06-23
-Wconversion triggering errors when using gcc-7.1


cpockrandt:
Hi,

I moved from using the latest gcc-6 to gcc-7.1 and noticed that the additional flag `-Wconversion` causes a couple of errors in your library which worked perfectly fine with previous gcc versions.

Do you unofficially support this flag or was it kind of a "happy accident" that it used to work before? 😃 All tests succeeded and with the newest gcc-7 only a couple are failing with the same error message.

Here's a minimal example:

    std::vector<char> xs {'A', 'B'};
    std::vector<int> ys {1, 2};
    for (auto [ x, y ] : ranges::view::zip(xs, ys))
        std::cout << x << ", " << y << std::endl;

Output:

    ../range-v3/include/range/v3/utility/common_type.hpp:76:45: warning: choosing ‘ranges::v3::common_pair<F, S>::operator std::pair<_U1, _U2>() const & [with F2 = char; S2 = int; int _concept_requires_276 = 42; typename std::enable_if<((_concept_requires_276 == 43) || (ranges::v3::concepts::models<ranges::v3::concepts::Constructible, F2, const F&>() && ranges::v3::concepts::models<ranges::v3::concepts::Constructible, S2, const S&>())), int>::type <anonymous> = 0; F = char&; S = int&]’ over ‘constexpr std::pair<_T1, _T2>::pair(const std::pair<_U1, _U2>&) [with _U1 = char&; _U2 = int&; typename std::enable_if<(std::_PCC<((! std::is_same<_T1, _U1>::value) || (! std::is_same<_T2, _U2>::value)), _T1, _T2>::_ConstructiblePair<_U1, _U2>() && std::_PCC<((! std::is_same<_T1, _U1>::value) || (! std::is_same<_T2, _U2>::value)), _T1, _T2>::_ImplicitlyConvertiblePair<_U1, _U2>()), bool>::type <anonymous> = 1; _T1 = char; _T2 = int]’ [-Wconversion]
             using _cond_res = decltype(true ? std::declval<T>() : std::declval<U>());
                                        ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ../range-v3/include/range/v3/utility/common_type.hpp:76:45: warning:   for conversion from ‘’ to ‘const std::pair<char, int>’ [-Wconversion]
[...]

PR #689, opened 2017-07-02
View interface at

apmccartney:
I've added the relative tests to `index.cpp`. If there was a more appropriate location, please let me know and I'll make the corresponding corrections.

### DIFF
--- a/include/range/v3/at.hpp
+++ b/include/range/v3/at.hpp
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-                if( (n < 0) or (n >= ranges::distance(rng)) )
+                if(n < 0 || n >= ranges::distance(rng))
                 {
                     throw std::out_of_range("ranges::at");
                 }
--- a/include/range/v3/at.hpp
+++ b/include/range/v3/at.hpp
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-                if(n >= ranges::distance(rng))
+                if( (n >= ranges::distance(rng)) or (n < 0) )
                 {
                     throw std::out_of_range("ranges::at");
                 }
--- a/include/range/v3/at.hpp
+++ b/include/range/v3/at.hpp
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-                if( (n >= ranges::distance(rng)) or (n < 0) )
+              if( (n < 0) or (n >= ranges::distance(rng)) )
                 {
                     throw std::out_of_range("ranges::at");
                 }
--- a/include/range/v3/at.hpp
+++ b/include/range/v3/at.hpp
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-              if( (n < 0) or (n >= ranges::distance(rng)) )
+                if( (n < 0) or (n >= ranges::distance(rng)) )
                 {
                     throw std::out_of_range("ranges::at");
                 }
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -306,7 +306,7 @@ namespace ranges
                 using size_type = range_size_type_t<Derived>;
                 if (n < 0 || size_type(n) >= ranges::size(derived()))
                 {
-                  throw std::out_of_range("view_interface::at");
+                    throw std::out_of_range("view_interface::at");
                 }
                 return derived().begin()[n];
             }
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -299,16 +299,16 @@ namespace ranges
             }
             /// Returns a reference to the element at specified location pos, with bounds checking.
             template<typename D = Derived,
-                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>())>
-            auto at( range_difference_type_t<D> n) ->
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>() && SizedRange<D>())>
+            auto at(range_difference_type_t<D> n) ->
                 decltype(std::declval<D &>().begin()[n])
             {
-                using size_type = decltype( derived().size() );
-                if ( (n < 0) or
-                     (size_type(n) >= derived().size()) ){
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
                   throw std::out_of_range("view_interface::at");
                 }
-                return derived()[n];
+                return derived().begin()[n];
             }
             /// Implicit conversion to something that looks like a container.
             template<typename Container, typename D = Derived,
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -310,6 +310,19 @@ namespace ranges
                 }
                 return derived().begin()[n];
             }
+            /// \overload
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D const>() && SizedRange<D const>())>
+            auto at(range_difference_type_t<D> n) const  ->
+                decltype(std::declval<D const &>().begin()[n])
+            {
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
+                    throw std::out_of_range("view_interface::at");
+                }
+                return derived().begin()[n];
+            }
             /// Implicit conversion to something that looks like a container.
             template<typename Container, typename D = Derived,
                 typename = typename Container::allocator_type, // HACKHACK
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -303,7 +303,9 @@ namespace ranges
             auto at( range_difference_type_t<D> n) ->
                 decltype(std::declval<D &>().begin()[n])
             {
-              if ( (n < 0) or (n >= derived().size()) ){
+              using size_type = decltype( derived().size() );
+              if ( (n < 0) or
+                   (size_type(n) >= derived().size()) ){
                 throw std::out_of_range("view_interface::at");
               }
               return derived()[n];
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -297,6 +297,17 @@ namespace ranges
             {
                 return Slice{}(detail::move(derived()), offs.from, offs.to);
             }
+            /// Returns a reference to the element at specified location pos, with bounds checking.
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>())>
+            auto at( range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+              if ( (n < 0) or (n >= derived().size()) ){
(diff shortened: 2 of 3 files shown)


include/range/v3/at.hpp
```diff
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-                if(n >= ranges::distance(rng))
+                if( (n < 0) or (n >= ranges::distance(rng)) )
```
apmccartney: Note that, unlike its standard library analogue, this method accepts a `difference_type` (which is typically signed) rather than a a `size_type` (which is typically unsigned). Consequently, extended the verification to ensure the queried index is non-negative.

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
...
             }
+            /// Returns a reference to the element at specified location pos, with bounds checking.
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>())>
+            auto at( range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+                using size_type = decltype( derived().size() );
+                if ( (n < 0) or
+                     (size_type(n) >= derived().size()) ){
```
apmccartney: Note that we've already verified `n` is non-negative. Therefore the `size_type` conversion is safe.

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
             {
                 return Slice{}(detail::move(derived()), offs.from, offs.to);
             }
+            /// Returns a reference to the element at specified location pos, with bounds checking.
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>())>
+            auto at( range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+                using size_type = decltype( derived().size() );
```
CaseyCarter: You haven't required `SizedRange<D>()`, so this may fail to instantiate. Once you have required `SizedRange`, you can simply use `range_size_type_t<Derived>` for its size type.

include/range/v3/at.hpp
CaseyCarter: range-v3 prefers to use the primary tokens (`||`) over their alternatives (`or`). We also don't typically parenthesize subexpressions unnecessarily or use spaces-inside-parens.

include/range/v3/view_interface.hpp
CaseyCarter: Ditto `or` and `(n < 0)` and `( foo )`. I would use `ranges::size(derived())` instead of `derived().size()` since that's the expression that `SizedRange` requires to be valid. (It's possible that the default member implementations of `size` don't work and some user implements non-member `size` for their view.)

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
...
+                if ( (n < 0) or
+                     (size_type(n) >= derived().size()) ){
+                  throw std::out_of_range("view_interface::at");
+                }
+                return derived()[n];
```
CaseyCarter: `RandomAccessRange` doesn't require a `[]` operator or require that `derived()[n]` has the same type as `derived().begin()[n]`; this should return `derived().begin()[n]`. (A user may have overriden or disabled `[]` in `Derived`.)

test/index.cpp
```diff
@@ -38,6 +38,38 @@ int main()
...
                                 ranges::view::c_str("ranges::at")));
         }
+
+        try {
+            ranges::at(vi, -1);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("ranges::at")));
+
```
CaseyCarter: Please remove this empty line at the end of the block.

test/index.cpp
```diff
@@ -38,6 +38,38 @@ int main()
...
+                                ranges::view::c_str("view_interface::at")));
+        }
+
+        try {
+            viv.at(-1);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("view_interface::at")));
+
```
CaseyCarter: Ditto empty line.

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
             {
                 return Slice{}(detail::move(derived()), offs.from, offs.to);
             }
+            /// Returns a reference to the element at specified location pos, with bounds checking.
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>())>
+            auto at( range_difference_type_t<D> n) ->
```
CaseyCarter: Ditto space inside parens.

test/index.cpp
```diff
@@ -38,6 +38,38 @@ int main()
...
+        try {
+            ranges::at(vi, -1);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("ranges::at")));
+
+        }
+
+        auto viv = ranges::make_iterator_range( vi.begin(), vi.end() );
```
CaseyCarter: Ditto spaces-inside-parens.

include/range/v3/view_interface.hpp
CaseyCarter: This block also needs to be Allman style and indented 4 spaces:
```c++
if (stuff)
{
    do_things();
}
```

apmccartney:
Thank you for the detailed review Casey. I believe I've made the requested corrections.

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
...
+            /// Returns a reference to the element at specified location pos, with bounds checking.
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>() && SizedRange<D>())>
+            auto at(range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
+                  throw std::out_of_range("view_interface::at");
```
CaseyCarter: This line needs two more spaces of indentation.

apmccartney: Whoops, my editor is working against me. I'll correct this and double check for other instances.

apmccartney:
Closes #686

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,19 @@ namespace ranges
...
+            auto at(range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
+                    throw std::out_of_range("view_interface::at");
+                }
+                return derived().begin()[n];
+            }
```
ericniebler: This needs a `const` overload, and maybe make them lvalue ref qualified.

apmccartney: > This needs a const overload

Well that's embarrassing. I apologize for the oversight. Correction incoming.

> maybe make them lvalue ref qualified

C++ 14 has spoiled me. Is it not the case that the trailing return type on line 304 would respect const-ness, ref-ness, and violatile-ness? Or have I misunderstood your feedback?

```cpp
decltype(std::declval<D &>().begin()[n])
```

apmccartney: Extended tests incoming

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,32 @@ namespace ranges
...
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D>() && SizedRange<D>())>
+            auto at(range_difference_type_t<D> n) ->
+                decltype(std::declval<D &>().begin()[n])
+            {
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
+                    throw std::out_of_range("view_interface::at");
+                }
+                return derived().begin()[n];
```
FloopCZ: I suggest explicit cast of `n` to `size_type` to avoid some `-Wconversion` signedness warnings. The same on L304. (relates #680, gcc would probably stay quiet, but clang would produce a warning)

include/range/v3/view_interface.hpp
```diff
@@ -297,6 +297,32 @@ namespace ranges
...
+                CONCEPT_REQUIRES_(Same<D, Derived>() && RandomAccessRange<D const>() && SizedRange<D const>())>
+            auto at(range_difference_type_t<D> n) const  ->
+                decltype(std::declval<D const &>().begin()[n])
+            {
+                using size_type = range_size_type_t<Derived>;
+                if (n < 0 || size_type(n) >= ranges::size(derived()))
+                {
+                    throw std::out_of_range("view_interface::at");
+                }
+                return derived().begin()[n];
```
FloopCZ: ditto

include/range/v3/view_interface.hpp
apmccartney: Strangely, while working with Clang 4.0.0, adding an explicit conversion (see below) results in `-Wsign-conversion` warning and `-Werror` errors out the compilation. 

```cpp
return derived().begin()[size_type(n)]
```

FloopCZ: That is strange. Are you sure the warning is produced on the line with the explicit conversion? Isn't it from somewhere else? I suppose the range-v3 library may have a few places where it implicitly changes signedness.

[This](https://gist.github.com/FloopCZ/f9ff686883056096b28a09f25be52dc4) seems to work as expected.

apmccartney: Note that we're applying the bracket operator on a random access iterator. As a result, it _should_ accept a difference type rather than a size type. Consider the following:

```cpp
int main() {
    std::vector<int> v{1,2,3,4};
    assert( v[2] == v.end()[-2] );
 }
```
Consider what whould happen here if the bracket operator accepted a size type rather than a difference type.

FloopCZ: You are right, it _should_ accept a difference type so it should not produce a warning. Good point. One more reason to check the location of the `-Wconversion` errors.

test/index.cpp
```diff
@@ -38,6 +38,58 @@ int main()
...
+                                ranges::view::c_str("view_interface::at")));
+        }
+
+        try {
+            cviv.at(-1);
+            CHECK(false);
+        } catch(std::out_of_range const& e) {
+            CHECK(ranges::equal(ranges::view::c_str(e.what()),
+                                ranges::view::c_str("view_interface::at")));
+        }
```
ericniebler: [Allman-style braces](https://en.wikipedia.org/wiki/Indent_style#Allman_style) here please.

include/range/v3/at.hpp
```diff
@@ -40,7 +40,7 @@ namespace ranges
             auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const ->
                 decltype(begin(rng)[n])
             {
-                if(n >= ranges::distance(rng))
+                if(n < 0 || n >= ranges::distance(rng))
```
ericniebler: 👍

test/index.cpp
apmccartney: To clarify, do you prefer
(a)
```cpp
try
{
    ...
}
catch(...)
{
    ...
}
```
or (b)
```cpp
try {
    ...
} catch(...) 
{
    ...
}
```
I'll go ahead and update the existing tests as well

ericniebler: From Wikipedia:
> This style puts the brace associated with a control statement on the next line, indented to the same level as the control statement.

So this:

```c++
try
{
    ...
}
catch(...)
{
    ...
}
```

Just update your own tests. If there are other tests that are ill-formatted, leave them and we can update them in a separate PR.

Thanks.

PR #690, opened 2017-07-02
Inclusive scan

khlebnikov:
It's quite interesting how I found some more time to work on ranges exactly after you've made the change regarding coerce. 

I've rebased onto partial_sum branch. Several things here:
1) I've reworked the way the tests are done to reduce some duplication
2) I wanted to use partial_sum as implementation for no-init version so I've extracted "detail::partial_sum_no_check"
3) I've given more descriptive names to 'X' and 'Y' in the partial_sum concept
4) inclusive_scan itself. The concept is based on the description in n4659.

When we find the consensus on inclusive_scan, exclusive_scan will be a breeze.

Rostislav.

### DIFF
--- a/include/range/v3/numeric/inclusive_scan.hpp
+++ b/include/range/v3/numeric/inclusive_scan.hpp
@@ -1,0 +1,168 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Rostislav Khlebnikov 2017
+//  Copyright Eric Niebler 2013-2014
+//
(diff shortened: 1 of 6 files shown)


khlebnikov:
Oh, Travis errored in c++11 build. I'll fix this.

khlebnikov:
Also have to fix the potential dangling reference that I've introduced in the same place. I'll close for now and then resubmit

PR #699, opened 2017-07-15
Fix #671 - CMakeLists.txt when including it from a parent directory

### LINKED ISSUE #671: CMakeLists.txt - Error when using add_subdirectory(range-v3) from CMakeLists.txt of parent project
On `master`, commit 6dba64d0073ffeb8e33e643e1e8e32be70dee935

When range-v3 was in a subdirectory of a project foo and `add_subdirectory(range-v3)` is called from the CMakeLists.txt of project foo, an error message is raised when running cmake:
```
CMake Error at range-v3/cmake/TestHeaders.cmake:93 (message):
The `HEADERS` argument must be provided.
range-v3/CMakeLists.txt:103 (generate_standalone_header_tests)
```
`RANGE_V3_PUBLIC_HEADERS` is empty, because the variable `CMAKE_SOURCE_DIR` is used.
See [bug-CMakeLists.zip](https://github.com/ericniebler/range-v3/files/1082412/bug-CMakeLists.zip)

Using `CMAKE_CURRENT_SOURCE_DIR` instead fixes the issue (see attached patch file 
[range-v3_FixCMakeLists.txt](https://github.com/ericniebler/range-v3/files/1082413/range-v3_FixCMakeLists.txt)).

Marc--Olivier:
When range-v3 was in a subdirectory of a project foo and
add_subdirectory(range-v3) was called from the CMakeLists.txt
of project foo, an error message was raised when running cmake:
""
CMake Error at range-v3/cmake/TestHeaders.cmake:93 (message):
  The `HEADERS` argument must be provided.
  range-v3/CMakeLists.txt:103 (generate_standalone_header_tests)
""
RANGE_V3_PUBLIC_HEADERS was empty, because the variable
CMAKE_SOURCE_DIR was used.
Using CMAKE_CURRENT_SOURCE_DIR instead fixed the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,12 +104,12 @@ endif()
 
 # Test all headers
 file(GLOB_RECURSE RANGE_V3_PUBLIC_HEADERS
-                  RELATIVE "${CMAKE_SOURCE_DIR}/include"
-                  "${CMAKE_SOURCE_DIR}/include/*.hpp")
+                  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
+                  "${CMAKE_CURRENT_SOURCE_DIR}/include/*.hpp")
 
 # Add header files as sources to fix MSVS 2017 not finding source during debugging
 file(GLOB_RECURSE RANGE_V3_PUBLIC_HEADERS_ABSOLUTE
-                  "${CMAKE_SOURCE_DIR}/include/*.hpp")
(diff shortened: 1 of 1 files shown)


ericniebler:
@tahonermann You had some objections to this direction in #671. I'm not a CMake guy. What are the arguments here?

tahonermann:
My comment wasn't intended as an objection, but rather a suggestion that another approach might be more inline with how CMake intends projects to be composed.  I don't see anything to be concerned about with Marc's changes.  I'm not a CMake guy either.

PR #703, opened 2017-07-20
fix view::group_by weirdness

ericniebler:


test/view/group_by.cpp
```diff
@@ -86,5 +86,13 @@ int main()
         check_equal(*rng.begin(), {1, 3, 5});
     }
 
+    {
+        std::vector<int> v {0,1,2,3,4,5,6,7,8,9};
+        auto rng0 = ranges::view::group_by(v, [](int i, int j){ return j - i < 3; });
+        check_equal(*rng0.begin(), {0, 1, 2});
+        check_equal(*next(rng0.begin()), {3, 4, 5});
+        check_equal(*next(rng0.begin(), 2), {6, 7, 8});
+        check_equal(*next(rng0.begin(), 3), {9});
```
FloopCZ: (nitpicking) For the sake of completeness, you may consider adding something like:
```c++
CHECK(next(rng0.begin(), 4) == rng0.end());
```

PR #709, opened 2017-08-11
Fix broken link to N4128 document

unwiredben:
Markdown was missing http:// prefix, so acted like a local page ref


ericniebler:
Thanks! Apologies for the lag. I've been out on vacation.

PR #711, opened 2017-08-17
Span updates

CaseyCarter:
Changes to span, its support machinery, and the library's `Contiguous` concepts upon which span relies.

See the individual commit messages for detailed change descriptions.


include/range/v3/range_fwd.hpp
```diff
@@ -393,12 +393,11 @@ namespace ranges
         struct ordered_less;
         struct ident;
 
-        enum cardinality
+        enum cardinality : std::ptrdiff_t
```
ericniebler: Neat. I thought only enum classes could have an enum base.

gnzlbg: AFAIK This is not an enum base (as in, inheriting cases from another enum) but just stating that the underlying storage of the enum will be of type `std::ptrdiff_t` (and hence `sizeof(cardinality) == sizeof(std::ptrdiff_t)`).

CaseyCarter: The two features - enum classes and explicit underlying types - both came into the language at the same time, so people often conflate them.

> AFAIK This is not an enum base ...
The actual name of the grammar non-terminal in the Standard that matches an explicit underlying type declaration is [*enum-base*](http://eel.is/c++draft/dcl.enum#nt:enum-base).

PR #712, opened 2017-08-17
Invalid sentinel type in the iter_transform2_view::sentinel constructor (second range)

BigSeb:
The sentinel type for the second range in the constructor of the sentinel class of the iter_transform2_view is in the wrong type (reference to the first range instead of the second range).

This is similar to another fix made to the transform view (#267)


ericniebler:
Oops, thanks.

PR #716, opened 2017-08-26
Inclusive scan

khlebnikov:
Concept is still the same (based on the Standard) - there's no clear analogy for semigroup when the accumulator type is different from value type.

### DIFF
--- a/include/range/v3/numeric/inclusive_scan.hpp
+++ b/include/range/v3/numeric/inclusive_scan.hpp
@@ -1,0 +1,164 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Rostislav Khlebnikov 2017
+//  Copyright Eric Niebler 2013-2014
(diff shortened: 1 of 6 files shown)


ericniebler:
I haven't forgotten about this. Day job has kept me busy. I suspect there are more meaningful concepts to be found here, I just haven't had time to sit down with the algorithm to confirm.

khlebnikov:
I certainly understand. I'm sure there is a better concept hidden inside, but I wanted to push the changes before madness started (I moved countries and started a new job in the meanwhile :) )

ericniebler:
Concept-ifying the numerics algorithms is the work of @cjdb in [p1813](http://wg21.link/p1813), however I see that he didn't yet get around to `inclusive_scan` and `exclusive_scan`. I would like to see that work and this harmonized.

PR #717, opened 2017-08-26
Workaround gcc5 constexpr 14 bug in at_fn

CaseyCarter:
Re-enable constexpr 14 support on GCC 5 in `.travis.yml`

(Also strip the UTF-8 BoM from .travis.yml - no idea how it got one.)


ericniebler:
Mondo Travis failures.

PR #718, opened 2017-08-28
Rework ranges::reference_wrapper in light of LWG#2993

ericniebler:
Also, change the spelling of a wrapper of an rvalue from `reference_wrapper<T, true>` to `reference_wrapper<T &&>`.

Future: Further consider combining this with `box<>`.

### DIFF
--- a/include/range/v3/utility/invoke.hpp
+++ b/include/range/v3/utility/invoke.hpp
@@ -41,8 +41,8 @@ namespace ranges
                 is_reference_wrapper<uncvref_t<T>>>
         {};
 
-        template<typename T, bool RValue>
-        struct is_reference_wrapper<reference_wrapper<T, RValue>>
+        template<typename T>
+        struct is_reference_wrapper<reference_wrapper<T>>
           : std::true_type
         {};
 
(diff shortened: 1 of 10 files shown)


include/range/v3/utility/invoke.hpp
```diff
@@ -129,29 +129,73 @@ namespace ranges
...
-              : t_(std::addressof(t))
+            template<typename U,
+                CONCEPT_REQUIRES_(Constructible<base_, U>() &&
+                    !Same<uncvref_t<U>, reference_wrapper>())>
+            constexpr reference_wrapper(U &&u)
+                noexcept(std::is_nothrow_constructible<base_, U>::value)
+              : detail::reference_wrapper_<T>{static_cast<U &&>(u)}
             {}
-            reference_wrapper(meta::if_c<RValue, T &, T &&>) = delete;
             constexpr reference get() const noexcept
```
CaseyCarter: This wrapper is unnecessary, you can simply `using base_::get;`

ericniebler: I know, but that Doxygen's poorly.

PR #725, opened 2017-10-12
[constexpr] alg.equal

gnzlbg:
Make the `ranges::equal` algorithm constexpr.

PR #727, opened 2017-10-13
[constexpr] alg.copy

gnzlbg:


gnzlbg:
Depends on #725 .

test/algorithm/copy.cpp
```diff
@@ -67,5 +77,9 @@ int main()
         CHECK(std::strcmp(sz, buf) == 0);
     }
 
+#if RANGES_CXX_CONSTEXPR >= RANGES_CXX_CONSTEXPR_14
```
CaseyCarter: I would replace line 21 with this line...

test/algorithm/copy.cpp
```diff
@@ -67,5 +77,9 @@ int main()
...
+    static_assert(test_constexpr_copy(), "");
+#endif
```
CaseyCarter: ...and then move these two lines to line 29. (i.e., keep the `constexpr` test together as a single unit.)

gnzlbg:
@CaseyCarter I feel a bit bad about duplicating tests for constexpr purposes.

test/algorithm/copy.cpp
gnzlbg: done

gnzlbg:
@ericniebler FWIW it worked (all green), so IIUC what travis does is apply the commits in the pull-request to the receiving branch (in this case `ericniebler/range-v3:master`).

This is both good and bad. Good because I can send a PR, branch it, keep on working on something that needs those changes, and when that is done, cherry-pick into a clean branch, and pull request that. Test will initially fail, but once the dependent branches are merged they will pass.

Bad because a PR becomes "green" when the tests pass. But if you merge a PR, the tests on all other green PRs won't be re-run, so merging subsequent pull-requests can still break the build. This leads to "absurd" situations in which one PR is green but has merge conflicts (how can the tests pass if the PR can't even be merged?).

PR #730, opened 2017-10-19
stride_view refactor:

CaseyCarter:
stride views must track the distance in the base range between the last dereferenceable iterator and the end iterator ("offset"). The offset is used when decrementing bidirectional end iterators. Since there's one unique offset value per underlying range, we can store a single offset value in the stride_view rather than in each of its iterators.

Using a mutable std::atomic to store the offset and provide const-iterable stride_views was a design mistake: all users pay the cost of the atomic despite that few users will actually access the same stride_view simultaneously from multiple threads. Let's instead store the offset in a plain vanilla member variable, and make the view not const iterable when it must update the offset during iteration.


ericniebler:
LOVE this, thank you Casey!

PR #733, opened 2017-10-21
(adjacent_)?remove_if cleanup:

CaseyCarter:
* Store the predicate in a `box` for EBO
* Only adapt the sentinel if it's an iterator
* Don't cache an incorrect begin iterator if an exception is thrown in satisfy_forward
* constexpr and noexcept
* Use `base()` instead of `mutable_base()` to access the underlying range

* `adjacent_remove_if`:
  * model `BidirectionalRange` when the base range does

### DIFF
--- a/include/range/v3/view/adjacent_remove_if.hpp
+++ b/include/range/v3/view/adjacent_remove_if.hpp
@@ -16,12 +16,14 @@
 
 #include <utility>
 #include <meta/meta.hpp>
-#include <range/v3/detail/satisfy_boost_range.hpp>
-#include <range/v3/range_fwd.hpp>
 #include <range/v3/begin_end.hpp>
+#include <range/v3/range_fwd.hpp>
 #include <range/v3/view_adaptor.hpp>
-#include <range/v3/utility/iterator.hpp>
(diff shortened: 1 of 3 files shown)


include/range/v3/view/adjacent_remove_if.hpp
```diff
@@ -38,65 +40,108 @@ namespace ranges
...
                 {}
-                iterator_t<Rng> begin(adjacent_remove_if_view &) const
+                RANGES_CXX14_CONSTEXPR static iterator_t<Rng> begin(adjacent_remove_if_view &rng)
+                    noexcept(std::is_nothrow_copy_constructible<iterator_t<Rng>>::value)
+                {
+                    return *rng.begin_;
+                }
+                RANGES_CXX14_CONSTEXPR void next(iterator_t<Rng> &it) const
+                    noexcept(noexcept((void)(it != ranges::end(std::declval<Rng &>())),
+                        std::declval<adjacent_remove_if_view &>().satisfy_forward(++it)))
```
CaseyCarter: It's egregiously silly to have exception specifiers that include calls to `satisfy_(forward|reverse)` which is always `noexcept(true)`.

PR #734, opened 2017-10-21
reverse_view cleanup:

CaseyCarter:
* Don't use `mutable_base` (it was hiding a const bug in the `BoundedRange` implementation of `get_end` when `!Range<Rng const>()`)

* constexpr / conditional noexcept all the things

### DIFF
--- a/include/range/v3/view/reverse.hpp
+++ b/include/range/v3/view/reverse.hpp
@@ -74,18 +74,18 @@ namespace ranges
             {
                 return ranges::end(this->base());
             }
-            // BoundedRange == true
             RANGES_CXX14_CONSTEXPR iterator_t<Rng> get_end_(std::true_type)
                 noexcept(noexcept(ranges::end(std::declval<Rng &>())))
             {
+                CONCEPT_ASSERT(BoundedRange<Rng>());
                 return ranges::end(this->base());
             }
-            // BoundedRange == false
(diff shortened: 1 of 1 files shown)


include/range/v3/view/reverse.hpp
```diff
@@ -42,152 +42,201 @@ namespace ranges
...
                     RANGES_ASSERT(it != ranges::end(rng_->base()));
                     ++it;
                 }
                 CONCEPT_REQUIRES(RandomAccessRange<Rng>())
+                RANGES_CXX14_CONSTEXPR
                 void advance(iterator_t<Rng> &it, range_difference_type_t<Rng> n) const
+                    noexcept(noexcept(
+                        void(it - ranges::begin(std::declval<base_range_t &>())),
+                        std::declval<reverse_view_t &>().get_end(),
+                        ranges::advance(it, -n)))
```
ericniebler: Your obsession with making _everything_ conditionally `noexcept` is making me queasy.

CaseyCarter: Looking at this now, I'm not pleased with the choice to have the conditional noexcept reflect debugging code. I think e.g. this function should be `noexcept(noexcept(ranges::advance(it, -n)))`, and an exception thrown while checking contracts should irrecoverable.

CaseyCarter: ...alternately, you could tell me to rip some of these out. Your house, your rules.

PR #736, opened 2017-10-22
view_adaptor update

### LINKED ISSUE #726: Why does the stored range in `view_adaptor` need to be mutable?
[`view_adaptor`]() has the following data member:

```c++
            // Mutable here. Const-correctness is enforced below by disabling
            // begin_cursor/end_cursor if "BaseRng const" does not model
            // the View concept.
            mutable base_range_t rng_;
```

I've been trying to understand why this isn't implemented as:

```c++
            base_range_t rng_;
            base_range_t & mutable_base() const
            {
                return const_cast<base_range_t&>(rng_);
            }
```

without much luck (the comment doesn't really help). Making the member `mutable` allows modifying `rng_` even if `view_adaptor` is stored in a const object without introducing undefined behavior. However, modifying the `rng_` object from a `const` method without synchronization is unsafe. Since `const` views are close to useless, I also wonder why this matters.

---

To put this in context, `mutable` data members cannot be accessed by constexpr methods, so this prevents many views from being usable in `constexpr` contexts.

CaseyCarter:
Fixes #726.

`view_adaptor`:
* decay the types of `(begin|end)_adaptor_t` and `adapted_(iterator|sentinel)_t`
* Deduce return type/noexcept for `adaptor_base::(begin|end)`
* remove `base_range_t`
* remove `mutable` specifier from base range, and remove `mutable_base()`
* constexpr / noexcept

`view_facade`:
* decay the types of `begin_cursor_t` and `end_cursor_t`

`view_interface`:
* Cleanup constraints on `empty`
* DRY the range printing code

`range_access`:
* conditional noexcept for `(begin|end)_(adaptor|cursor)`
* Remove `base_range`

### DIFF
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -33,19 +33,19 @@ namespace ranges
         {
             template<typename Derived>
             using begin_adaptor_t =
-                decltype(range_access::begin_adaptor(std::declval<Derived &>(), 42));
+                detail::decay_t<decltype(range_access::begin_adaptor(std::declval<Derived &>(), 42))>;
 
             template<typename Derived>
             using end_adaptor_t =
-                decltype(range_access::end_adaptor(std::declval<Derived &>(), 42));
+                detail::decay_t<decltype(range_access::end_adaptor(std::declval<Derived &>(), 42))>;
 
             template<typename Derived>
             using adapted_iterator_t =
-                decltype(std::declval<begin_adaptor_t<Derived>>().begin(std::declval<Derived &>()));
+                detail::decay_t<decltype(std::declval<begin_adaptor_t<Derived>>().begin(std::declval<Derived &>()))>;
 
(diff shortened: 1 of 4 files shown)


include/range/v3/view_adaptor.hpp
```diff
@@ -379,7 +378,20 @@ namespace ranges
                 adaptor_cursor_t<D>,
                 adaptor_sentinel<detail::adapted_sentinel_t<D>, detail::end_adaptor_t<D>>>;
 
-        template<typename Derived, typename BaseRng, cardinality Cardinality /*= range_cardinality<BaseRng>::value*/>
+        /// \cond
+        namespace detail
+        {
+            template<typename Derived, typename Base>
+            using AdaptorFooable = meta::and_<
```
CaseyCarter: Fix this name.

ericniebler:
Can you check the failing tests?

PR #737, opened 2017-10-23
counted_iterator: fix post-increment bug

CaseyCarter:
When the base iterator's post-increment operator has `void` return type, we can't bind a reference thereto.


ericniebler:
Lots of test failures. Can you have a look?

CaseyCarter:
It turns out that `counted_iterator<foo> x{}; x++;` is not a great test.

PR #740, opened 2017-11-07
visiting an out-of-range variant index is UB

### LINKED ISSUE #739: view::set_difference() on deque and vector triggers GCC's null-dereference warning
Compiling this:

~~~.cpp
#include <range/v3/all.hpp>

#include <deque>
#include <vector>

int main() {
	const std::deque <size_t> a = { 1, 2,   };
	const std::vector<size_t> b = {    2, 3 };
	const std::vector<size_t> c = ranges::view::set_difference( a, b );
}
~~~

...with `g++ -std=c++11 -O2 -Werror -Wnull-dereference -isystem range-v3/include a.cpp` gives the following warning :

~~~
In function ‘int main()’:
cc1plus: error: null pointer dereference [-Werror=null-dereference]
cc1plus: all warnings being treated as errors
~~~ 

...under GCC 6.2.0 and 6.3.0.

(Of course, I understand that you might well not choose to care about this warning and that you may judge this to be a problem with GCC's null-dereference warning.)

CaseyCarter:
Fixes #739.

### DIFF
--- a/include/range/v3/detail/variant.hpp
+++ b/include/range/v3/detail/variant.hpp
@@ -306,6 +306,7 @@ namespace ranges
             template<typename Fun, typename Proj = indexed_element_fn>
             constexpr int variant_visit_(std::size_t, variant_nil, Fun, Proj = {})
             {
+                RANGES_EXPECT(false);
                 return 0;
             }
             template<typename Data, typename Fun, typename Proj = indexed_element_fn>


tonyelewis:
Thanks very much. There's a C++11 constexpr problem - please see [this comment](https://github.com/ericniebler/range-v3/issues/739#issuecomment-342391190).

PR #742, opened 2017-11-10
add first examples from range_by_example to docs

JeffGarland:
a first batch

### DIFF
--- a/example/count.cpp
+++ b/example/count.cpp
@@ -27,12 +27,15 @@ using std::cout;
 int main() {
 
   std::vector<int> v { 6, 2, 3, 4, 5, 6 };
-  int c = ranges::count( v, 6 );
+  //note the count return is a numeric type 
+  //like int or long -- auto below make sure 
+  //it matches the implementation
+  auto c = ranges::count( v, 6 );
   cout << "vector:   " << c << "\n";
 
   std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
   c = ranges::count( a, 6 );
-  cout  << "array:    "<< c << "\n";
+  cout << "array:    " << c << "\n";
 }
 
 ///[count]
--- a/example/count.cpp
+++ b/example/count.cpp
@@ -1,0 +1,38 @@
+
+// Range v3 library
+//
+//  Copyright Jeff Garland 2017
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+// 
+
+
+///[count]
+// This example demonstrates counting the number of 
+// elements that match a given value. 
+// output...
+// vector:   2
+// array:    2
+
+#include <range/v3/algorithm/count.hpp>   //specific includes
+#include <vector>
+#include <iostream>
+using std::cout;
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count( v, 6 );
+  cout << "vector:   " << c << "\n";
+
+  std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
+  c = ranges::count( a, 6 );
+  cout  << "array:    "<< c << "\n";
+}
+
+///[count]
--- a/example/count_if.cpp
+++ b/example/count_if.cpp
@@ -11,7 +11,6 @@
 // Project home: https://github.com/ericniebler/range-v3
 //
 
-
 ///[count_if]
 // This example counts element of a range that match a supplied predicate.
 // output
@@ -29,7 +28,7 @@ auto is_six = [](int i) -> bool { return i == 6; };
 int main() {
 
   std::vector<int> v { 6, 2, 3, 4, 5, 6 };
-  int c = ranges::count_if( v, is_six );
+  auto c = ranges::count_if( v, is_six );
   cout << "vector:   " << c << "\n"; //2
 
   std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
--- a/example/count_if.cpp
+++ b/example/count_if.cpp
@@ -1,0 +1,39 @@
+
+// Range v3 library
+//
+//  Copyright Jeff Garland 2017
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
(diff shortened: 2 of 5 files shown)


ericniebler:
Thanks! Can you look into the build failures on Travis? https://travis-ci.org/ericniebler/range-v3/jobs/300399003

Let me know if you need a hand.

example/count.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count( v, 6 );
+  cout << "vector:   " << c << "\n";
+
+  std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
+  c = ranges::count( a, 6 );
+  cout  << "array:    "<< c << "\n";
```
FloopCZ: Two spaces after `cout`, no space after `"`.

example/hello.cpp
```diff
@@ -0,0 +1,31 @@
...
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+// 
+
+///[hello]
+#include <range/v3/all.hpp>  //get everything
+namespace rng = ranges::v3;  //easy access
```
FloopCZ: This alias is never used.

example/count_if.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+auto is_six = [](int i) -> bool { return i == 6; };
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count_if( v, is_six );  
+  cout << "vector:   " << c << "\n"; //2
+
+  std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
```
FloopCZ: ditto array initialization

example/count.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <iostream>
+using std::cout;
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count( v, 6 );
+  cout << "vector:   " << c << "\n";
+
+  std::array<int, 6> a { 6, 2, 3, 4, 5, 6 };
```
FloopCZ: Unfortunately, you will probably have to use
```
std::array<int, 6> a {{ 6, 2, 3, 4, 5, 6 }};
```
to satisfy older compilers.

example/count.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <range/v3/algorithm/count.hpp>   //specific includes
+#include <vector>
+#include <iostream>
+using std::cout;
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count( v, 6 );
```
FloopCZ: `ranges::count` returns `difference_type_t`, so `long` may be more appropriate in here instead of `int` to avoid conversion warnings

example/count_if.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <array>
+#include <iostream>
+using std::cout;
+
+auto is_six = [](int i) -> bool { return i == 6; };
+
+int main() {
+
+  std::vector<int> v { 6, 2, 3, 4, 5, 6 };
+  int c = ranges::count_if( v, is_six );
```
FloopCZ: ditto conversion warnings

CaseyCarter:
The first error I saw is an overzealous warning (promoted by `-Werror`) from clang about missing braces in `std::array foo{1, 2, 3};`. I'm very much in favor of the draconian warning settings we use for the test suite - ideally users should be able to include range-v3 in a compile with diagnostic settings of their choice without the headers emitting diagnostics - but it doesn't make sense to expose examples to the same environment. Examples should be written clearly and naturally and should not emit diagnostics when exposed to "normal" warning settings. We certainly don't want to make them less readable by riddling them with warning suppressions.

I'll take a look at wrangling the CMake config.

example/count_if.cpp
CaseyCarter: There are a couple of trailing spaces on this line.

example/count.cpp
CaseyCarter: `auto` would be a superior choice: the difference type of `vector` iterators is implementation-defined. It's going to be `ptrdiff_t` on any sane implementation, which is equivalent to `int` or `long` or `long long` on most implementations. Choosing too small a type results in a warning here, choosing too large a type results in conversion warnings elsewhere when `c` is converted to a shorter type.

That said, `int` is dead simple here. If we want to turn off conversion warnings for examples I would not object.

example/count.cpp
CaseyCarter: > to satisfy older compilers

Like Clang 5. I'm going to disable `-Wmissing-braces` for examples.

CaseyCarter:
> I'll take a look at wrangling the CMake config.

I have a fix (https://github.com/JeffGarland/range-v3/compare/master...CaseyCarter:fewer_example_warnings) for this in CI (https://travis-ci.org/CaseyCarter/range-v3/builds/300653715, https://ci.appveyor.com/project/CaseyCarter/range-v3/build/1.0.327). I'll PR it into @JeffGarland's branch once CI completes and I'm sure I didn't miss anything.

I did decide to disable 64-to-32 warnings for examples to allow truncation, speak up if anyone disagrees with that choice.

ericniebler:
@CaseyCarter you're still getting a `-Werror,-Wmissing-braces` for one test. Strange.

CaseyCarter:
The clang modules build is the bane of my existence.  -Wno-foo is right there on the command line, but it's still emitting the suppressed warnings. I'll look into it when I'm at the computer again.

example/count.cpp
JeffGarland: So let me say first that gcc-head gave me no warnings with the cmake.

That said I was also unaware of what it might be here.  So auto works fine for me, but I'm pondering adding a comment that explains the return type in the example.

example/count.cpp
JeffGarland: Fixed in my working copy, in the next push.

example/hello.cpp
JeffGarland: gone in my working copy.  This alias was used extensively in the range_by_example code, but wanted these examples to match all the other docs.

example/count_if.cpp
JeffGarland: fixed in working copy

example/count_if.cpp
JeffGarland: So for now I'm not changing the array init based on casey's comments.

JeffGarland:
@CaseyCarter I've pulled your cmake changes - updating momentarily

PR #752, opened 2017-11-21
LLVM 33314 is purportedly fixed on trunk

CaseyCarter:
...disable the workarounds for clang >= 6


CaseyCarter:
Travis failure (https://travis-ci.org/ericniebler/range-v3/jobs/305515891) is transient.

PR #760, opened 2017-12-08
unique_view: don't evaluate iterator_t<non-range>

CaseyCarter:
... so that mutable-only base views actually work.


CaseyCarter:
This begs for a regression test.

PR #761, opened 2017-12-12
view_adaptor: guard const overloads of (begin|end)_cursor

### LINKED ISSUE #758: View adaptor update destroyed compilation times
Hi everyone,

In the place where I work, we have a larger project utilizing `range-v3` heavily. The project takes about 4 minutes and 20GB of RAM to compile with clang++-5.0 (`--std=c++17`). Unfortunately, after 25f7db1d96968e1d19b8f4109560efea4c3b8a02 (found with `git bisect`) it takes at least 30 minutes to compile (to be honest, I never waited till the end actually :-)) and eats about 40GB of memory. Do you have any idea what could cause such a difference (@CaseyCarter)? I cannot see anything that evil in the commit.

(relates to https://github.com/ericniebler/range-v3/issues/332)

Thank you!

CaseyCarter:
...with `Range<foo const>()`. Compilers - clang 5, at least - cache the value of *successful* template substitution, not the value of *failed* template substitution.

Fixes #758.

### DIFF
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -417,7 +417,8 @@ namespace ranges
             {
                 return view_adaptor::begin_cursor_(derived());
             }
-            template<typename D = Derived, CONCEPT_REQUIRES_(Same<D, Derived>())>
+            template<typename D = Derived,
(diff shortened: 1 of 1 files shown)


ericniebler:
Thanks!

PR #763, opened 2017-12-12
Ref view

CaseyCarter:


include/range/v3/view/all.hpp
```diff
@@ -29,72 +31,89 @@ namespace ranges
...
+                return ranges::empty(*rng_);
+            }
+            template<class R = Rng>
+            constexpr auto data() const
+                RANGES_DECLTYPE_NOEXCEPT(ranges::data(std::declval<R &>()))
+            {
+                return ranges::data(*rng_);
+            }
+        };
+
```
ericniebler: These members need constraints, per you recent discovery about the ills of SFINAE return types, no? It is that only for alias templates?

test/iterator_range.cpp
```diff
@@ -66,8 +65,6 @@ int main()
     CHECK(l0.end() == li.end());
     CHECK(l0.size() == li.size());
 
-    l0 = view::all(li);
-
```
ericniebler: Why?

test/view/delimit.cpp
```diff
@@ -27,7 +27,7 @@ int main()
     ::models<concepts::View>(aux::copy(rng0));
     ::models_not<concepts::BoundedView>(aux::copy(rng0));
     ::models<concepts::RandomAccessIterator>(rng0.begin());
-    CONCEPT_ASSERT(RandomAccessView<delimit_view<std::vector<int>, int>>());
+    CONCEPT_ASSERT(RandomAccessView<delimit_view<std::vector<int> &, int>>());
```
ericniebler: This is odd. The views adaptor class templates all expect to be parameterized over views. A reference to a container is not a view. What's up?

include/range/v3/view/all.hpp
CaseyCarter: I really want compilers to cache the result of overload resolution even when that result is failure, but for all I know this return type SFINAE could perform as badly as the alias template substitution failure case from #758. I'll constrain these.

include/range/v3/view/all.hpp
```diff
@@ -103,8 +122,7 @@ namespace ranges
             RANGES_INLINE_VARIABLE(all_fn, all)
 
             template<typename Rng>
-            using all_t =
-                meta::_t<std::decay<decltype(all(std::declval<Rng>()))>>;
+            using all_t = decltype(all(std::declval<Rng>()));
```
CaseyCarter: Note that this decay is unnecessary, all of `all_fn`'s `operator()` overloads return cv-and-ref-unqualified types.

test/iterator_range.cpp
CaseyCarter: This ought to work; maybe `sized_iterator_range` doesn't have the same implicit conversions as in cmcstl2/P0789? I'll investigate.

CaseyCarter: Yip - that's it. We need to port `subrange` to range-v3.

test/view/delimit.cpp
CaseyCarter: IIRC, this is leakage from a failed experiment in which I hacked `view::all(ContiguousContainer&)` to return `span`. I clearly need to go through this patch line-by-line. (This patch is several months old, I only recently found time to polish it up and workaround GCC5 trying to instantiate `declval` in C++11 mode as it loves to do. I *thought* I had left the branch clean but that is obviously not the case.)

include/range/v3/view/all.hpp
```diff
@@ -29,72 +31,89 @@ namespace ranges
...
+                        "Cannot get a view of a temporary container");
+                    return ref_view<meta::_t<std::remove_reference<T>>>{t};
                 }
 
-                // TODO handle char const * by turning it into a delimited range?
-
-            public:
+                /// Translate std::reference_wrapper<Rng> into ref_view<Rng>
                 template<typename T,
                     CONCEPT_REQUIRES_(Range<T>())>
```
CaseyCarter: This could be `reference_wrapper<View>`. Ditto 112.

test/view/all.cpp
```diff
@@ -22,21 +22,27 @@ int main()
...
     x = view::all(x);
     y = view::all(y);
     z = view::all(z);
 
-    CHECK(x.size() == 7u);
-    CHECK(y.size() == 7u);
-    CHECK(z.size() == 7u);
+    CHECK(ranges::size(x) == 7u);
+    CHECK(ranges::size(y) == 7u);
+    CHECK(ranges::size(z) == 7u);
```
CaseyCarter: This change isn't *incorrect* but it's unnecessary: `ref_view<SizedRange>` always has member `size`. Let's put this back.

test/view/all.cpp
```diff
@@ -22,21 +22,27 @@ int main()
...
     std::vector<int> vi(begin(rgi), end(rgi));
-    std::list<int> li(begin(rgi), end(rgi));
+    std::list<int> const li(begin(rgi), end(rgi));
 
-    iterator_range<int *> x = view::all(rgi);
-    iterator_range<std::vector<int>::iterator> y = view::all(vi);
-    sized_iterator_range<std::list<int>::iterator> z = view::all(li);
+    auto x = view::all(rgi);
+    ::has_type<ref_view<decltype(rgi)> &>(x);
+    CONCEPT_ASSERT(range_cardinality<decltype(x)>::value == 7);
```
CaseyCarter: Another advantage of `ref_view` over `iterator_range`: it preserves static extent.

include/range/v3/view/all.hpp
CaseyCarter: So constrained. I'm not married to the name `CanEmptyRange`.

include/range/v3/view/all.hpp
```diff
@@ -15,86 +15,136 @@
...
-                    return {begin(t), end(t), size(t)};
+                    return static_cast<T &&>(t);
                 }
-
-                template<typename T>
-                static iterator_range<iterator_t<T>, sentinel_t<T>>
-                from_container(T & t, concepts::SizedRange*, concepts::SizedSentinel*)
+                template<typename T,
+                    CONCEPT_REQUIRES_(View<T>())>
+                constexpr T operator()(std::reference_wrapper<T> ref) const
```
CaseyCarter: `view::all` now turns `reference_wrappers` that reference `View`s into `View`s.

test/iterator_range.cpp
CaseyCarter: Would you like me to create a reminder issue that we need to port `subrange`, or block this merge until that happens, or...?

test/view/all.cpp
CaseyCarter: I suppose this is actually another side of `ref_view`'s single advantage over `subrange`: it preserves the range type instead of decaying to an iterator/sentinel pair.

test/iterator_range.cpp
ericniebler: You could patch this up by adding implicit conversions from `ref_view` to `[sized_]iterator_range`.

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,431 @@
...
+#include <range/v3/begin_end.hpp>
+#include <range/v3/distance.hpp>
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/view_interface.hpp>
+#include <range/v3/utility/compressed_pair.hpp>
+#include <range/v3/utility/dangling.hpp>
+
+// TODO:
+// * Port the "ReferenceableRange" changes back to cmcstl2 after getting Eric's input
```
CaseyCarter: This comment needs to be struck.

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,431 @@
...
+                if constexpr (StoreSize) {
+                    return size_();
+                } else {
+                    return last_() - first_();
+                }
+#else
+                return size_helper(meta::bool_<StoreSize>());
+#endif
+            }
+
```
CaseyCarter: `data()`? Encode contiguity in `subrange_kind`?

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,431 @@
...
+            static_cast<subrange_kind>(SizedRange<R>() ||
+                SizedSentinel<sentinel_t<R>, iterator_t<R>>())>;
+
+        template<typename R, CONCEPT_REQUIRES_(ReferenceableRange<R>())>
+        subrange(R &&, range_difference_type_t<R>) ->
+            subrange<iterator_t<R>, sentinel_t<R>, subrange_kind::sized>;
+#endif // RANGES_CXX_DEDUCTION_GUIDES
+
+        template<typename R, CONCEPT_REQUIRES_(Range<R>())>
+        using safe_subrange_t =	meta::if_<
```
CaseyCarter: There's a stray tab here.

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,431 @@
...
+
+        template<std::size_t N, typename I, typename S, subrange_kind K,
+            CONCEPT_REQUIRES_(N == 1)>
+        constexpr S get(subrange<I, S, K> const &r) {
+            return r.end();
+        }
+#endif
+
+        template<typename I, typename S, CONCEPT_REQUIRES_(Sentinel<S, I>())>
+        constexpr auto make_subrange(I i, S s)
```
CaseyCarter: Should this be named `view::subrange`?

test/view/subrange.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+    l0 = view::all(li);
+
+    subrange<std::list<int>::iterator> l1 = l0;
+    ::models_not<concepts::SizedRange>(l1);
+    CHECK(l1.begin() == li.begin());
+    CHECK(l1.end() == li.end());
+
+    {
+#if RANGES_CXX_DEDUCTION_GUIDES >= RANGES_CXX_DEDUCTION_GUIDES_17
```
CaseyCarter: Be sure to port these deduction guide tests back to cmcstl2.

test/view/subrange.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+// https://bugs.llvm.org/show_bug.cgi?id=34776
+RANGES_DIAGNOSTIC_IGNORE_UNDEFINED_FUNC_TEMPLATE
+
+using namespace ranges;
+
+template<typename Rng>
+safe_subrange_t<Rng> algorithm(Rng &&rng);
+
+template<class...> class show_type;
```
CaseyCarter: Strike this debugging tool.

include/range/v3/view/subrange.hpp
ericniebler: I'd rather leave it as `ranges::make_subrange`.

include/range/v3/view/subrange.hpp
ericniebler: > `data()`?

If `I` is a pointer, sure.

> Encode contiguity in `subrange_kind`?

Nah.

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,428 @@
...
+
+            CONCEPT_REQUIRES(K == subrange_kind::sized)
+            RANGES_CXX14_CONSTEXPR difference_type_t<I> size() const
+            {
+#if RANGES_CXX_IF_CONSTEXPR >= RANGES_CXX_IF_CONSTEXPR_17
+                if constexpr (StoreSize) {
+                    return size_();
+                } else {
+                    return last_() - first_();
+                }
```
ericniebler: Brace formatting here, and elsewhere. https://en.wikipedia.org/wiki/Indentation_style#Allman_style

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,428 @@
...
+                std::get<1>(static_cast<PairLike &&>(r)), n}
+            {}
+
+            template<typename PairLike,
+                CONCEPT_REQUIRES_(!Range<PairLike>() &&
+                    detail::PairLikeConvertibleFrom<PairLike, I const &, S const &>())>
+            constexpr operator PairLike() const
+            {
+                return PairLike(first_(), last_());
+            }
```
ericniebler: I'm supposing this is enough to make `subrange` convertible to `iterator_range`, correct?

test/algorithm/copy_backward.cpp
```diff
@@ -40,7 +40,7 @@ int main()
     CHECK(std::equal(a, a + size(a), out));
 
     std::fill_n(out, size(out), std::make_pair(0, 0));
-    auto res2 = ranges::copy_backward(ranges::view::all(a), end(out));
+    auto res2 = ranges::copy_backward(ranges::make_iterator_range(a), end(out));
```
ericniebler: Why?

include/range/v3/iterator_range.hpp
```diff
@@ -59,6 +61,17 @@ namespace ranges
...
+                CONCEPT_REQUIRES_(Constructible<I, iterator_t<Rng>>() &&
+                    Constructible<S, sentinel_t<Rng>>())>
+            constexpr iterator_range(ref_view<Rng> r)
+              : base_t{r.begin(), r.end()}
+            {}
+            template<typename X, typename Y, subrange_kind K,
+                CONCEPT_REQUIRES_(Constructible<I, X>() && Constructible<S, Y>())>
+            constexpr iterator_range(subrange<X, Y, K> const &rng)
+              : base_t{rng.begin(), rng.end()}
+            {}
```
ericniebler: Ah, there. Thanks.

include/range/v3/view/subrange.hpp
```diff
@@ -0,0 +1,428 @@
...
+
+        template<typename P, CONCEPT_REQUIRES_(detail::IteratorSentinelPair<P>())>
+        subrange(P, difference_type_t<meta::_t<std::tuple_element<0, P>>>) ->
+            subrange<meta::_t<std::tuple_element<0, P>>,
+                meta::_t<std::tuple_element<1, P>>, subrange_kind::sized>;
+
+        template<typename R, CONCEPT_REQUIRES_(ReferenceableRange<R>())>
+        subrange(R &&) -> subrange<iterator_t<R>, sentinel_t<R>,
+            static_cast<subrange_kind>(SizedRange<R>() ||
+                SizedSentinel<sentinel_t<R>, iterator_t<R>>())>;
```
CaseyCarter: This disjunction is theoretically necessary: a single-pass range could satisfy `SizedSentinel<sentinel_t<R>, iterator_t<R> && !SizedRange<R>` since `ranges::size` cannot determine the range size from `end() - begin()` for non-forward ranges. A range could legitimately overload `size` to return something other than the size of the range to meet some strange external requirement, and specialize `disable_sized_range`.

test/view/subrange.cpp
CaseyCarter: Done.

include/range/v3/view/subrange.hpp
CaseyCarter: This can wait for the `ContiguousRange` paper.

test/algorithm/copy_backward.cpp
CaseyCarter: These algorithm tests ensure that the algorithm returns a `dangling` result when passed an argument that is not a `ReferenceableRange`. `ref_view` and `subrange` both satisfy `ReferenceableRange`, so using `make_subrange` here would break the test just the same as does using `view:all`. Since `iterator_range` doesn't satisfy `ReferenceableRange`, and we're keeping it around for backwards compatibility, this was an easy fix. (Although I note in this specific instance that `a` is a container so `std::move(a)` would work as well.)

In cmcstl2 I fixed the algorithm tests by adding a test-only `unref_view` view adaptor that simply wraps the view with which it's initialized. I could/should do the same here, which would (a) be consistent across both libraries, which is valuable in itself, and (b) allow us to deprecate `iterator_range` with `RANGES_DEPRECATE` if you'd like to push people to `subrange` more quickly.

include/range/v3/view/all.hpp
```diff
@@ -15,86 +15,139 @@
...
+            constexpr auto data() const
+                RANGES_DECLTYPE_NOEXCEPT(ranges::data(std::declval<R &>()))
+            {
+                return ranges::data(*rng_);
+            }
+        };
+
+        template<typename R>
+        struct is_referenceable_range<ref_view<R>> : std::true_type {};
+
```
ericniebler: Not a fan of this, as I describe [here](https://github.com/CaseyCarter/cmcstl2/pull/126#pullrequestreview-85859245).

test/algorithm/copy_backward.cpp
ericniebler: Yeah... about that.... https://github.com/CaseyCarter/cmcstl2/pull/126#pullrequestreview-85859245

include/range/v3/view/subrange.hpp
ericniebler: A comment in the code to that effect would be a nice addition.

ericniebler:
Things range-v3 needs before this PR can be merged:

1. New `begin`/`end` customization point objects.
2. `subrange` from P0789.
3. A pony.

(3) is optional. And once we have (1) we don't need `ReferenceableRange`.

ericniebler:
@CaseyCarter do you have any plans for this PR, or should I close it?

CaseyCarter:
> I won't get back to this anytime soon - let's close it for now.

PR #770, opened 2018-01-16
[doc] Change return type of is_word_boundary to bool

Burgos:
is_word_boundary example using C++ iterators was incorrectly returning
void, while it should return bool.


ericniebler:
Thanks. N4128 is a historical document, presented to the C++ Committee years ago, but it couldn't hurt to accept bug fixes since modern-day folks might read it to learn the design.

PR #794, opened 2018-02-18
Implement non-const view_interface::empty and ::size

### LINKED ISSUE #793: No empty() function in remove_if_view
While we can check the view empty or not with `a.begin() == a.end()`, could we have an empty() implement there as well?
 
```
#include <iostream>
#include <range/v3/all.hpp>
#include <vector>

int main() {
  std::vector<int> vec{1, 2, 3};
  auto a = vec | ranges::view::remove_if([](auto& val) { return val > 0; });
  std::cout << (a.begin() == a.end()) << "\n";

  // compile error here
  std::cout << a.empty() << "\n";
  // std::cout << a.size() << "\n";
}
```

CaseyCarter:
Assume that ranges with infinite cardinality are never empty.

Fixes #793.

Drive-by: sprinkle some constexpr on view_interface

### DIFF
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -39,7 +39,7 @@ namespace ranges
                 To to;
                 template<typename F, typename T,
                     CONCEPT_REQUIRES_(ConvertibleTo<F, From>() && ConvertibleTo<T, To>())>
-                slice_bounds(F from, T to)
+                constexpr slice_bounds(F from, T to)
                   : from(from), to(to)
                 {}
             };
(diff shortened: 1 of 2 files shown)


include/range/v3/view_interface.hpp
```diff
@@ -85,13 +85,25 @@ namespace ranges
             view_interface &operator=(view_interface &&) = default;
             view_interface &operator=(view_interface const &) = default;
             // A few ways of testing whether a range can be empty:
-            CONCEPT_REQUIRES(Cardinality >= 0)
+            CONCEPT_REQUIRES(Cardinality >= 0 || Cardinality == infinite)
```
CaseyCarter: This is a behavioral change as well: assume ranges with `infinite` cardinality are never empty.

ericniebler:
Not to self: This needs to be done for P0789 also.

PR #795, opened 2018-02-18
upgrade to conan 1.0

memsharded:
conan 1.0 got some improvements (thanks for your feedback back then). With them it is no longer necessary to use the template system, as versions don't need to be hardcoded in the recipes. Also, instead of the ``export + test_package`` the command ``conan create`` is more intuitive and direct.


CMakeLists.txt
```diff
@@ -104,30 +104,18 @@ if(NOT RANGE_V3_NEW_VERSION_HPP STREQUAL RANGE_V3_OLD_VERSION_HPP)
...
-  configure_file(test_package/conanfile.py.in ${CMAKE_CURRENT_SOURCE_DIR}/test_package/conanfile.py @ONLY)
-  execute_process(
-    COMMAND ${GIT_EXECUTABLE} -C "${CMAKE_CURRENT_SOURCE_DIR}" add -u
-  )
-  execute_process(
-    COMMAND ${GIT_EXECUTABLE} -C "${CMAKE_CURRENT_SOURCE_DIR}" commit --amend --no-edit
-  )
-  execute_process(
-    COMMAND ${GIT_EXECUTABLE} -C "${CMAKE_CURRENT_SOURCE_DIR}" tag -f -a "${RANGE_V3_VERSION}" -m "${RANGE_V3_VERSION}"
-  )
```
ericniebler: These three `execute_process` commands are important and have nothing to do with conan. Please reinstate them.

memsharded: Sorry, my mistake, I quickly assumed they were related to the commit of the generated conanfiles, without carefully reading them. Fixed.

PR #800, opened 2018-02-22
span::subspan<N, -1>() is equivalent to subspan<N>()

### LINKED ISSUE #799: span::subspan<0, -1>() is broken
Per [P0122](https://wg21.link/p0122) `subspan<0, dynamic_extent>()` should be equivalent to `subspan<0>()` instead of being ill-formed.

CaseyCarter:
Fixes #799.

### DIFF
--- a/include/range/v3/span.hpp
+++ b/include/range/v3/span.hpp
@@ -97,6 +97,13 @@ namespace ranges
             private:
                 span_index_t size_ = 0;
             };
+
+            constexpr span_index_t subspan_extent(
+                span_index_t Extent, span_index_t Offset, span_index_t Count) noexcept
+            {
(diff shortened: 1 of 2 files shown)


include/range/v3/span.hpp
```diff
@@ -97,6 +97,10 @@ namespace ranges
             private:
                 span_index_t size_ = 0;
             };
+
+            template<span_index_t Extent, span_index_t Offset, span_index_t Count>
+            using subspan_extent = std::integral_constant<span_index_t,
+                Count == dynamic_extent && Extent != dynamic_extent ? Extent - Offset : Count>;
```
ericniebler: Why isn't this just a `constexpr` function?

CaseyCarter: For absolutely no good reason.

PR #803, opened 2018-02-27
[split_view] split functions must also accept (iterator, iterator)

CaseyCarter:
...to accommodate usage in the `cursor` constructor.

Bug reported at https://stackoverflow.com/questions/49015671

### DIFF
--- a/include/range/v3/view/split.hpp
+++ b/include/range/v3/view/split.hpp
@@ -151,23 +151,28 @@ namespace ranges
                 struct predicate_pred
                 {
                     semiregular_t<Pred> pred_;
+
+                    template<class S, CONCEPT_REQUIRES_(Sentinel<S, iterator_t<Rng>>())>
                     std::pair<bool, iterator_t<Rng>>
-                    operator()(iterator_t<Rng> cur, sentinel_t<Rng> end) const
+                    operator()(iterator_t<Rng> cur, S end) const
                     {
                         auto where = ranges::find_if_not(cur, end, std::ref(pred_));
-                        return std::make_pair(cur != where, where);
+                        return std::pair<bool, iterator_t<Rng>>{cur != where, where};
                     }
                 };
                 template<typename Rng>
                 struct element_pred
                 {
                     range_value_type_t<Rng> val_;
+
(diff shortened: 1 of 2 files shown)


include/range/v3/view/split.hpp
```diff
@@ -202,6 +209,7 @@ namespace ranges
                 using FunctionConcept = meta::and_<
                     ForwardRange<Rng>,
                     Invocable<Fun&, iterator_t<Rng>, sentinel_t<Rng>>,
+                    Invocable<Fun&, iterator_t<Rng>, iterator_t<Rng>>,
```
CaseyCarter: This additional constraint is really "breaking" since the implementation was already relying on it implicitly (See line 114).

include/range/v3/view/split.hpp
```diff
@@ -151,49 +151,56 @@ namespace ranges
...
                     std::pair<bool, iterator_t<Rng>>
-                    operator()(iterator_t<Rng> cur, sentinel_t<Rng> end) const
+                    operator()(iterator_t<Rng> cur, S end) const
                     {
-                        using P = std::pair<bool, iterator_t<Rng>>;
                         RANGES_EXPECT(cur != end);
-                        return *cur == val_ ? P{true, ranges::next(cur)} : P{false, cur};
+                        bool const match = *cur == val_;
+                        if (match) ++cur;
+                        return std::pair<bool, iterator_t<Rng>>{match, cur};
```
ericniebler:
```c++
return {0 != advance(cur, *cur == val, unreachable{}), cur};
```

:-)

CaseyCarter: :gag:

CaseyCarter: This is like playing the "C++11 constexpr" game.

ericniebler: You're right. I need to move the iterator:

```c++
return {0 != advance(cur, *cur == val, unreachable{}), std::move(cur)};
```

Perfect! :-D

ericniebler: Drat! Foiled! Curses! And if it weren't for you meddling kids....

CaseyCarter: image

PR #805, opened 2018-03-06
fix bug with decrementing the end of a non-const-iterable, bounded, bidirectional stride view

ericniebler:
@CaseyCarter I feel that `stride` needs a thorough rethink, but this puts a bandaid on the problem for now.

A better design, IMO, would be for a `stride` view of a Bidirectional, Common ("Bounded"), non-Sized range would not attempt to satisfy CommonRange since the end iterator could not be correctly decremented in O(1); the offset is not yet known.

### DIFF
--- a/include/range/v3/view/stride.hpp
+++ b/include/range/v3/view/stride.hpp
@@ -77,9 +77,9 @@ namespace ranges
                 void set_offset(range_difference_type_t<Rng> const) const noexcept
                 {}
                 RANGES_CXX14_CONSTEXPR
-                range_difference_type_t<Rng> get_offset() const noexcept
+                range_difference_type_t<Rng> get_offset(bool check = true) const noexcept
                 {
-                    RANGES_EXPECT(0 <= offset_);
+                    RANGES_EXPECT(!check || 0 <= offset_);
                     return offset_;
                 }
 
@@ -117,7 +117,7 @@ namespace ranges
                 void set_offset(range_difference_type_t<Rng> const) const noexcept
                 {}
                 RANGES_CXX14_CONSTEXPR
-                range_difference_type_t<Rng> get_offset() const noexcept
+                range_difference_type_t<Rng> get_offset(bool = true) const noexcept
                 {
                     return 0;
                 }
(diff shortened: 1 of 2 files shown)


CaseyCarter:
Dropping my veil of objectivity, hiding linear traversals inside "amortized O(1)" is always super icky. The carefully-crafted math doesn't really reflect the performance impact that a linear traversal has on the memory hierarchy. It's been suggested to me by at least one person that e.g. `view::reverse` should only adapt `Common` ranges to avoid the hidden linear traversal.

I'm tempted to agree with your suggestion simply because `Bidirectional && Common && !Sized` seems relatively rare and dropping `Common` from the view adaptor in that case removes the icky hidden linear traversal.

ericniebler:
There is a difference in degree of ickiness between `filter::begin` on one hand and `stride` and `reverse` on the other. Iterate a `filter` view start to finish, and every element of the underlying range exactly once. But take the `stride` view above and iterate it in reverse order, and every element gets visited _twice_. Likewise for iterating a `reverse` view of a non-`Common` range. I think we should rethink that.

gnzlbg:
> Likewise for iterating a reverse view of a non-Common range. I think we should rethink that.

Is there a way to iterate a non-Common range in reverse without touching every element twice?

ericniebler:
> > Likewise for iterating a reverse view of a non-Common range. I think we should rethink that.
> 
> Is there a way to iterate a non-Common range in reverse without touching every element twice?

Egad, you're right, and I misspoke. A reverse of a non-Common range visits every element _three times_. :-(

CaseyCarter:
> Egad, you're right, and I misspoke. A reverse of a non-Common range visits every element three times. :-(

It depends on how we interpret "visit an element". I would say that it's necessary to examine every iterator value in the range three times, but that is all done in two independent linear sweeps across the range. One sweep to find the end examining each iterator once, and one to perform the reverse iteration which examines each iterator twice: once to see if it's the begin iterator value, and once to access the element it denotes. This certainly doesn't have the cache effects that three independent linear sweeps through the range would.

PR #808, opened 2018-03-11
view::chunk should tolerate mutable-only base ranges

CaseyCarter:
Fixes https://stackoverflow.com/questions/49210190.


ericniebler:
Thanks! I had just started making these changes myself, but you beat me to the punch.

PR #809, opened 2018-03-17
Replace [indirect_]result_of[_t] with [indirect_]invoke_result[_t]

ericniebler:


include/range/v3/action/action.hpp
```diff
@@ -132,7 +132,7 @@ namespace ranges
                 CONCEPT_REQUIRES_(is_pipeable<Action>() && Range<Rng &>() &&
                     Invocable<bitwise_or, ref_t<Rng &>, Action &>() &&
                     Same<ref_t<Rng &>,
-                        result_of_t<bitwise_or(ref_t<Rng &> &&, Action &)>>())>
+                        invoke_result_t<bitwise_or, ref_t<Rng &> &&, Action &>>())>
```
CaseyCarter: `ref_t<Rng &> &&` could be simplified to `ref_t<Rng &>` here (because `declval`).

include/range/v3/utility/functional.hpp
```diff
@@ -206,11 +206,11 @@ namespace ranges
             struct Invocable
             {
                 template<typename Fun, typename... Args>
-                using result_t = result_of_t<Fun &&(Args &&...)>;
+                using result_t = invoke_result_t<Fun &&, Args &&...>;
```
CaseyCarter: Ditto, we could equivalently strip the `&&`s here.

include/range/v3/utility/functional.hpp
```diff
@@ -206,11 +206,11 @@ namespace ranges
...
 
                 template<typename Fun, typename... Args>
                 auto requires_() ->
-                    meta::void_<result_of_t<Fun &&(Args &&...)>>;
+                    meta::void_<invoke_result_t<Fun &&, Args &&...>>;
```
CaseyCarter: ...and here (I'll stop commenting these occurrences now).

PR #811, opened 2018-03-24
Implement view::exclusive_scan

mitsutaka-takeda:
Implement the sequential version of the algorithm std::exclusive_scan as a view.

### DIFF
--- a/include/range/v3/view/exclusive_scan.hpp
+++ b/include/range/v3/view/exclusive_scan.hpp
@@ -1,0 +1,164 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Mitsutaka Takeda 2018-present
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGE_V3_VIEW_EXCLUSIVE_SCAN_HPP
+#define RANGE_V3_VIEW_EXCLUSIVE_SCAN_HPP
+
+
+#include <range/v3/utility/concepts.hpp>
+#include <range/v3/view_adaptor.hpp>
+#include <range/v3/view/view.hpp>
+
+namespace ranges {
+    inline namespace v3 {
+        /// \addtogroup group-views
+        /// @{
+        template<typename Rng, typename T, typename Fun>
+        struct exclusive_scan_view
+                : view_adaptor<exclusive_scan_view<Rng, T, Fun>, Rng> {
+        private:
+            friend range_access;
+            T init_;
+            semiregular_t<Fun> fun_;
+            using single_pass = SinglePass<iterator_t<Rng>>;
+            using use_sentinel_t = meta::or_<meta::not_<BoundedRange<Rng>>, single_pass>;
+
+            template<bool IsConst>
+            struct adaptor : adaptor_base {
+            private:
+                using exclusive_scan_view_t = meta::const_if_c<IsConst, exclusive_scan_view>;
+                // sum_'s type doesn't have to be range_value_type_t<Rng>.
+                semiregular_t<range_value_type_t<Rng>> sum_;
+                exclusive_scan_view_t *rng_;
+            public:
+                using single_pass = exclusive_scan_view::single_pass;
+                adaptor() = default;
+                adaptor(exclusive_scan_view_t &rng)
+                        : rng_(&rng)
+                {}
+                iterator_t<Rng> begin(exclusive_scan_view_t &)
+                {
+                    sum_ = rng_->init_;
+                    return ranges::begin(rng_->base());
+                }
+                range_value_type_t<Rng> read(iterator_t<Rng>) const
+                {
+                    return sum_;
+                }
+                void next(iterator_t<Rng> &it)
+                {
+                    if (it != ranges::end(rng_->base()))
+                    {
+                        auto &current = static_cast<range_value_type_t<Rng> &>(sum_);
+                        sum_ = invoke(rng_->fun_, current, *it);
+                        ++it;
+                    }
+                }
+                void prev() = delete;
+            };
+
+            adaptor<false> begin_adaptor()
+            {
+                return {*this};
+            }
+            meta::if_<use_sentinel_t, adaptor_base, adaptor<false>> end_adaptor()
+            {
+                return {*this};
+            }
+            CONCEPT_REQUIRES(Range<Rng const>() &&
+                             Invocable<Fun const&, range_common_reference_t<Rng>,
+                                     range_common_reference_t<Rng>>())
+            adaptor<true> begin_adaptor() const
+            {
+                return {*this};
+            }
+            CONCEPT_REQUIRES(Range<Rng const>() &&
+                             Invocable<Fun const&, range_common_reference_t<Rng>,
+                                     range_common_reference_t<Rng>>())
+            meta::if_<use_sentinel_t, adaptor_base, adaptor<true>> end_adaptor() const
+            {
+                return {*this};
+            }
+
+        public:
+            exclusive_scan_view() = default;
+            exclusive_scan_view(Rng rng, T init, Fun fun)
+            : exclusive_scan_view::view_adaptor{std::move(rng)}
+            , init_(std::move(init))
+            , fun_(std::move(fun))
+            {}
+            CONCEPT_REQUIRES(SizedRange<Rng>())
+            range_size_type_t<Rng> size() const
+            {
+                return ranges::size(this->base());
+            }
+        };
+
+        namespace view
+        {
+            struct exclusive_scan_fn
+            {
+            private:
+                friend view_access;
+                template<typename T, typename Fun = plus>
+                static auto bind(exclusive_scan_fn exclusive_scan, T init, Fun fun = {})
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                        make_pipeable(std::bind(exclusive_scan, std::placeholders::_1,
+                                                std::move(init),
+                                                protect(std::move(fun))))
+                )
+            public:
+                template<typename Rng, typename T, typename Fun>
+                using Concept = meta::and_<
+                        InputRange<Rng>,
+                        MoveConstructible<T>,
+                        IndirectInvocable<Fun, iterator_t<Rng>, iterator_t<Rng>>,
+                        ConvertibleTo<result_of_t<Fun&(T, range_common_reference_t<Rng> &&)>, T>>;
+
+                template<typename Rng, typename T, typename Fun,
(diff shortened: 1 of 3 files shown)


include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+            semiregular_t<Fun> fun_;
+            using single_pass = SinglePass<iterator_t<Rng>>;
+            using use_sentinel_t = meta::or_<meta::not_<BoundedRange<Rng>>, single_pass>;
+
+            template<bool IsConst>
+            struct adaptor : adaptor_base {
+            private:
+                using exclusive_scan_view_t = meta::const_if_c<IsConst, exclusive_scan_view>;
+                // sum_'s type doesn't have to be range_value_type_t<Rng>.
+                semiregular_t<range_value_type_t<Rng>> sum_;
```
CaseyCarter: Despite that the specification of `exclusive_scan` in the Standard is completely broken, I think the intent of [[exclusive.scan]/3.2](http://eel.is/c++draft/exclusive.scan#3.2):

> All of `binary_op(init, init)`, `binary_op(init, *first)`, and `binary_op(*first, *first)` shall be convertible to `T`.

is that the intermediate values be stored as `T`s, although I could be easily convinced otherwise.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+        /// \addtogroup group-views
+        /// @{
+        template<typename Rng, typename T, typename Fun>
+        struct exclusive_scan_view
+                : view_adaptor<exclusive_scan_view<Rng, T, Fun>, Rng> {
+        private:
+            friend range_access;
+            T init_;
+            semiregular_t<Fun> fun_;
+            using single_pass = SinglePass<iterator_t<Rng>>;
```
CaseyCarter: This view cannot be multipass without also requiring that the type of the "accumulator" `sum_` in the adaptor is constructible from `const T&` as necessary to call `begin` repeatedly.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+            using use_sentinel_t = meta::or_<meta::not_<BoundedRange<Rng>>, single_pass>;
+
+            template<bool IsConst>
+            struct adaptor : adaptor_base {
```
CaseyCarter: `{` needs its own line.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                exclusive_scan_view_t *rng_;
+            public:
+                using single_pass = exclusive_scan_view::single_pass;
+                adaptor() = default;
+                adaptor(exclusive_scan_view_t &rng)
+                        : rng_(&rng)
+                {}
+                iterator_t<Rng> begin(exclusive_scan_view_t &)
+                {
+                    sum_ = rng_->init_;
```
CaseyCarter: This requires `Constructible<uncvref_t<decltype(*sum_)>, T&>` which is unconstrained. (I think I would tag dispatch on `single_pass` here, and use `sum_ = std::move(rng_->init_)` when single pass and `sum_ = rng_->init_` when multipass to avoid overconstraining the single-pass case.)

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                    return ranges::begin(rng_->base());
+                }
+                range_value_type_t<Rng> read(iterator_t<Rng>) const
+                {
+                    return sum_;
```
CaseyCarter: This requires `CopyConstructible<unvcref_t<decltype(*sum_)>>`.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                }
+                void next(iterator_t<Rng> &it)
+                {
+                    if (it != ranges::end(rng_->base()))
```
CaseyCarter: It's a precondition violation to increment an end iterator: this should be `RANGES_EXPECT(it != ranges::end(rng_->base()));` instead of a conditional.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+            private:
+                using exclusive_scan_view_t = meta::const_if_c<IsConst, exclusive_scan_view>;
+                // sum_'s type doesn't have to be range_value_type_t<Rng>.
+                semiregular_t<range_value_type_t<Rng>> sum_;
+                exclusive_scan_view_t *rng_;
+            public:
+                using single_pass = exclusive_scan_view::single_pass;
+                adaptor() = default;
+                adaptor(exclusive_scan_view_t &rng)
+                        : rng_(&rng)
```
CaseyCarter: Ditto "only two additional spaces before `:` so that `rng_` is one four-space indent to the right of `adaptor`"

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                range_value_type_t<Rng> read(iterator_t<Rng>) const
+                {
+                    return sum_;
+                }
+                void next(iterator_t<Rng> &it)
+                {
+                    if (it != ranges::end(rng_->base()))
+                    {
+                        auto &current = static_cast<range_value_type_t<Rng> &>(sum_);
+                        sum_ = invoke(rng_->fun_, current, *it);
```
CaseyCarter: This should be `current = invoke(rng_->fun_, current, *it);`, which requires `Assignable<decltype((current)), invoke_result_t<decltype((rng_->fun_)), decltype((current)), decltype(*it)>` which is unconstrained.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+            {
+                return {*this};
+            }
+
+        public:
+            exclusive_scan_view() = default;
+            exclusive_scan_view(Rng rng, T init, Fun fun)
+            : exclusive_scan_view::view_adaptor{std::move(rng)}
+            , init_(std::move(init))
+            , fun_(std::move(fun))
```
CaseyCarter: The three above lines need two more spaces of indent.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+            {}
+            CONCEPT_REQUIRES(SizedRange<Rng>())
```
CaseyCarter: This applies `size` to a `const` range: it requires `SizedRange<Rng const>`. It wouldn't hurt to have both a `const` and non-`const` overload of `size`, since some underlying views will only provide a non-`const` `size`.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                        make_pipeable(std::bind(exclusive_scan, std::placeholders::_1,
+                                                std::move(init),
+                                                protect(std::move(fun))))
+                )
+            public:
+                template<typename Rng, typename T, typename Fun>
+                using Concept = meta::and_<
+                        InputRange<Rng>,
+                        MoveConstructible<T>,
+                        IndirectInvocable<Fun, iterator_t<Rng>, iterator_t<Rng>>,
```
CaseyCarter: You also need `IndirectInvocable<Fun, iterator_t<Rng>, T*>` here if you change the "accumulator" type to `T`.

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -0,0 +1,164 @@
...
+                        ConvertibleTo<result_of_t<Fun&(T, range_common_reference_t<Rng> &&)>, T>>;
```
CaseyCarter: I would simply toss this requirement: despite the Standard's fondness for conversion constraints, they're rarely useful in practice. Note that some of the syntax used in your implementation *may* result in conversions, but none of it *must* result in conversions.

test/view/exclusive_scan.cpp
```diff
@@ -0,0 +1,55 @@
...
+            models<concepts::ForwardView>(aux::copy(rng));
+            models_not<concepts::BidirectionalView>(aux::copy(rng));
+            ::check_equal(rng, {0, 1, 3, 6, 10, 15, 21, 28, 36, 45});
+        }
+
+        {// Test exclusive_scan with a mutable lambda
+            int cnt = 0;
+            auto mutable_rng = view::exclusive_scan(rgi, 0, [cnt](int i, int j) mutable {return i + j + cnt++;});
+            ::check_equal(mutable_rng, {0, 1, 4, 9, 16, 25, 36, 49, 64, 81});
+            CHECK(cnt == 0);
```
CaseyCarter: We probably don't need to test the core language here: let's give the compiler the benefit of the doubt.

test/view/exclusive_scan.cpp
```diff
@@ -0,0 +1,55 @@
...
+            CONCEPT_ASSERT(View<decltype(mutable_rng)>());
+            CONCEPT_ASSERT(!View<decltype(mutable_rng) const>());
+        }
+    }
+
+    {// For an emtpy range.
```
CaseyCarter: s/emtpy/empty/

include/range/v3/view/exclusive_scan.hpp
```diff
@@ -137,7 +137,7 @@ namespace ranges {
                         InputRange<Rng>,
                         MoveConstructible<T>,
```
mitsutaka-takeda: Should we change this constraint to `Semiregular<T>`? Since to hold the initial value in the view, we require the T to be a semiregular.

CaseyCarter: I'd suggest changing to `CopyConstructible<T>` which will allow `semiregular_t<T>` to be `Copyable` as needed in `adaptor::begin` when you copy the initial value into the iterator. (Theoretically, you could leave the requirement at `MoveConstructible` and write the view such that it is a `ForwardView` when `T` is `CopyConstructible` and a move-only `InputView` otherwise, but that may be more work than you're interested in investing.)

mitsutaka-takeda: Thank you for the suggestion! I'll stick with `CopyConstructible<T>` for now.

mitsutaka-takeda:
Is there anything I need to fix with this PR?

CaseyCarter:
> Is there anything I need to fix with this PR?

I've submitted a PR with some small changes into your branch.

CaseyCarter:
@ericniebler: I'm happy enough with this to merge, despite a minor case of "implementation as constraints" that is a consequence of the underspecification of the algorithm requirements in the IS. Shout before the CI completes if you feel differently.

ericniebler:
Thanks guys! Sorry I've been MIA on this PR. I've been snowed under by Committee stuff.

mitsutaka-takeda:
@ericniebler Thank you for your hard work! I'm really looking forward Range-v3 coming to C++ standard.

PR #814, opened 2018-03-28
view::drop is not drop_exactly: don't overrun random access ranges

### LINKED ISSUE #813: empty() return "wrong" value after view::drop
The following code :

```c++
    std::vector<int> list = {1,2,3,4,5};

    auto l = list | view::drop(20);
    bool b = l.empty();
    std::cout << b;
```

output 0

Is this by the design?

CaseyCarter:
Fixes #813

### DIFF
--- a/include/range/v3/view/drop.hpp
+++ b/include/range/v3/view/drop.hpp
@@ -21,6 +21,7 @@
 #include <range/v3/range_fwd.hpp>
 #include <range/v3/range_traits.hpp>
 #include <range/v3/view_interface.hpp>
+#include <range/v3/algorithm/min.hpp>
 #include <range/v3/detail/satisfy_boost_range.hpp>
 #include <range/v3/utility/box.hpp>
 #include <range/v3/utility/functional.hpp>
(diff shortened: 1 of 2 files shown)


include/range/v3/view/drop.hpp
```diff
@@ -157,7 +157,7 @@ namespace ranges
                 static iterator_range<iterator_t<Rng>, sentinel_t<Rng>>
                 invoke_(Rng && rng, range_difference_type_t<Rng> n, concepts::RandomAccessRange*)
                 {
-                    return {next(begin(rng), n), end(rng)};
+                    return {next(begin(rng), n, end(rng)), end(rng)};
```
CaseyCarter: The requirements are insufficient to guarantee that `next(begin(rng), n, end(rng))` is O(1). O(1) would additionally require `SizedRange` and (I think) `Assignable<iterator_t<Rng>&, sentinel_t<Rng>>`.

CaseyCarter:
@tahonermann I hope you find the changes satisfactory ;)

ericniebler:
Thanks, guys!

PR #821, opened 2018-04-28
Added explicit options to enable or disable testing

apmccartney:
These options are visible from the cmake-gui and cmake-curses-gui;
their presence is more discoverable than the local variables used
previously.

In addition, the default values for these options is contingent
upon whether range-v3 is the highest level project. When range-v3
is composed in another project (as a git submodule, for instance)
tests are not built and no option appears in the GUI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ if(RANGE_V3_EXAMPLES)
   add_subdirectory(example)
 endif()
 
-if(RANGE_V3_BENCHMARKS)
+if(RANGE_V3_PERF)
   add_subdirectory(perf)
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -31,15 +30,15 @@ target_link_libraries(range-v3 INTERFACE meta)
...
   add_subdirectory(test)
 endif()
 
-if(NOT RANGE_V3_NO_EXAMPLE)
+if(RANGE_V3_EXAMPLES)
   add_subdirectory(example)
 endif()
 
-if(NOT RANGE_V3_NO_PERF)
+if(RANGE_V3_BENCHMARKS)
```
CaseyCarter: `RANGE_V3_BENCHMARKS` may be overselling `perf`. (We probably shouldn't be shipping `perf` at all.) A more obscure name might steer people away and avoid wasting their time.

CaseyCarter: Yes, I think that's an improvement.

PR #822, opened 2018-04-30
Fix view::cartesian_product's cursor when adapting a non-Bounded rand…

### LINKED ISSUE #820: cartesian_product/chunk interaction
I was attempting to lazily generate a cartesian grid points when I seem to have encountered a strange error. I've included a minimal reproducing example below.

It appears a `chunk_view` of a`cartesian_product` of `iota` cannot evaluate its elements. 

[Online example](https://wandbox.org/permlink/J4MPobvysXnFcfqz) reproduced here.
```c++
#include <iostream>
#include <cstdlib>
#include <tuple>

std::ostream& operator<<(std::ostream& os, const std::tuple<int,int>& t){
  os << '(' << std::get<0>(t)
     << ',' << std::get<1>(t) << ')' << std::endl;
  return os;
}

#include <range/v3/all.hpp>

int main(){
    auto x = ranges::view::iota(0) | ranges::view::take_exactly(3);
    auto y = ranges::view::iota(0) | ranges::view::take_exactly(3);
    auto product = ranges::view::cartesian_product(x,y); 
    std::cout << product.front() << std::endl;
    auto yslices = product | ranges::view::chunk(3);
    auto y0 = yslices.front();
}
```

ericniebler:
…om-access range, fixes #820

### DIFF
--- a/include/range/v3/view/cartesian_product.hpp
+++ b/include/range/v3/view/cartesian_product.hpp
@@ -174,7 +174,7 @@ namespace ranges
                     cursor const &that, meta::size_t<1>, dist_info const inf) const
                 {
                     auto const my_distance = std::get<0>(that.its_) - std::get<0>(its_);
-                    return my_distance * inf.size_product + inf.distance;
+                    return static_cast<std::ptrdiff_t>(my_distance * inf.size_product + inf.distance);
                 }
                 template<std::size_t N>
                 std::ptrdiff_t distance_(
@@ -245,8 +245,8 @@ namespace ranges
                 cursor(end_tag, constify_if<cartesian_product_view> &view, std::false_type) // !Bounded
                   : cursor(begin_tag{}, view)
                 {
-                    std::get<0>(its_) =
-                        std::get<0>(its_) + ranges::distance(std::get<0>(view.views_));
+                    // Only called when the 0th view type is !Bounded && RandomAccess && Sized
+                    std::get<0>(its_) += ranges::distance(std::get<0>(view.views_));
                 }
             public:
                 using value_type = std::tuple<range_value_type_t<Views>...>;
--- a/include/range/v3/view/cartesian_product.hpp
+++ b/include/range/v3/view/cartesian_product.hpp
@@ -237,6 +237,17 @@ namespace ranges
                     return check_at_end_(meta::size_t<N - 1>{}, at_end ||
                         std::get<N - 1>(its_) == ranges::end(std::get<N - 1>(view_->views_)));
                 }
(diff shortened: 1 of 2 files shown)


include/range/v3/view/cartesian_product.hpp
```diff
@@ -237,6 +237,17 @@ namespace ranges
...
                         std::get<N - 1>(its_) == ranges::end(std::get<N - 1>(view_->views_)));
                 }
+                cursor(end_tag, constify_if<cartesian_product_view> &view, std::true_type) // Bounded
+                  : cursor(begin_tag{}, view)
+                {
+                    std::get<0>(its_) = ranges::end(std::get<0>(view.views_));
+                }
+                cursor(end_tag, constify_if<cartesian_product_view> &view, std::false_type) // !Bounded
+                  : cursor(begin_tag{}, view)
+                {
```
CaseyCarter: `// Only called when the 0th view type is !Bounded && RandomAccess && Sized` would save readers a few minutes determining that this is O(1).

include/range/v3/view/cartesian_product.hpp
```diff
@@ -174,7 +174,7 @@ namespace ranges
                     cursor const &that, meta::size_t<1>, dist_info const inf) const
                 {
                     auto const my_distance = std::get<0>(that.its_) - std::get<0>(its_);
-                    return my_distance * inf.size_product + inf.distance;
+                    return static_cast<std::ptrdiff_t>(my_distance * inf.size_product + inf.distance);
```
CaseyCarter: `inf.size_product` and `inf.distance` are already `ptrdiff_t`....did you get a warning for a range whose distance type is larger than `ptrdiff_t`? I can easily see this being a problem for 32-bit builds with ranges that use 64-bit difference types. The intent behind using `std::ptrdiff_t` directly - instead of something like `common_type</*...*/<Views>...>` - was to get more "clearance" for the size product.

Maybe all of the `std::ptrdiff_t` uses should be replaced with `common_type_t<std::ptrdiff_t, difference_type_t<iterator_t<Views>>...>`?

ericniebler: Yes, I got a warning, and it broke Appveyor. See <https://ci.appveyor.com/project/ericniebler/range-v3/build/1.0.592>.

I would rather get this fix in and deal with the truncation issues separately. After all, that problem is not new. (I'm not sure why we're seeing it only now.) Mind if I assign you the bug? There's no rush to fix it.

CaseyCarter: I'm fine with all of the above.

PR #825, opened 2018-05-08
Build improvements:

CaseyCarter:
* Use cmake 3.6.3 on Travis since `cmake_minimum_required(VERSION 3.6)`
* Add a new `rv3-test` interface library to hold test-specific configuration
* Define function `rv3_add_test(<name of test>, <name of executable>, <sources>...)` to simplify `CMakeLists.txt`


test/CMakeLists.txt
```diff
@@ -1,68 +1,39 @@
 include(../cmake/ranges_diagnostics.cmake)
 
+add_library(rv3-tests INTERFACE)
+target_link_libraries(rv3-tests INTERFACE range-v3)
+if(NOT RANGE_V3_TRY_THREAD)
+    target_compile_definitions(rv3-tests INTERFACE "-DRANGES_CXX_THREAD=0")
```
FloopCZ: This one surprised me a little bit, the [documentation states](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_DEFINITIONS.html#prop_tgt:COMPILE_DEFINITIONS) that definitions are in form `VAR` or `VAR=value`, but I tried this and it seems to work as well.

CaseyCarter: I had the same thought looking at the PR. Obviously it has been working for years, but it would be nice not to fix this and not rely on undocumented behavior.

CaseyCarter: I suppose technically it's been working with `add_compile_options`, not with `target_compile_definitions` - but yes, I've noticed in the past that `-DFOO` and `-DFOO=BAR` work with `target_compile_definitions` despite the documentation.

FloopCZ: CMake is full of surprises. :-)

Issue #826, opened 2018-05-09
Compilation fails with GCC 8.1.0


FloopCZ:
The corresponding outputs from the [`cmake ..`](https://github.com/ericniebler/range-v3/files/1987914/cmake.log) command and [`make`](https://github.com/ericniebler/range-v3/files/1987915/make.log) command. g++ version:

```
[floop@pear build (master)]$ g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/8.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --enable-multilib --disable-werror --enable-checking=release --enable-default-pie --enable-default-ssp
Thread model: posix
gcc version 8.1.0 (GCC)
```

cpplearner:
g++ does not like

```c++
std::vector<int> v;
v = std::move(v) | push_front({1,2,3});
```

which triggers the instantiation of `std::_Bind<push_front_fn(std::_Placeholder<1>, std::initializer_list<int>)>::operator()<std::vector<int>>`. The `const volatile` overload of such `operator()` queries `std::result_of<const volatile push_front_fn&(std::vector<int>&&, const volatile std::initializer_list<int>&)>`, which eventually calls `ranges::begin(const volatile std::initializer_list<int>&)`, which delegates to `ranges::v3::_begin_::fn::impl_`, which should cause a substitution failure in [begin_end.hpp#L65](https://github.com/ericniebler/range-v3/blob/8ccd974/include/range/v3/begin_end.hpp#L65) (because `begin(std::declval<const volatile std::initializer_list<int> &>())` is invalid), but instead causes a hard error.

[...]

PR #827, opened 2018-05-13
Force failure for begin/end on volatile initializer_list

### LINKED ISSUE #126: Extension header for is_range
The trait `is_range ` fails for e.g. `boost::container::vector`, `boost::container::static_vector`, ... 

Maybe we could have an extensions directory to allow these containers to be used with the library, e.g.:

- in `range/v3/extensions/boost/container/vector.hpp` :

  ```c++
namespace ranges {

template <class T, class Alloc>
struct is_range<boost::container::vector<T, Alloc>> : std::false_type {};

}  // namespace ranges
  ```

- in `range/v3/extensions/boost/container/static_vector.hpp` :

  ```c++
namespace ranges {

template <class T, std::size_t capacity>
struct is_range<boost::container::static_vector<T, capacity>>
 : std::false_type {};

}  // namespace ranges
  ```

I don't really know why `is_range ` fails for `boost::container::vector` tho.

CaseyCarter:
Fixes #126.

### DIFF
--- a/include/range/v3/begin_end.hpp
+++ b/include/range/v3/begin_end.hpp
@@ -41,6 +41,14 @@ namespace ranges
             template<typename T>
             void begin(T const &) = delete;
 
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 8
+            // Workaround https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85765
+            template<typename T>
+            void begin(std::initializer_list<T> volatile &) = delete;
+            template<typename T>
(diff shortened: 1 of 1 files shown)


CaseyCarter:
> Fixes #126.

Or #826, whatever. I'm certain this works around *some* issue, somewhere.

ericniebler:
If you say so...

CaseyCarter:
> If you say so...

Did I mention that it's time to trim back the Travis matrix again so we can add Clang 6 and GCC 7/8?

PR #831, opened 2018-05-23
Don't workaround clang 37556 on non-clang

CaseyCarter:
Move class template declarations from nested detail namespaces into `ranges:v3`, and - on clang only - implement friends with the same names as CPOs in those same detail namespaces. This affects:

* `tagged`
* `basic_iterator`
* `common_iterator`
* `counted_iterator`

NOTE TO REVIEWERS: This changes indentation for some big chunks of code, you may want to ignore whitespace changes to focus on the actual differences.

### DIFF
--- a/include/range/v3/algorithm/stable_sort.hpp
+++ b/include/range/v3/algorithm/stable_sort.hpp
@@ -48,7 +48,6 @@
 #include <range/v3/utility/iterator_concepts.hpp>
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/functional.hpp>
-#include <range/v3/utility/counted_iterator.hpp>
 #include <range/v3/algorithm/merge.hpp>
 #include <range/v3/algorithm/min.hpp>
 #include <range/v3/algorithm/sort.hpp>
--- a/include/range/v3/utility/basic_iterator.hpp
+++ b/include/range/v3/utility/basic_iterator.hpp
@@ -352,20 +352,22 @@ namespace ranges
             using box<T>::get;
         };
 
-        /// \cond
-        namespace _basic_iterator_
-        {
-        /// \endcond
+#if RANGES_BROKEN_CPO_LOOKUP
+        namespace _basic_iterator_ { template <typename> struct adl_hook {}; }
+#endif
 
         template<typename Cur>
         struct basic_iterator
           : range_access::mixin_base_t<Cur>
           , detail::iterator_associated_types_base<Cur>
+#if RANGES_BROKEN_CPO_LOOKUP
+          , private _basic_iterator_::adl_hook<basic_iterator<Cur>>
+#endif
         {
         private:
-            template<typename Cur2>
+            template<typename>
             friend struct basic_iterator;
-            friend struct ranges::range_access;
+            friend range_access;
             using mixin_t = range_access::mixin_base_t<Cur>;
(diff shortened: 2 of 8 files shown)


CaseyCarter:
Note that the clang bug only triggers for hidden friends of class templates. `any`, for example, has a `friend swap` that is just fine since it is a non-template class.

ericniebler:
I never reported this bug? /facepalm

CaseyCarter:
> I never reported this bug? /facepalm

It was during the "grand refactor to match the TS" in Kona. You rewrote most everything in a week. I forgive you.

I'm working on cleaning up the Travis failures now - I should have tested older GCCs before bothering to PR this.

include/range/v3/algorithm/stable_sort.hpp
```diff
@@ -48,7 +48,6 @@
 #include <range/v3/utility/iterator_concepts.hpp>
 #include <range/v3/utility/iterator_traits.hpp>
 #include <range/v3/utility/functional.hpp>
-#include <range/v3/utility/counted_iterator.hpp>
```
CaseyCarter: Drive-by: this algorithm no longer uses `counted_iterator`.

include/range/v3/utility/basic_iterator.hpp
```diff
@@ -352,18 +352,22 @@ namespace ranges
...
+        namespace _basic_iterator_ { template <typename> struct adl_hook {}; }
+#endif
 
         template<typename Cur>
         struct basic_iterator
           : range_access::mixin_base_t<Cur>
           , detail::iterator_associated_types_base<Cur>
+#if RANGES_BROKEN_CPO_LOOKUP
+            // Workaround https://bugs.llvm.org/show_bug.cgi?id=37556
+          , _basic_iterator_::adl_hook<basic_iterator<Cur>>
```
CaseyCarter: `private`

include/range/v3/utility/basic_iterator.hpp
```diff
@@ -352,18 +352,22 @@ namespace ranges
             using box<T>::get;
         };
 
-        /// \cond
-        namespace _basic_iterator_
-        {
-        /// \endcond
+#if RANGES_BROKEN_CPO_LOOKUP
+        // Workaround https://bugs.llvm.org/show_bug.cgi?id=37556
```
CaseyCarter: All of these `// Workaround BUG#` comments should be centralized into the definition of `RANGES_BROKEN_CPO_LOOKUP` in `config.hpp` instead of being spammed throughout the source.

include/range/v3/utility/basic_iterator.hpp
```diff
@@ -352,18 +352,22 @@ namespace ranges
...
+#if RANGES_BROKEN_CPO_LOOKUP
+            // Workaround https://bugs.llvm.org/show_bug.cgi?id=37556
+          , private _basic_iterator_::adl_hook<basic_iterator<Cur>>
+#endif
         {
         private:
-            template<typename Cur2>
+            template<typename>
             friend struct basic_iterator;
             friend struct ranges::range_access;
```
CaseyCarter: Remove `struct` here.

CaseyCarter:
AppVeyor failure is transient. (You should go to https://ci.appveyor.com/team and add me as a collaborator so I can manage AppVeyor builds.)

PR #833, opened 2018-05-26
Clang/C2 is dead - long live ClangCL!

### LINKED ISSUE #765: Update AppVeyor CI to use Clang/LLVM
...instead of Clang/C2. Microsoft has stopped maintaining Clang/C2, we should drop support as well.

This needs some corresponding updates to `README.md`.

CaseyCarter:
Individual commit logs:
* LLVM 28385 is fixed in clang 6

* Always implement `RANGES_ASSERT` on MSFTL

  `assert` failures don't necessarily terminate the program on Windows. In addition to being silly, this is problematic for range-v3 since we sometimes write:
  ```c++
  [[noreturn]] void foo() {
    RANGES_EXPECT(false);
  }
  ```
  and clang will complain about this `[[noreturn]]` function returning.

* Remove bogus `span` test

  `std::array`'s iterators aren't necessarily pointers.

* Remove specialization of `std::is_integral` in safe_int

   No, no, no! BAD!

   (Also, don't check `is_literal` since it's deprecated in C++17.)

* Support clang-cl, drop ClangC2 like a hot potato

   * Travis changes:
     * Define `_SILENCE_CXX17_TEMPORARY_BUFFER_DEPRECATION_WARNING` when building on MSFTL to silence deprecation warnings for `std::(get|return)_temporary_buffer`.
     * Some warning flags have different meaning to clang-cl than GCC and vanilla clang. We don't want `-Wall`, for example, which is the same as vanilla `-Weverything`.
     * Clang-CL apparently tries to be brain-damage compatible with MSVC - `iter_move` will recurse without the unqualified lookup block.
   
   * AppVeyor changes:
     * Use the default cmake; 3.6 does some stupid things with Clang-CL
     * Ninja is roughly 774289343928 times faster than msbuild
     * Test {x86,x64}x{Debug,Release}x{C++latest} (I honestly have no idea if clang-cl treats /std:c++latest as C++17 or C++2a.)
   
   Fixes #765.
   
* Workaround MSFTL tuple issues

   MSFTL `tuple` is recursive: `tuple<int, char>` derives from `tuple<char>` derives from `tuple<>`. Consequently, types derived from `tuple` - like `common_tuple` and `tagged<tuple>` - can't be passed to functions that deduce `tuple<Ts...>` because the deduction is ambiguous. This is a HUGE PITA, but can't be fixed until the next ABI break.

   The workaround is to define non-member `get` for range-v3 tuple types, use `ranges::get` when the concrete type of a tuple-like is known, and `ranges::detail::adl_get` when the concrete type is not known.

   We also need to define converting constructors from `common_tuple` to `common_tuple` since they can't implicitly convert to `std::tuple` and use the `std::tuple` converting constructors.

### DIFF
--- a/cmake/ranges_flags.cmake
+++ b/cmake/ranges_flags.cmake
@@ -18,13 +18,21 @@ endmacro()
 # All compilation flags
 # Language flag: version of the C++ standard to use
 message("[range-v3]: C++ std=${RANGES_CXX_STD}")
-ranges_append_flag(RANGES_HAS_CXXSTD "-std=c++${RANGES_CXX_STD}")
+if (RANGES_CXX_COMPILER_CLANGCL)
+  ranges_append_flag(RANGES_HAS_CXXSTDCOLON "-std:c++${RANGES_CXX_STD}")
+else()
+  ranges_append_flag(RANGES_HAS_CXXSTD "-std=c++${RANGES_CXX_STD}")
+endif()
 
 # Enable "normal" warnings and make them errors:
-ranges_append_flag(RANGES_HAS_WERROR -Werror)
-ranges_append_flag(RANGES_HAS_WX -WX)
-ranges_append_flag(RANGES_HAS_WALL -Wall)
-ranges_append_flag(RANGES_HAS_WEXTRA -Wextra)
+if (RANGES_CXX_COMPILER_CLANGCL)
+  ranges_append_flag(RANGES_HAS_W3 -W3)
+  ranges_append_flag(RANGES_HAS_WX -WX)
+else()
+  ranges_append_flag(RANGES_HAS_WALL -Wall)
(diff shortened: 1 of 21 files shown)


cmake/ranges_flags.cmake
```diff
@@ -18,13 +18,21 @@ endmacro()
 # All compilation flags
 # Language flag: version of the C++ standard to use
 message("[range-v3]: C++ std=${RANGES_CXX_STD}")
-ranges_append_flag(RANGES_HAS_CXXSTD "-std=c++${RANGES_CXX_STD}")
+if (RANGES_CXX_COMPILER_CLANGCL)
+  ranges_append_flag(RANGES_HAS_CXXSTDCOLON "-std:c++${RANGES_CXX_STD}")
```
ericniebler: Everybody's gotta be different, don't they?

PR #836, opened 2018-05-30
Cleanup CPO declarations:

CaseyCarter:
* Eradicate the inline namespace `function_objects`. `RANGES_INLINE_VARIABLE` puts variables directly in the containing namespace when inline variables are available, and in an inline unnamed namespace otherwise.

  * *Explicitly* place CPOs that are truly customization points in an inline namespace `CPOs` inside `ranges::v3` (which keeps them from colliding with hidden friend functions of the same name)

  (Note that `advance` is a "true" CPO because `counted_iterator` defines a hidden friend overload in namespace `ranges::v3`.)

Bundle of dozens of drive-by changes:

* Implement `meta::at` with `__type_pack_element` when available

* Use wg21.link for LWG issues so we don't need to care about active vs. closed vs. defect [NFC]

* Wrap `view/all.hpp` at 100 [NFC]

* `experimental/view/shared.hpp`: `view::shared` requires the user to pass a container *rvalue*, the user doesn't care what if any reference type we bind to that argument. [NFC]

* `meta.hpp`: `s/#if defined(X)/#ifdef X/`

* `meta.cpp`: GCC #64970 has been fixed in GCC 5+

* Copypasta in `swap.hpp`

* MSVC-friendly feature detection/configuration (Use `_MSVC_LANG` when it's defined instead of `__cplusplus`)

* Allows users to override detection in meta as was already the case for range-v3.

* Don't define components - even deprecated ones - in terms of deprecated components.

* Simplify `CONCEPT_REQUIRES` macros

* Restrict the `gcc_bugs_bugs_bugs` workaround to affected versions of GCC 5.

* Remove extraneous `basic_iterator` constructor (we inherit the mixin's cursor constructors - there's no need to implement one.)

* `basic_mixin` constructors from `T` (and consequently `basic_iterator`'s cursor constructors) should be `explicit`

* `const`-ify source range in alg.copy test

* Sprinkle some `RANGES_CONSTEXPR_IF` in `ranges::equal`

* use simplified friend declarations in `ranges::variant` [NFC]

* Simplify `meta::as_list`: `curry<quote_trait<id>>` expands to `compose<quote_trait<id>, quote<list>>` which is equivalent to `quote<list>` [Almost NFC, negligible compile time reduction.]

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Release Notes:
   - Add an experimental `shared` view for views that need container-like scratch
     space to do their work.
   - Faster, simpler `reverse_view`.
-  - Rework `ranges::reference_wrapper` to avoid [LWG\#2993](http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#2993).
+  - Rework `ranges::reference_wrapper` to avoid [LWG\#2993](https://wg21.link/lwg2993).
   - Reworked `any_view`, the type-erased view wrapper.
   - `equal` algorithm is `constexpr` in C++14.
   - `stride_view` no longer needs an `atomic` data member.
@@ -107,7 +107,7 @@ Release Notes:
   - Various portability fixes for gcc and clang trunk.
 * **0.3.0** June 30, 2017
   - Input views may now be move-only (from @CaseyCarter)
-  - Input `any_view`s are now *much* more efficicient (from @CaseyCarter)
+  - Input `any_view`s are now *much* more efficient (from @CaseyCarter)
   - Better support for systems lacking a working `<thread>` header (from @CaseyCarter)
 * **0.2.6** June 21, 2017
   - Experimental coroutines with `ranges::experimental::generator` (from @CaseyCarter)
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -17,11 +17,11 @@
 
 #include <cstddef>
 #include <initializer_list>
-#include <meta/meta_fwd.hpp>
 #include <type_traits>
 #include <utility>
+#include <meta/meta_fwd.hpp>
 
-#if defined(__clang__)
+#ifdef __clang__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunknown-pragmas"
 #pragma GCC diagnostic ignored "-Wpragmas"
@@ -139,7 +139,7 @@ namespace meta
         template <typename T>
         using _t = typename T::type;
 
-#if defined(__cpp_variable_templates) || defined(META_DOXYGEN_INVOKED)
+#if META_CXX_VARIABLE_TEMPLATES || defined(META_DOXYGEN_INVOKED)
         /// Variable alias for \c T::type::value
         /// \note Requires C++14 or greater.
         /// \ingroup invocation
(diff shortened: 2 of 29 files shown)


ericniebler:
Thanks. With your recent changes, are you by chance working toward MSVC support?

CaseyCarter:
> Thanks. With your recent changes, are you by chance working toward MSVC support?

Yeah, I've been doing a deep dive to determine feasibility and checking in things I drive by while localizing bug repros. I don't think 15.8 is going to be it, though. There are lots of little bugs like:
```c++
void foo(int&& i) {
    // Writable says hello:
    const_cast<int const&&>(static_cast<int&&>(i)); // error C2102: '&' requires l-value??!?
}

void bar() {
    int a[10];
    auto&& c = true ? a : a;

    // "true ? a : a" is an lvalue int[10], right?
    static_assert(std::is_same_v<decltype(c), int(&)[10]>); // Yes it is

    // except inside decltype...
    static_assert(std::is_same_v<decltype(true ? a : a), int(&)[10]>); // Fails
    // ...where it's a prvalue int[10]
    static_assert(!std::is_same_v<decltype(true ? a : a), int[10]>);   // Fails
}
```
that someone on the compiler team could probably fix in an hour or two, but alias templates - and specifically compositions of alias templates with parameter pack expansions - are just a dumpster fire. 
There's a full rewrite of alias templates in progress that I think is the last missing piece for me to be able to upstream reasonable changes.

For an idea of what I consider to be *un*-reasonable changes,
https://github.com/ericniebler/range-v3/compare/master...CaseyCarter:msvc has the changes from my investigation thus far. With those ~500 lines of changes, 26% of the tests will build.  Things like this `variant.hpp` change:
```diff
+#ifdef _MSC_VER // Workaround VSO#206790
+            template<typename... Ts>
+            struct _variant_data_helper_
+            {
+                using type = meta::as_list<meta::make_index_sequence<sizeof...(Ts)>>;
+            };
+
+            template<typename... Ts>
+            using variant_data =
+                meta::_t<variant_data_<
+                    meta::transform<
+                        meta::list<Ts...>,
+                        meta::_t<_variant_data_helper_<Ts...>>,
+                        meta::quote<indexed_datum>>>>;
+#else
             template<typename... Ts>
             using variant_data =
                 meta::_t<variant_data_<
                     meta::transform<
                         meta::list<Ts...>,
                         meta::as_list<meta::make_index_sequence<sizeof...(Ts)>>,
                         meta::quote<indexed_datum>>>>;
+#endif
```
are particularly frightening. I can't find a way to localize the workaround for this composition-of-aliases bug into one of the components. Without a way to "harden" the individual components in `meta`, I can't predict how they'll behave when composed. It's unreasonable that someone (me) may have to spend half a day localizing and devising a workaround for a metaprogramming bug or three every time we touch range-v3. Even if I work up a couple of thousand lines of changes that will suffice to compile the range-v3 tests it's extremely likely that users will see bugs galore if they use `meta` and/or the concepts emulation in ways that the tests don't.

IMO the necessary changes are both too invasive and too fragile to upstream for now.

ericniebler:
Blech. I agree, it's not quite there yet. Thanks for all your efforts, tho. 👍

PR #837, opened 2018-05-31
Cleanup:

CaseyCarter:
* Fix broken guards for LLVM 28385 workarounds in `meta.hpp`
* Simplify `iter_zip_with_fn::Concept`

### DIFF
--- a/include/range/v3/view/zip_with.hpp
+++ b/include/range/v3/view/zip_with.hpp
@@ -321,8 +321,8 @@ namespace ranges
             {
                 template<typename Fun, typename ...Rngs>
                 using Concept = meta::and_<
-                    meta::and_<InputRange<Rngs>...>,
-                    CopyConstructible<Fun>(),
+                    InputRange<Rngs>...,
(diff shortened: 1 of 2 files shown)


include/range/v3/view/zip_with.hpp
```diff
@@ -321,8 +321,8 @@ namespace ranges
             {
                 template<typename Fun, typename ...Rngs>
                 using Concept = meta::and_<
-                    meta::and_<InputRange<Rngs>...>,
-                    CopyConstructible<Fun>(),
+                    InputRange<Rngs>...,
```
ericniebler: Funny, I'm certain there was a reason I needed the inner `meta::and_`, but from the passing tests it seems that it's no longer needed.

include/range/v3/view/zip_with.hpp
```diff
@@ -321,8 +321,8 @@ namespace ranges
...
+                    CopyConstructible<Fun>,
```
ericniebler: Egad.

PR #843, opened 2018-06-06
Fix #839 Macro to detect clang version on Xcodes Toolchain.

### LINKED ISSUE #839: [meta] detail::find_if_ breaks on apple clang
recently @CaseyCarter added

```
#if defined(__clang__) && __clang_major__ < 6
                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
                static constexpr bool s_v[sizeof...(List)] = {invoke<Fun, List>::type::value...};
#else
                static constexpr bool s_v[] = {invoke<Fun, List>::type::value...};
#endif
```

But this fails with apples clang:

```
clang++ --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

```

where `___clang_major__` is `9`.

maikel:
I propose this fix to issue #839

### DIFF
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -2139,10 +2139,11 @@ namespace meta
             struct find_if_<list<List...>, Fun,
                             void_<integer_sequence<bool, bool(invoke<Fun, List>::type::value)...>>>
             {
-#if defined(__clang__) && __clang_major__ < 6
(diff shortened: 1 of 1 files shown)


maikel:
I had a typo. Will fix it fast.

PR #844, opened 2018-06-06
Fix #839 macro to the clang version for MacOs

### LINKED ISSUE #839: [meta] detail::find_if_ breaks on apple clang

maikel:
This should fix #839

### DIFF
--- a/include/meta/meta.hpp
+++ b/include/meta/meta.hpp
@@ -1990,8 +1990,9 @@ namespace meta
             template <typename... T, typename V>
             struct find_index_<list<T...>, V>
             {
-#if defined(__clang__) && __clang_major__ < 6
+#if (defined(__clang__) && __clang_major__ < 6) || defined(__apple_build_version__)
                 // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
+                // Track Apple's version here: https://en.wikipedia.org/wiki/Xcode#Toolchain_Versions
                 static constexpr bool s_v[sizeof...(T)] = {std::is_same<T, V>::value...};
 #else
                 static constexpr bool s_v[] = {std::is_same<T, V>::value...};
@@ -2047,8 +2048,9 @@ namespace meta
             template <typename... T, typename V>
             struct reverse_find_index_<list<T...>, V>
             {
-#if defined(__clang__) && __clang_major__ < 6
+#if (defined(__clang__) && __clang_major__ < 6) || defined(__apple_build_version__)
                 // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
+                // Track Apple's version here: https://en.wikipedia.org/wiki/Xcode#Toolchain_Versions
                 static constexpr bool s_v[sizeof...(T)] = {std::is_same<T, V>::value...};
 #else
                 static constexpr bool s_v[] = {std::is_same<T, V>::value...};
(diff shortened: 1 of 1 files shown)


johelegp:
For future reference, you could have reused the other pull request by force pushing the changes, i.e., `git push -f`.

include/meta/meta.hpp
```diff
@@ -2139,8 +2139,9 @@ namespace meta
             struct find_if_<list<List...>, Fun,
                             void_<integer_sequence<bool, bool(invoke<Fun, List>::type::value)...>>>
             {
-#if defined(__clang__) && __clang_major__ < 6
+#if (defined(__clang__) && __clang_major__ < 6) || defined(__apple_build_version__)
```
CaseyCarter: There are roughly half a dozen occurrences of `// Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385` in here which need to be fixed similarly.

include/meta/meta.hpp
```diff
@@ -2139,8 +2139,9 @@ namespace meta
...
                 // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
+                // Track Apples version here: https://en.wikipedia.org/wiki/Xcode#Toolchain_Versions
```
CaseyCarter: "Apples" should be "Apple's"

include/meta/meta.hpp
maikel: Right, I haven't looked for more occurrences. My code base used `meta::find_if` directly at several places. I've searched this file for the string '28385' and fixed it at every place which I've found that way.

    grep -Rn '28385' .
    ./meta/meta.hpp:1994:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
    ./meta/meta.hpp:2052:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
    ./meta/meta.hpp:2145:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
    ./meta/meta.hpp:2206:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
    ./meta/meta.hpp:2334:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
    ./meta/meta.hpp:2379:                // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385

include/meta/meta.hpp
```diff
@@ -1990,8 +1990,9 @@ namespace meta
             template <typename... T, typename V>
             struct find_index_<list<T...>, V>
             {
-#if defined(__clang__) && __clang_major__ < 6
+#if (defined(__clang__) && __clang_major__ < 6) || defined(__apple_build_version__)
                 // Explicitly specify extent to avoid https://llvm.org/bugs/show_bug.cgi?id=28385
+                // Track Apple's version here: https://en.wikipedia.org/wiki/Xcode#Toolchain_Versions
```
CaseyCarter: Having now actually looked at this page, there's no useful information there: I don't think we need these Wikipedia links. We like having links to trackable bugs for workarounds like this so the workarounds can be scoped to the buggy compiler versions and eventually dropped along with support for said buggy compiler versions. I vaguely recall hearing someone mention that Apple doesn't have a publicly accessible bug tracking database so there may be no way for us to get such a link in this case. If you happen to know that I'm mistaken, then please speak up, but otherwise we should simply remove the wikipedia links and hope that someone some day will tell us this bug has been fixed in some Apple clang version.

maikel: Alright, I somehow assumed the link provided a mapping of Apple's LLVM version to the official one. But I was apparently mistaken as it doesn't include newer versions.

Issue #845, opened 2018-06-06
call to 'advance' is ambiguous


DarkWingMcQuack:
When I use the `accumulate` algorithm in my project I get the following error:
```
In file included from /home/XXX/Projects/kylec/src/analysis/lookup/FunctionLookup.cpp:1:
In file included from /home/XXX/Projects/kylec/include/analysis/convertability/CastConvertability.hpp:3:
In file included from /home/XXX/Projects/kylec/include/analysis/convertability/GenericConvertability.hpp:11:
In file included from /home/XXX/Projects/kylec/external/range_v3/include/range/v3/all.hpp:17:
In file included from /home/XXX/Projects/kylec/external/range_v3/include/range/v3/core.hpp:17:
In file included from /home/XXX/Projects/kylec/external/range_v3/include/range/v3/begin_end.hpp:27:
/home/XXX/Projects/kylec/external/range_v3/include/range/v3/utility/iterator.hpp:112:21: error: call to 'advance' is ambiguous
                    advance(i, n);
```

candidates are:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/6.4.0/include/g++-v6/bits/stl_iterator_base_funcs.h:194:5: note: candidate function [with _InputIterator = __gnu_cxx::__normal_iterator<const ~LongLongType~, _Distance = long]
    advance(_InputIterator& __i, _Distance __n)
```

and 

```
/usr/include/boost/iterator/advance.hpp:68:9: note: candidate function [with InputIterator = __gnu_cxx::__normal_iterator<const ~LongLongType~, Distance = long]
        advance(InputIterator& it, Distance n)
```
I dont realy know if I accidentialy populated the namespace, but if not this could be easily fixed

DarkWingMcQuack:
https://github.com/ericniebler/range-v3/pull/846

maikel:
Can you provide more context? I do not understand why it would choose `boost::advance` here. Do you use `using namespace boost` or `using boost::advance` somehwere?

DarkWingMcQuack:
I did a grep through my entire project and I am never using `using namespace boost` or `using boost::advance`.
I am using `using namespace ranges` quite often, could that be a problem?

CaseyCarter:
[...]

PR #846, opened 2018-06-07
removed `using std::advance` and replace advance with std::advance in…

DarkWingMcQuack:
… code to prevent clashes with boosts advance


DarkWingMcQuack:
This should fix https://github.com/ericniebler/range-v3/issues/845

DarkWingMcQuack:
nevermind I have seen that this does not work, I am sorry

Issue #849, opened 2018-06-18
Feature request: Improve conan package


pleroux0:
## Feature request

Change conan package to use cmake install so that the cmake targets can be used in a cmake package. 

This allows for consumer packages to use modern cmake in their build systems:

```cmake
find_package(range-v3 REQUIRED)
target_link_libraries(my_target PRIVATE range-v3)
```
instead of

```cmake
target_include_directories(my_target PRIVATE ${CONAN_INCLUDE_DIRS_RANGE-V3})
```

If a consumer uses the `cmake_paths` generator one can even skip the usual

```cmake
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
conan_basic_setup()
```

which conan requires with the `cmake` generator. Making the consumer project's use of conan transparent.

I would be happy to make a pull request if you give the go ahead.

ericniebler:
I believe this is resolved fixed. @pleroux0 please reopen if there is still more to do here.

PR #850, opened 2018-06-19
Feature/improve conan integration

pleroux0:
This PR addresses #849 

The settings line in the conanfile.py is needed until conan-io/conan#3011 is released.

I also added a switch to change whether the documentation is built or not since it did not appear optional if doxygen was found.

If any changes are required just let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,9 @@ set(CMAKE_REQUIRED_FLAGS "-fcoroutines-ts")
 check_cxx_source_compiles("${RANGE_V3_CORO_TEST_CODE}" RANGE_V3_HAS_FCOROUTINES_TS)
 unset(CMAKE_REQUIRED_FLAGS)
 
-add_subdirectory(doc)
+if(RANGE_V3_DOCS)
+  add_subdirectory(doc)
+endif()
 
 if(RANGE_V3_TESTS)
   add_subdirectory(test)
--- a/cmake/ranges_options.cmake
+++ b/cmake/ranges_options.cmake
@@ -44,4 +44,8 @@ CMAKE_DEPENDENT_OPTION(RANGE_V3_PERF
   "Build the Range-v3 performance benchmarks"
   ON "${is_subproject}" OFF)
 
+CMAKE_DEPENDENT_OPTION(RANGE_V3_DOCS
+  "Build the Range-v3 documentation"
+  ON "${is_subproject}" OFF)
+
 mark_as_advanced(RANGE_V3_PERF)
--- a/conanfile.py
+++ b/conanfile.py
@@ -17,6 +17,7 @@ class Rangev3Conan(ConanFile):
     license = "Boost Software License - Version 1.0 - August 17th, 2003"
     url = "https://github.com/ericniebler/range-v3"
     description = """Experimental range library for C++11/14/17"""
+    settings = "compiler", "arch"
     exports_sources = "include*", "LICENSE.txt", "CMakeLists.txt", "cmake/*", "Version.cmake", "version.hpp.in"
     build_policy = "missing"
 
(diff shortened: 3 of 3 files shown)


ericniebler:
Apologies for not getting back to you sooner. Does this place a version requirement on conan?

pleroux0:
No worries. It only becomes relevant for the next release and I imagine you've been pretty busy with getting the ranges into the C++ standard.

The settings workaround makes it work with conan 1.4 and earlier. I haven't tested further back, but from the conan changelog: 0.25.0 introduced cmake.install() which the new script uses. Is there a specific required version the library aims for?

ericniebler:
I just tried this out and got the following error:

```txt
-- Conan: called by CMake conan helper
CMake Error at build/7dd9e1a83ffe63c45af7c671c6032d748dc83894/conanbuildinfo.cmake:349 (message):
  Incorrect 'apple-clang' version 'compiler.version=9.0' is not the one
  detected by CMake: 'Clang=9.1'
  build/7dd9e1a83ffe63c45af7c671c6032d748dc83894/conanbuildinfo.cmake:406 (conan_error_compiler_version)
  build/7dd9e1a83ffe63c45af7c671c6032d748dc83894/conanbuildinfo.cmake:477 (check_compiler_version)
  build/7dd9e1a83ffe63c45af7c671c6032d748dc83894/conanbuildinfo.cmake:114 (conan_check_compiler)
  CMakeLists.txt:17 (conan_basic_setup)


-- Configuring incomplete, errors occurred!
See also "/Users/eniebler/Code/range-v3/test_package/build/7dd9e1a83ffe63c45af7c671c6032d748dc83894/CMakeFiles/CMakeOutput.log".
ERROR: PROJECT: Error in build() method, line 24
	cmake.configure()
	ConanException: Error 256 while executing cd '/Users/eniebler/Code/range-v3/test_package/build/7dd9e1a83ffe63c45af7c671c6032d748dc83894' && cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE="Release" -DCONAN_EXPORTED="1" -DCONAN_COMPILER="apple-clang" -DCONAN_COMPILER_VERSION="9.0" -DCONAN_CXX_FLAGS="-m64" -DCONAN_SHARED_LINKER_FLAGS="-m64" -DCONAN_C_FLAGS="-m64" -DCONAN_LIBCXX="libc++" -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY="ON" -Wno-dev '/Users/eniebler/Code/range-v3/test_package'
```

pleroux0:
The conan default profile is set to `apple-clang` version `9.0` where your compiler has been updated to `9.1`. 

Try running the following to update the default profile:

`conan profile update settings.compiler.version=9.1 default`

agurtovoy:
This PR assumes that everyone who is *installing* range-v3 conan package is using CMake/has CMake installed. Typically even the libraries that use CMake as their build system do not require CMake for consumption of their conan packages, unless you explicitly request building from source. Unconditionally using CMake in the `install` method effectively forces everyone who wants to use range-v3 conan package to install CMake even though there is nothing to build here.

I'm not a conan expert, but from what I understand the right way to resolve this would be to introduce a package option that explicitly enables cmake targets generation; then whoever needs this feature can easily enable it when they do `conan install`:

```
conan install .. -o *:cmake_targets=True
```

Thoughts?

ericniebler:
Sounds ok to me, but I'm not a Conan expert. Thanks. PR?

agurtovoy:
Yep, working on a PR.

agurtovoy:
@ericniebler After looking into this more, I realized it was the settings workaround (f24dca18fae) + ` build_policy = "missing"` that were triggering an unconditional CMake "rebuild" on package installation for any arch/compiler different from the ones used to publish the package. The issue that prompted the workaround has since been fixed, so I simply reworked the recipe to follow Conan's current recommendations for header only libraries — see #1236.

Issue #875, opened 2018-08-13
Slice from End


tete17:
Hi everyone:

I really wanted to be collaborated with this project so i have been looking through the code and found a TODO:

`action/slice.hpp:64`

I though it could be a good starting point and i wanted to write this issue to clarify a couple of things before getting my hands dirty, and also have a ticket to reference against.

 * First: Actual behaviour. From the comment the behaviour i understand is something similar to 
```
std::vector rng{1,2,3,4,5,6,7,8,9,10};

std::vector sliced = rng | action::slice_from_back(1, 3)

ASSERT(sliced, {9,10}); 

```
 * Secondly naming: I used `slice_from_back` in the example but maybe you guys had something else in mind, please let me know;

Is there is anything i understood incorrectly please let me know. 
Also tips for newbies  in ranges (not in c++) are welcome :smile:

ericniebler:
I think this one was fixed in #876. Closing.

Issue #802, opened 2018-02-25
Wishlist for range-v3 version 0.4


ericniebler:
- [ ] Redesigned range access (`begin`, `end`, etc.) customization points and `safe_iterator_t`.
- [ ] Optionally use C++20 concepts when available.
- [ ] Use SFINAE-in-return-type instead of SFINAE-in-template-argument-list to avoid [LLVM\#36158](https://bugs.llvm.org/show_bug.cgi?id=36158)
- [ ] Better diagnostics for failed requirements checks.
- [ ] Views conforming to [P0789](http://wg21.link/P0789).
- [ ] Handle `const`/non-`const` view iteration like in P0789 (see <tt><i>simple-view</i>&lt;Rng&gt;</tt>).
- [ ] Replace `iterator_range` with `subrange`.
- [ ] Algorithms return `subrange<>` or `dangling<subrange<>>` where appropriate.
- [ ] Class template argument deduction where appropriate.
- [ ] Add `ref_view` / remove hack-ish `begin`/`end` support for `reference_wrapper`.
- [ ] Comparison function objects (`less`, etc.) conform to the Ranges TS.
- [ ] Require C++14(?)
- [ ] _more?_

Generally, make a pass through the Ranges TS and the proposals in-flight, find differences between them and range-v3, and _squash them!_

@CaseyCarter Can you think of more interface-breaking changes we should lump into a major new range-v3 release?

CaseyCarter:
That's a reasonable set of features. We should try to coordinate with clang folk on 36158: we don't want to invest in a workaround if the bug is on someone's plate to investigate in the next few months.

I think we should be violently militant and make `range_size_type_t` an alias for `range_difference_type_t`. range-v3 can "fix" broken ranges by giving them a properly signed size ;)

I'd like to pull `optional` into the `detail` namespace - that's interface-breaking. In a perfect world, we would stop using `optional<T&>`/`variant<T&>` so we can use the standard components instead of our implementations when they're available.

ericniebler:
> In a perfect world, we would stop using `optional<T&>`/`variant<T&>` so we can use the standard components instead of our implementations when they're available.

We'll have more problems moving away from my home-rolled `variant`: the `visit_i` interface is used in places and not supported by `std::variant`.

CaseyCarter:
[...]

PR #876, opened 2018-08-16
Action slice from end

tete17:
Ok silly me in my issue #875. 

I soon found out what the meaning of "from end" was and avoided creating a whole new `action::` element.

 I would ask please if someone with a bit more experience in ranges concepts took a look at my requirements. In the case of "slice from end" i asked for `BidirectionalRanges` instead of `SizedRanges` thinking it would be less restrictive. Im not sure about this decision so please let me know. 

Everything else hopefully is fine. 

Eager to hear your feedback guys :smile:


tete17:
Hi this build broke because valgrind was broken on mac and i can't reschedule it

tete17:
Seems like the valgrind package of homebrew is broken and it does not install.
This error has nothing to do with my changes
Ref. https://bugs.kde.org/show_bug.cgi?id=365327

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
-                    return static_cast<Rng&&>(rng);
+                    return static_cast<Rng &&>(rng);
                 }
 
+              template<typename Rng,
+                  typename I = iterator_t<Rng>,
+                  typename D = range_difference_type_t<Rng>,
+                  CONCEPT_REQUIRES_(BidirectionalRange<Rng>() && ErasableRange<Rng, I, I>())>
+              Rng operator()(Rng &&rng, D from, detail::from_end_<D> to) const {
+                  RANGES_EXPECT(0 <= from && to.dist_ <= 0);
```
CaseyCarter: `RANGES_EXPECT(!SizedRange<Rng> || from - to.dist_ <= distance(rng));` wouldn't hurt here. Or even the potentially O(N) `RANGES_EXPECT(from - to.dist_ <= distance(rng));` since we're about to perform O(N) erasures anyway.

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
+                  ranges::action::erase(rng, begin(rng), next(begin(rng), from));
+                  ranges::action::erase(rng, prev(end(rng), -to.dist_), end(rng));
```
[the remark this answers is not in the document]
CaseyCarter: Should we make this line conditional on `to.dist_ != 0` to avoid the cost of finding the end iterator when it's not needed?

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
+                  return static_cast<Rng &&>(rng);
+              }
+
+              template<typename A>
+              using is_end_type = meta::or_<meta::is<A, detail::from_end_>, std::is_same<A, end_fn>>;
+
             #ifndef RANGES_DOXYGEN_INVOKED
                 template<typename Rng, typename T, typename U,
-                    CONCEPT_REQUIRES_(!Concept<Rng, T, U>())>
+                    CONCEPT_REQUIRES_(!ForwardRange<Rng>())>
```
CaseyCarter: This isn't constrained to match exactly the cases that the other overloads do not admit.

include/range/v3/action/slice.hpp
```diff
@@ -86,11 +139,11 @@ namespace ranges
...
                         "removal.");
-                    CONCEPT_ASSERT_MSG(meta::and_<ConvertibleTo<T, range_difference_type_t<Rng>>,
-                            ConvertibleTo<U, range_difference_type_t<Rng>>>(),
-                        "The bounds passed to action::slice must be convertible to the range's "
-                        "difference type. TODO slicing from the end with 'end-2' syntax is not "
-                        "supported yet, sorry!");
+                    CONCEPT_ASSERT_MSG(meta::or_<BidirectionalRange<Rng>,
+                        meta::negate<meta::or_<is_end_type<T>, is_end_type<U>>>>(),
+                        "To slice from the end the object on which action::slice operates it must be a model"
+                        "of the BidirectionalRange concept.");
```
CaseyCarter: I'd just say "The range upon which action::slice operates must model BidrectionalRange to slice from the end." here. (Modulo how you choose to address the earlier comment about this requirement being insufficient, that is ;))

include/range/v3/begin_end.hpp
```diff
@@ -199,6 +199,13 @@ namespace ranges
                 (
                     Fn()(ref.get())
                 )
+
+                template<typename Int, CONCEPT_REQUIRES_(Integral<Int>())>
+                detail::from_end_<meta::_t<std::make_signed<Int>>> operator-(Int dist) const
+                {
+                  RANGES_EXPECT(0 <= static_cast<meta::_t<std::make_signed<Int>>>(dist));
```
CaseyCarter: Here's a another block of 2 space indents that should be 4.

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
+                  std::bind(slice, std::placeholders::_1, from, to)
+              )
+              template<typename D, CONCEPT_REQUIRES_(Integral<D>())>
+              static auto bind(slice_fn slice, detail::from_end_<D> from, detail::from_end_<D> to)
+              RANGES_DECLTYPE_AUTO_RETURN
+              (
+                  std::bind(slice, std::placeholders::_1, from, to)
+              )
+              template<typename D, CONCEPT_REQUIRES_(Integral<D>())>
+              static auto bind(slice_fn slice, D from, const end_fn &to)
```
CaseyCarter: range-v3 uses east `const`. (Occurs repeatedly.)

test/action/slice.cpp
```diff
@@ -23,22 +23,34 @@ int main()
 {
     using namespace ranges;
 
-    std::vector<int> v = view::ints(0,100);
```
CaseyCarter: I'd prefer that we leave the existing tests untouched - wrap them in `{}`s to scope them out of the way - and only add new ones.

Our jumbled together test style is inexcusable for `action` tests where every line modifies something and we have to go through the test line-by-line to audit anything. (Shhh, don't tell @ericniebler I'm trash-talking his tests.)

CaseyCarter:
> Seems like the valgrind package of homebrew is broken

I suggest you rebase your branch onto master - we just committed a Travis cleanup.

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
-                    typename I = iterator_t<Rng>,
-                    typename D = range_difference_type_t<Rng>,
-                    CONCEPT_REQUIRES_(Concept<Rng, D, D>())>
-                Rng operator()(Rng && rng, range_difference_type_t<Rng> from,
-                    range_difference_type_t<Rng> to) const
+              template<typename Rng,
+                  typename I = iterator_t<Rng>,
+                  typename D = range_difference_type_t<Rng>,
+                  CONCEPT_REQUIRES_(ForwardRange<Rng>() && ErasableRange<Rng, I, I>())>
+              Rng operator()(Rng &&rng, D from, D to) const
```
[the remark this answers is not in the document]
tete17: Im afraid I did not fully understood the last comment.
Could you please expand on what that wrapper would do

include/range/v3/action/slice.hpp
tete17: Im leaning towards keeping the `!SizedRange` since i do not want the extra overhead. But i can change it if you prefer it

include/range/v3/action/slice.hpp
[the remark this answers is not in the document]
CaseyCarter: Ideally, it would `RANGES_EXPECT(detail::can_represent<range_difference_type_t<Rng>>(x))` where `x` is `from` or `to` (unwrapped if they're detail::from_end_`) and `detail::can_represent` is a range-v3 port of `__can_represent` from cmcstl2 (https://github.com/CaseyCarter/cmcstl2/blob/master/include/stl2/detail/algorithm/is_permutation.hpp#L30), and then invoke the `slice_fn` call operator with the values converted to `range_difference_type_t<Rng>` (or `from_end<range_difference_type_t<Rng>>`).

include/range/v3/action/slice.hpp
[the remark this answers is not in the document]
tete17: Regarding the first comment i think we can do those 2 things simultaniously. I just need to overload the operator for `CommonRange<Rng>` and non common ranges.

That way we get flexibility and performance when possible
What do you think?

include/range/v3/action/slice.hpp
CaseyCarter: I have no strong preference either way. (I have some vague icky feelings about `slice` not always being O(1), but that's a pre-existing property and not specific to this change.)

test/action/slice.cpp
tete17: To be honest i was not happy either with the way i left them either.
Im happy by leaving the old test but im also in favour of fully rewrite them from scratch. If you point me to some good unit test somewhere in this library to pick the style i will get to it.

PS: Dont worry man. I got you. @ericniebler will never know

include/range/v3/action/slice.hpp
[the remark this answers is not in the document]
CaseyCarter: Overloading isn't necessary: the iterator+sentinel overload of `next` is O(1) when the sentinel type can be assigned to the iterator type, which is always the case for a `CommonRange` since they are the same type.

include/range/v3/action/slice.hpp
```diff
@@ -36,47 +36,100 @@ namespace ranges
...
-                            concepts::model_of<concepts::ForwardRange, Rng>(),
-                            concepts::model_of<concepts::ErasableRange, Rng, I, I>(),
-                            concepts::model_of<concepts::ConvertibleTo, T, D>(),
-                            concepts::model_of<concepts::ConvertibleTo, U, D>()
-                        ));
-                };
 
-                template<typename Rng, typename T, typename U>
-                using Concept = concepts::models<ConceptImpl, Rng, T, U>;
+              // Overloads for the pipe syntax: rng | view::slice(from,to)
```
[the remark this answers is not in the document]
tete17: Is it too late to include 
- [ ] add a `.clang-format` file

to #802 ?

test/action/slice.cpp
CaseyCarter: It doesn't need to be anything special; just wrap the individual test cases in block scopes or named functions so they're somewhat encapsulated.

tete17:
Should i Squash the commits before merging?

tete17:
Im not sure if you guys wante me to do any more changes
If so please let me know :smile:

ericniebler:
@CaseyCarter this looks ok to me. Has all your feedback been addressed? If I don't hear back, I'll just merge it.

@tete17 Thanks for your work on this. Sorry it has taken us so long to merge.

tete17:
@ericniebler It's fine I am happy to contribute in any form. 😄

CaseyCarter:
> @CaseyCarter this looks ok to me. Has all your feedback been addressed? If I don't hear back, I'll just merge it.

There was nothing egregiously broken in the first place, go ahead and merge. If there are nitpicky corner cases left we can fix them if and when problems appear.

Issue #877, opened 2018-08-17
RegularInvocable does not check that the callable isn't modified


h-2:
The source code suggests that RegularInvocable actually does not perform any checks beyond those of Invocable and while some things are hard to check, I think it would be easy to add code that checks whether a `const` qualified version of the callable would still be invocable – thus verifying that the Invocable does not change during invocation.

Am I right, or did I miss something? Would you accept a PR?

Here is an example of where compilation should fail on the assert, but only fails later on invocation:

```cpp
#include <range/v3/range_concepts.hpp>
#include <range/v3/utility/functional.hpp>

auto v = [count = 0] (char c) mutable { ++count; return c == '\n'; };


static_assert(ranges::RegularInvocable<decltype(v), char>());

int main()
{
    v('A');

    auto const & v2 = v;
    v2('C');
}
```

h-2:
Hm, I see this is all already according to the newly merged concepts into the IS. Is the motivation for not implementing RegularInvocable's constraints to allow for callables that *could* change, but *don't* ?

Why was this chosen? Because for non-object non-member functions you don't know if they keep state?

ericniebler:
> allow for callables that could change, but don't ?

Yes, and also because checking whether the `const`-qualified type is invocable doesn't actually mean it's regular. It could be referencing function-local statics or thread locals, or a host of other ways of subverting `const`-correctness. Basically, `const` doesn't mean a lot in C++, and using it to infer regularity of functions seems specious.

PR #878, opened 2018-08-17
implement part of RegularInvocable's axioms

h-2:
this would solve #877, at least the part concerning the immutability for the functor.

### DIFF
--- a/include/range/v3/utility/functional.hpp
+++ b/include/range/v3/utility/functional.hpp
@@ -217,6 +217,12 @@ namespace ranges
               : refines<Invocable>
             {
                 // Axiom: equality_preserving(invoke(f, args...))
+                template<typename Fun, typename... Args>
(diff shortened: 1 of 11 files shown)


h-2:
Ok, this actually triggered a bunch of test failures, because many predicates from the tests where not `const`-qualified although they could/should haven been. I have changed this, because I would argue that they can be const so they should be const, but I also realise this could be a breaking change for other people.

Four tests are still failing: `view::take_while`, `view::drop_while`, `view::replace_if` and `view::remove_if`, because the views check that the invocable they are given models `Predicate` and some tests pass a mutable lambda like in #877 which doesn't model `Predicate`, because it intentionally mutates state.

This clearly violates the axioms (even if not enforced). I would argue that the views should only require `Invocable` (with additional constraints for returning `Boolean`) and not `Predicate`, but before I change that: what's your opinion on the whole matter?

ericniebler:
As I mentioned in #877, I think this is ill-advised. Closing.

PR #881, opened 2018-08-29
any_view category inheritance should affect only the iterator categor…

### LINKED ISSUE #880: Can not assign bidirectional to forward any_view
Consider the following:

```
int main()
{
    map<string, string> mm{ make_pair("hi", "there"), make_pair("good", "bye") };
    auto just_keys = mm | view::keys;
    v3::any_view<string, v3::category::forward | v3::category::sized> as_any = just_keys;
    
    cout << "The keys: " << as_any;
}
```

It fails with:

```
/opt/wandbox/range-v3/include/range/v3/view/any_view.hpp:325:56: error: non-virtual member function marked 'override' hides virtual member function
                bool equal(Forward const &that_) const override
```

But works fine if my `any_view` is declared with concept bidirectional. I know that the view returned by `keys` is bidirectional but I think I should be allowed to assign it a any_view with strictly fewer requirements.

Live demo: https://wandbox.org/permlink/I3Cflj0MCkOi5lOS

I'm fairly new to ranges so perhaps I'm just thinking about this wrong. My use case is that I have an interface that returns an `any_view` that's `sized` and `forward`. I have a subclass that overrides that method and it seems to me that there should be some way to convert my bidirectional range into a forward one so I can correctly override the base class method.

CaseyCarter:
…y portion of category

(Yes, we have a problem with name overloading.)

Fixes #880.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -219,7 +219,8 @@ matrix:
 before_install:
   - export CHECKOUT_PATH=`pwd`;
   - |
-    if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    set
+    if [ "$TRAVIS_OS_NAME" == "osx" -o "$TRAVIS_OS_NAME" == "max os x" ]; then
       brew update
       brew install valgrind
       brew install gnu-sed --with-default-names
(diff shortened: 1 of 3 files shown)


CaseyCarter:
Closing this while I do battle with Travis.

PR #883, opened 2018-08-31
Travis rework

CaseyCarter:
* Don't use `sudo`, which means installing the built libc++ into `$TRAVIS_BUILD_DIR/llvm` instead of `/usr`

* `brew` is griping about `valgrind`, disable it on OSX for now

* clang modules build is on the floor with an insane error about initializing the same `static constexpr value` member of a type trait twice on the same line in `<type_traits>`

* I'm ripping out the unused Boost machinery.


CaseyCarter:
@gnzlbg @EricWF Attention clang modules folk: if you want range-v3 CI coverage for clang modules, you need to fix the build. I have no idea what to do about:
```
In module 'std' imported from /home/travis/build/CaseyCarter/range-v3/test/distance.cpp:13:
/home/travis/build/CaseyCarter/range-v3/llvm/include/c++/v1/type_traits:506:53: error: static data member 'value' already has an initializer
    static _LIBCPP_CONSTEXPR const _Tp      value = __v;
                                                    ^
[...]

/home/travis/build/CaseyCarter/range-v3/llvm/include/c++/v1/type_traits:506:53: note: previous initialization is here
    static _LIBCPP_CONSTEXPR const _Tp      value = __v;
                                                    ^
```

gnzlbg:
@CaseyCarter I've seen that before, clang modules is pretty much broken with respect to static variables, static data members, constexpr variables, etc. I've reported many bugs, but at some point I gave up and stopping using it.

gnzlbg:
That's why it sounded familiar. I reported that bug ages ago (https://bugs.llvm.org/show_bug.cgi?id=33746) and it should have been fixed, but it might well have reappeared. 

For all I can tell, I was the only person actually trying to use modules in production, since I was the only one reporting bugs.

CaseyCarter:
> it should have been fixed

Richard's comment on the bug with the fix revision is from 201711, which suggests to me that it's likely fixed in clang 6 but not in 5 which we test. I don't know why it was working before and not working now, but it's possible we can reenable the module build CI for clang 6 when I get around to updating the CI matrix with clang 6 and gcc 7/8.

PR #887, opened 2018-09-13
reference_t is an associated type of dereferenceable, not of Readable

CaseyCarter:
So let's relocate its definition to `associated_types.hpp` and use it directly in the definition of `iter_move`.


CaseyCarter:
FWIW, I also have ulterior motives: MSVC thinks `decltype(*std::declval<I>())` is an rvalue reference type when `I` is a pointer-to-array, and this change lets me workaround one occurrence instead of several.

ericniebler:
I wouldn't sweat the MSVC support until I land the new-concepts-syntax branch since it changes everything.

PR #891, opened 2018-09-19
Nasty hack to unblock fix-reverse-iterator

CaseyCarter:
...until I can localize the interaction between range-v3 and MSFTL's `_Vector_iterator`.

### DIFF
--- a/include/range/v3/algorithm/inplace_merge.hpp
+++ b/include/range/v3/algorithm/inplace_merge.hpp
@@ -63,9 +63,9 @@ namespace ranges
                     if(len1 <= len2)
                     {
                         auto p = ranges::move(begin, middle, tmpbuf.begin()).second;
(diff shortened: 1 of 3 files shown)


include/range/v3/algorithm/inplace_merge.hpp
```diff
@@ -63,9 +63,9 @@ namespace ranges
                     if(len1 <= len2)
                     {
                         auto p = ranges::move(begin, middle, tmpbuf.begin()).second;
-                        merge(make_move_iterator(buf), make_move_sentinel(p.base().base()),
+                        merge(make_move_iterator(buf), make_move_iterator(p.base().base()),
```
CaseyCarter: Unrelated drive-by changes: these arguments to `make_move_sentinel` are statically known to have the same type as the arguments passed to the corresponding `make_move_iterator`.

PR #893, opened 2018-09-19
avoid ambiguity created between std::advance and boost::advance; fixes #845

### LINKED ISSUE #845: call to 'advance' is ambiguous

ericniebler:


test/utility/iterator.cpp
```diff
@@ -157,6 +158,19 @@ CONCEPT_ASSERT(IndirectlyMovable<int const *, int *>());
...
+{
+    struct my_dummy{}; // just to have a type from boost namespace
+}
+
+// Regression test for https://github.com/ericniebler/range-v3/issues/845
+void test_845()
+{
+    std::list<boost::my_dummy> v = { boost::my_dummy{} };
+    auto itr = v.begin();
+    ranges::advance(itr, 1); // Should not create ambiguity
```
CaseyCarter: This would be a much more useful test if we use the namespace `notboost`, and put a conflicting definition of `advance` in it.

include/range/v3/utility/iterator.hpp
```diff
@@ -109,7 +87,7 @@ namespace ranges
                 {
                     // Use ADL here to give custom iterator types (like counted_iterator)
                     // a chance to optimize it (see utility/counted_iterator.hpp)
```
CaseyCarter: This comment has bitrotted - we should remove it.

PR #897, opened 2018-10-02
basic_iterator: inherit mixin via traits base, to avoid multiple path…

CaseyCarter:
…s for value_type


CaseyCarter:
I won't cry when we can rip out the EBO code and use `[[no_unique_address]]` everywhere.

PR #908, opened 2018-10-11
New concepts syntax

CaseyCarter:
Porting changes from master.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -57,7 +57,6 @@
         {
             "name": "Win32",
             "includePath": [
-                "C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/include",
                 "${workspaceRoot}/include"
             ],
             "defines": [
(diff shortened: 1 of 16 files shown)


ericniebler:
Looks like this broke Appveyor. 😑

CaseyCarter:
I know why it's broken: name resolution for `get(x)` fails when `x` is `tagged<std::tuple<Ts...>, /*...*/>` because the conversion to `std::tuple<Us...>` necessary for template argument deduction against `std::get(std::tuple<Us...>)` to succeed is ambiguous. What I don't know is why it didn't show up before.

CaseyCarter:
Unrelated: What's with the `constexpr /*c++14*/`s? Doesn't this branch require C++14?

CaseyCarter:
> the conversion to `std::tuple<Us...>` necessary for template argument deduction against `std::get(std::tuple<Us...>)` to succeed is ambiguous. What I don't know is why it didn't show up before.

Very much related: I despise MSVC's `tuple`.

ericniebler:
> Doesn't this branch require C++14?

Yes.

PR #909, opened 2018-10-11
make the generate and generate_n views work with move-only types

### LINKED ISSUE #905: generate seems broken with move-only return types
I haven't been able to get anything that uses this view to compile:
```
auto rng = view::generate_n([]()->MoveOnlyString{ return {"Hello, world!"}; }, 4);
```
[godbolt](https://godbolt.org/z/xf_ar5)

The issue seems to revolve around the cached value stored by `view::generate_n` (and `view::generate`).

To me this seems like an important use case, but hopefully I'm just misunderstanding something!

ericniebler:
fixes #905

NOTE WELL: This changes generators so that they return xvalue references to the cached value. This is a silent breaking change for some valid code. It is, however, arguably closest to the programmer's intent as I imagine it.

@CaseyCarter, please offer your perspectives here.

### DIFF
--- a/include/range/v3/detail/variant.hpp
+++ b/include/range/v3/detail/variant.hpp
@@ -391,14 +391,14 @@ namespace ranges
                 }
                 template<typename U>
                 meta::if_<std::is_object<U>>
-                    operator()(indexed_datum<U, meta::size_t<N>> &u)
+                operator()(indexed_datum<U, meta::size_t<N>> &u)
                     noexcept(std::is_nothrow_constructible<U, Ts...>::value)
                 {
                     this->construct_(u.get(), meta::make_index_sequence<sizeof...(Ts)>{});
                 }
                 template<typename U>
                 meta::if_<meta::not_<std::is_object<U>>>
-                    operator()(indexed_datum<U, meta::size_t<N>> &u)
+                operator()(indexed_datum<U, meta::size_t<N>> &u)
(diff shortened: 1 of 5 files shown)


CaseyCarter:
This is icky, in that it makes it impossible to achieve the prior behavior for functions that return prvalues of `Copyable` types, and makes it impossible to create a `generate_view` whose reference type is a prvalue in general. I'm tempted to suggest only xvalue-ifying when the result type of the function is a non-`Copyable` prvalue to keep simple things simple, but conditional semantics are icky as well.

ericniebler:
> This is icky, in that it makes it impossible to achieve the prior behavior for functions that return prvalues of Copyable types

```c++
generate([]{return T{};}) | transform([](auto const&t){return t;});
```

> makes it impossible to create a generate_view whose reference type is a prvalue in general

Yes, but non-regular functions are weird. Something has to give. &lt;shrug&gt;

> I'm tempted to suggest only xvalue-ifying when the result type of the function is a non-Copyable prvalue to keep simple things simple, but conditional semantics are icky as well.

I had considered that but considered it better to avoid the conditional semantics.

ericniebler:
... looks like I exposed a bug in `variant`. Whee!

ericniebler:
> leaves currently well-formed programs unchanged

I'm more interested in maximally useful behavior, not preserving strict back-compat. Now is the time for breaking changes.

I am still weakly in favor of the original change as implemented in this PR, as admittedly yucky as it is.

PR #910, opened 2018-10-12
stride_view<R> can be Common only when R is Common and either Sized o…

CaseyCarter:
…r not Bidirectional

We cannot calculate the offset for the end iterator in O(1) when `R` is `Bidirectional` and not `Sized`.


ericniebler:
Makes sense, thanks.

PR #928, opened 2018-10-28
Oooh, look. MSVC things.

CaseyCarter:
Various notes in no particular order:
* This is currently based on #923, so don't be surprised to see those changes in this PR. Click the "Support building with MSVC" commit to see only the changes unrelated to #923.

* Don't bother trying to build this. The target compiler is 15.9, but some of the necessary bug fixes haven't yet flowed into preview releases.

* I've made no effort to integrate bug workarounds. They all look like:
  ```c++
  #ifdef NASTY_MSVC_BUG_WORKAROUND
  // Disgusting hacks 
  #else // ^^^ workaround / no workaround vvv
  // Pristine and perfect range-v3 code here.
  #endif // NASTY_MSVC_BUG_WORKAROUND
  ```
  I've even done this with the workarounds for Clang #37556. I'm fond of this "name-and-shame" approach to bug workarounds which makes it easy to know which garbage we can excise from the codebase when we drop support for what compiler version.

* I've tagged individual workarounds with bug numbers from the MSVC internal tracker. These numbers obviously aren't terribly useful except as opaque IDs and/or to communicate to people on the Visual C++ Team. (FWIW, I'll be checking this into the compiler test suite post-merge to replace the three-year-old range-v3 test suite that the compiler front-end team runs in their inner loop. Range-v3 will have some mighty good protection against regressions on MSVC.)

* I'm not adding CI support for MSVC yet. Firstly because AppVeyor doesn't have the unreleased compiler versions, and second because I suspect that other maintainers aren't interested in working around MSVC template metaprogramming bugs. I suppose we should say MSVC is "provisionally" supported until we have CI gating set up. I'm hoping to keep the Visual C++ team focused on stomping range-v3 bugs which are going to become everyone's bugs in C++20: the effort invested by everyone in VC++ over the last several months to get us here has been incredible.

### DIFF
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -398,7 +398,7 @@ namespace ranges
         struct basic_mixin;
 
         template<typename Cur>
-        struct basic_iterator;
+        struct RANGES_EMPTY_BASES basic_iterator;
 
         template<cardinality>
         struct basic_view : view_base
@@ -440,7 +440,7 @@ namespace ranges
     #endif
 
         template<typename I, typename S = I>
-        struct iterator_range;
+        struct RANGES_EMPTY_BASES iterator_range;
 
         template<typename I, typename S = I>
         struct sized_iterator_range;
@@ -454,7 +454,7 @@ namespace ranges
         // Views
         //
         template<typename Rng, typename Pred>
-        struct adjacent_filter_view;
+        struct RANGES_EMPTY_BASES adjacent_filter_view;
 
         namespace view
         {
@@ -462,7 +462,7 @@ namespace ranges
         }
 
         template<typename Rng, typename Pred>
-        struct adjacent_remove_if_view;
+        struct RANGES_EMPTY_BASES adjacent_remove_if_view;
 
         namespace view
         {
@@ -500,7 +500,7 @@ namespace ranges
             basic_iterator<detail::move_into_cursor<I>>;
 
         template<typename Rng, bool = (bool) is_infinite<Rng>()>
-        struct cycled_view;
+        struct RANGES_EMPTY_BASES cycled_view;
 
         namespace view
         {
@@ -610,7 +610,7 @@ namespace ranges
         }
 
         template<typename Rng, bool = decltype(detail::double_reverse<Rng>(42))::value>
-        struct reverse_view;
+        struct RANGES_EMPTY_BASES reverse_view;
 
         namespace view
         {
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -475,14 +475,6 @@ namespace ranges
         }
 
         template<typename Rng>
-        struct bounded_view;
-
-        namespace view
-        {
-            struct bounded_fn;
-        }
-
-        template<typename Rng>
         struct const_view;
 
         namespace view
@@ -507,7 +499,7 @@ namespace ranges
         using move_into_iterator =
             basic_iterator<detail::move_into_cursor<I>>;
 
-        template<typename Rng>
+        template<typename Rng, bool = (bool) is_infinite<Rng>()>
         struct cycled_view;
 
         namespace view
@@ -609,7 +601,15 @@ namespace ranges
             struct repeat_fn;
         }
 
-        template<typename Rng>
+        namespace detail
+        {
+            template<typename>
+            std::false_type double_reverse(long);
+            template<typename R, meta::if_<std::is_same<R, typename R::unreverse_me>>* = nullptr>
+            std::true_type double_reverse(int);
+        }
+
+        template<typename Rng, bool = decltype(detail::double_reverse<Rng>(42))::value>
         struct reverse_view;
 
         namespace view
--- a/include/range/v3/utility/any.hpp
+++ b/include/range/v3/utility/any.hpp
@@ -61,7 +61,14 @@ namespace ranges
         template<typename T>
         T const * any_cast(any const *) noexcept;
 
+#if defined(RANGES_WORKAROUND_MSVC_589046) && !defined(RANGES_DOXYGEN_INVOKED)
+        namespace _any_ { struct adl_hook {}; }
+#endif
+
         struct any
+#if defined(RANGES_WORKAROUND_MSVC_589046) && !defined(RANGES_DOXYGEN_INVOKED)
+          : private _any_::adl_hook
+#endif
         {
         private:
             template<typename T>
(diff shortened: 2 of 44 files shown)


ericniebler:
Very cool. Have you tried the new-concepts-syntax branch, by any chance?

CaseyCarter:
> Have you tried the new-concepts-syntax branch, by any chance?

Only enough to determine that the merge needs lots of handwork due to code motion. It's on my plate for post-San Diego.

include/range/v3/range_fwd.hpp
```diff
@@ -609,8 +601,16 @@ namespace ranges
...
+        namespace detail
+        {
+            template<typename>
+            std::false_type double_reverse(long);
+            template<typename R, meta::if_<std::is_same<R, typename R::unreverse_me>>* = nullptr>
+            std::true_type double_reverse(int);
+        }
+
+        template<typename Rng, bool = decltype(detail::double_reverse<Rng>(42))::value>
+        struct RANGES_EMPTY_BASES reverse_view;
```
ericniebler: Boo. 👎

include/range/v3/utility/any.hpp
```diff
@@ -61,7 +61,14 @@ namespace ranges
...
         T const * any_cast(any const *) noexcept;
 
+#if defined(RANGES_WORKAROUND_MSVC_589046) && !defined(RANGES_DOXYGEN_INVOKED)
+        namespace _any_ { struct adl_hook {}; }
+#endif
+
         struct any
+#if defined(RANGES_WORKAROUND_MSVC_589046) && !defined(RANGES_DOXYGEN_INVOKED)
+          : private _any_::adl_hook
+#endif
```
ericniebler: empty base?

include/range/v3/view/stride.hpp
```diff
@@ -135,24 +135,35 @@ namespace ranges
             // stride_view const models Range if Rng const models Range, and
             // either (1) Rng is sized, so we can pre-calculate offset_, or (2)
             // Rng is not Bidirectional, so it does not need offset_.
+#ifdef RANGES_WORKAROUND_MSVC_711347
+            static constexpr bool const_iterable = Range<Rng const>() &&
+                (SizedRange<Rng const>() || !BidirectionalRange<Rng const>());
+#else // ^^^ workaround / no workaround vvv
```
ericniebler: Really? MSVC can't handle a constexpr static member function?

include/range/v3/view/stride.hpp
```diff
@@ -244,20 +255,34 @@ namespace ranges
             {
                 return adaptor<false>{*this};
             }
+#ifdef RANGES_WORKAROUND_MSVC_711347
+            CONCEPT_REQUIRES(const_iterable)
+#else // ^^^ workaround / no workaround vvv
             CONCEPT_REQUIRES(const_iterable())
+#endif // RANGES_WORKAROUND_MSVC_711347
```
ericniebler: These edits are unfortunate.

include/range/v3/utility/any.hpp
CaseyCarter: *One* empty base class is fine, only classes with *multiple* potentially empty base classes need the `RANGES_EMPTY_BASES` annotation.

include/range/v3/view/stride.hpp
CaseyCarter: Since we have two parsers, everything in MSVC is context-sensitive. The `constexpr` static member functions explode when used in the constraints for `begin` and `end`, yet the variable templates are fine. I don't even try to understand anymore, I just file the bugs.

include/range/v3/range_fwd.hpp
CaseyCarter: Which part are you "boo"ing here?

include/range/v3/view/stride.hpp
CaseyCarter: Is there a different style you'd prefer? I didn't give this much thought under the expectation that I'll be able to use the variable-template workaround for the "normal" case on the new-concepts-syntax branch.

include/range/v3/view_adaptor.hpp
```diff
@@ -80,7 +80,11 @@ namespace ranges
               : compressed_pair<BaseIter, Adapt>
             {
                 using compressed_pair<BaseIter, Adapt>::compressed_pair;
+#ifdef RANGES_WORKAROUND_MSVC_688606
+                using value_type = value_type_t<Adapt>;
```
CaseyCarter: This one is my favorite: since `Adapt` is an indirect base, MSVC effectively rewrites line 86 to `using value_type = typename compressed_pair<BaseIter, Adapt>::ebo_box<Adapt>::Adapt::value_type;`, and then complains that this class can't access that member because `ebo_box` is a private base.

include/range/v3/range_fwd.hpp
ericniebler: I don't think it makes sense for reversing a reverse view to be special, and if it were, I would expect a partial specialization of `reverse_view<reverse_view<...>>` to be sufficient. I don't like `unreverse_me`, and I don't like the extra `bool` template parameter.

CaseyCarter: >  I would expect a partial specialization of reverse_view<reverse_view<...>> to be sufficient.

This was part of my distaste for [Option (3) from the `cycled_view` PR](https://github.com/ericniebler/range-v3/pull/923#discussion_r226835878). "Just a partial specialization" correctly avoids reversing a `reverse_view`, so `ranges::equal(x, view::reverse(view::reverse(x)))` as desired, but then additional applications of `reverse` also do nothing which is *not* desired: `ranges::equal(x, view::reverse(view::reverse(view::reverse(x))))` holds.

CaseyCarter: The "sometimes `foo_view` does nothing" design seems more distasteful than the "`foo_view` and `view::foo` have different behaviors" design.

ericniebler: Wouldn't `reverse_view<reverse_view<R>>` just return `R`'s iterator/sentinel? If `R` happens to be a `reverse_view`, didn't the right thing just happen?

include/range/v3/view/stride.hpp
ericniebler: I was merely complaining about a large number of workarounds, the only purpose of which was the addition or removal of `()`. You're right, this should go away in the new-concepts-syntax branch.

include/range/v3/view_adaptor.hpp
ericniebler: 🤢

include/range/v3/range_fwd.hpp
CaseyCarter: *sigh* I'm going to take a month of vacation after Ranges gets merged - I'm cooked.

ericniebler: 🤣

CaseyCarter: Fixed in #931.

Issue #868, opened 2018-07-21
Feature request: ranges::ostream_joiner


denchat:
Now gcc has already implemented [std::experimental::ostream_joiner](https://en.cppreference.com/w/cpp/experimental/lib_extensions_2#ostream_joiner). It would be nice if range-v3 has it also.

johelegp:
This has been solved.

PR #930, opened 2018-10-30
Implemented ostream_joiner (#868)

sv1990:
I'm not sure if I should have constrained `operator=` in some way but in general this should do it.

### DIFF
--- a/include/range/v3/utility/iterator.hpp
+++ b/include/range/v3/utility/iterator.hpp
@@ -627,14 +627,13 @@ namespace ranges
             using char_type       = Char;
             using traits_type     = Traits;
             using ostream_type    = std::basic_ostream<Char, Traits>;
-            using delim_type      = detail::decay_t<Delim>;
 
             constexpr ostream_joiner() = default;
-            ostream_joiner(ostream_type &s, delim_type const &d) noexcept
-              : delim_(d), sout_(&s), first_(true)
+            ostream_joiner(ostream_type &s, Delim const &d) noexcept
+              : delim_(d), sout_(std::addressof(s)), first_(true)
             {}
-            ostream_joiner(ostream_type &s, delim_type &&d) noexcept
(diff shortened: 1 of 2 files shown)


include/range/v3/utility/iterator.hpp
```diff
@@ -619,6 +619,64 @@ namespace ranges
...
+        struct ostream_joiner
+        {
+            using difference_type = std::ptrdiff_t;
+            using char_type       = Char;
+            using traits_type     = Traits;
+            using ostream_type    = std::basic_ostream<Char, Traits>;
+
+            constexpr ostream_joiner() = default;
+            ostream_joiner(ostream_type &s, Delim const &d) noexcept
+              : delim_(d), sout_(std::addressof(s)), first_(true)
```
ericniebler: The `noexcept` is curious here, since `Delim` is not constrained to be a nothrow-copyable type. Is that really what is being standardized? My preference is to nuke the `noexcept` on the two constructors and on `make_ostream_joiner_fn::operator()`.

sv1990: You are right. I removed it

sv1990: I also had forgotten the noexcept on `operator++` and `operator*`...

include/range/v3/utility/iterator.hpp
```diff
@@ -619,6 +619,64 @@ namespace ranges
...
+            }
+          private:
+            Delim delim_;
+            ostream_type *sout_;
+            bool first_;
+        };
+
+        struct make_ostream_joiner_fn
+        {
+            template <typename Delim, typename Char, typename Traits>
```
CaseyCarter: `Delim` needs to model `SemiRegular` in order for the iterator to do so.; we should constrain this:
```c++
template <typename Delim, typename Char, typename Traits,
    CONCEPT_REQUIRES_(SemiRegular<Delim>())>
```

PR #931, opened 2018-10-30
Silence MSVC warnings, fix reverse_view

CaseyCarter:
The bogus truncation warnings are all occurrences of:
```c++
constexpr unsigned n = size(some_array);
```
in test code. MSVC considers conversions to be narrowing based on their types even when the source is a constant expression, so it warns for these (Yes, this has been fixed, but it doesn't meet the incredibly high bug bar for 15.9). I replaced `unsigned` with `auto` to silence the warnings, shout if you prefer `std::size_t` or "silence the stupid warnings in the tests" instead.


CaseyCarter:
...are you *sure* we don't want to drop support for GCC4.9?

ericniebler:
We can drop gcc-4.9 support. It's time.

CaseyCarter:
> We can drop gcc-4.9 support. It's time.

For now, I'm just disabling the `reverse_view` optimization on gcc-4.9 that's triggering the test failure. We can followup with a README change to officially "unsupport" 4.9 for the range-v3 0.5 release.

include/range/v3/view/reverse.hpp
```diff
@@ -107,22 +107,21 @@ namespace ranges
...
-            using unreverse_me = reverse_view;
         };
 
-        template<typename Rng_>
-        struct reverse_view<Rng_, true>
-          : identity_adaptor<decltype(std::declval<Rng_ &>().base())>
+#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 5
+        template<typename Rng>
+        struct reverse_view<reverse_view<Rng>>
+          : identity_adaptor<Rng>
```
ericniebler: This partial specialization is buggy. For every `reverse_view<T>`, `.base()` should return a `T`. So `reverse_view<reverse_view<Rng>>::base()` should return a `reverse_view<Rng>`. But this specialization, which gets its `base()` from `identity_adaptor<Rng>` _doesn't_ return a `reverse_view<Rng>`; it returns a `Rng`. This specialization of `reverse_view` needs to define its own `base()` that reconstitutes a `reverse_view<Rng>` from the `Rng` object -- from which it should inherit directly.

So two things:
1. Implementing `reverse_view<reverse_view<Rng>>` in terms of `identity_adaptor<Rng>` is wrong.
2. `identity_adaptor<Rng>::base()` is highly suspect and should be removed.

CaseyCarter: Yes, I've discovered how fragile this is; I'm planning to revert to the "weird customizations happen in `view::foo` design."

PR #933, opened 2018-11-09
[MSVC] Support for solution folders in CMake.

mkurdej:
This PR is just a nice-to-have feature for people that use MSVC.
It adds support for solution folders (grouping together different targets).

It works only partially with CMake < 3.12, because of the use of `CMAKE_FOLDER` variable which is not recognized these earlier versions (https://cmake.org/cmake/help/v3.12/variable/CMAKE_FOLDER.html)
Supporting those seems to me as an unnecessary maintenance burden (each target would need to set its `FOLDER` property instead of setting it once by filesystem folder/cmakelists.txt).


CaseyCarter:
This is amazing! It makes the "Cmake Targets View" in Visual Studio usable. Thanks.

Issue #859, opened 2018-07-12
Feature request - Add predicate function to view::unique


tuxfede:
Hi,

Could view::unique has a predicate function as parameter? (like action::unique or ranges::unique)

Thanks in advance!

CaseyCarter:
Yes, this is definitely possible. The implementation uses a predicate internally, it would simply be a matter of exposing that in the interface.

johelegp:
This can be closed.

PR #936, opened 2018-11-14
#859 Expose predicate view unique

tete17:
This Pull Request solves issue #859 

I have choosed to stay with the same interface than `action::unique` that meand I have to use `not_fn()` 

It is very self explanatory but let me know if there is something you guys think should be improved.


include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
                 using Concept = meta::and_<
                     ForwardRange<Rng>,
-                    EqualityComparable<range_value_type_t<Rng>>>;
+                    IndirectRelation<C, I>>;
 
-                template<typename Rng, CONCEPT_REQUIRES_(Concept<Rng>())>
-                unique_view<all_t<Rng>> operator()(Rng && rng) const
+                template<typename Rng, typename C = equal_to,
+                        CONCEPT_REQUIRES_(Concept<Rng, C>())>
+                auto operator()(Rng && rng, C pred = equal_to{}) const ->
```
CaseyCarter:
```suggestion
                auto operator()(Rng && rng, C pred = {}) const ->
```

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
+                    std::bind(unique, std::placeholders::_1, protect(std::move(pred)))
+                )
+
+            public:
+                template<typename Rng, typename C = equal_to,
+                        typename I = iterator_t<Rng>>
                 using Concept = meta::and_<
                     ForwardRange<Rng>,
-                    EqualityComparable<range_value_type_t<Rng>>>;
+                    IndirectRelation<C, I>>;
```
CaseyCarter: I suggest using `iterator_t<Rng>` directly here:

```suggestion
                    IndirectRelation<C, iterator_t<Rng>>>;
```
 and avoiding the extra template parameter to save a few microseconds of compile time.

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
         {
             struct unique_fn
             {
-                template<typename Rng>
+            private:
+                friend view_access;
+                template<typename C,
+                        CONCEPT_REQUIRES_(!Range<C>())>
```
CaseyCarter: We're admittedly inconsistent about whether template head continuation lines get a single additional indent (as on line 62) or are aligned with the parameters on the first line, but ideally you should pick one of those two styles and either add a space to this line or remove 3. (Also on line 48 and 54.)

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
 
-                template<typename Rng, CONCEPT_REQUIRES_(Concept<Rng>())>
-                unique_view<all_t<Rng>> operator()(Rng && rng) const
+                template<typename Rng, typename C = equal_to,
+                        CONCEPT_REQUIRES_(Concept<Rng, C>())>
+                auto operator()(Rng && rng, C pred = equal_to{}) const ->
+                adjacent_filter_view<all_t<Rng>, decltype(not_fn(pred))>
                 {
-                    return {all(static_cast<Rng&&>(rng)), not_equal_to{}};
+                    return {all(static_cast<Rng&&>(rng)), not_fn(pred)};
```
CaseyCarter: Pre-existing (probably my fault from before I gave up on leading @ericniebler down the one-true-path of "references and pointers hug left"):
```suggestion
                    return {all(static_cast<Rng &&>(rng)), not_fn(pred)};
```

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
+                template<typename Rng, typename C,
+                    CONCEPT_REQUIRES_(!Concept<Rng, C>())>
                 void operator()(Rng &&) const
                 {
                     CONCEPT_ASSERT_MSG(ForwardRange<Rng>(),
                         "The object on which view::unique operates must be a model the "
                         "ForwardRange concept.");
-                    CONCEPT_ASSERT_MSG(EqualityComparable<range_value_type_t<Rng>>(),
+                    using I = iterator_t<Rng>;
+                    CONCEPT_ASSERT_MSG(IndirectRelation<C, I>(),
```
CaseyCarter: Unlike on line 48, compile time isn't an issue here - but it would save a line to use `iterator_t<Rng>` directly here.

include/range/v3/view/unique.hpp
CaseyCarter: Should this be:
```suggestion
                    IndirectRelation<C&, I>>;
```
I literally don't recall if `IndirectRelation` is value category sensitive.

test/view/unique.cpp
```diff
@@ -95,7 +95,47 @@ int main()
...
+        models_not<concepts::SizedRange>(rng);
+        ::check_equal(rng, {4, 3, 2, 1});
+    }
+
+    {
+        auto const caseInsensitiveCompare = [](const std::string& s1, const std::string& s2){
+            if (s1.size() != s2.size())
+                return false;
+            for (unsigned i = 0; i < s1.size(); i++)
+                if (toupper(s1.at(i)) != toupper(s2.at(i)))
```
CaseyCarter: Ahhh, `at`! It burns my eyes!

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
+            public:
+                template<typename Rng, typename C = equal_to,
+                        typename I = iterator_t<Rng>>
                 using Concept = meta::and_<
                     ForwardRange<Rng>,
-                    EqualityComparable<range_value_type_t<Rng>>>;
+                    IndirectRelation<C, I>>;
 
-                template<typename Rng, CONCEPT_REQUIRES_(Concept<Rng>())>
-                unique_view<all_t<Rng>> operator()(Rng && rng) const
```
CaseyCarter: I *think* these are the only uses of `unique_view`. If so, we can remove the definition of the alias in `range_fwd.hpp`.

include/range/v3/view/unique.hpp
```diff
@@ -33,27 +33,44 @@ namespace ranges
...
                 {
-                    return {all(static_cast<Rng&&>(rng)), not_equal_to{}};
+                    return {all(static_cast<Rng&&>(rng)), not_fn(pred)};
                 }
             #ifndef RANGES_DOXYGEN_INVOKED
-                template<typename Rng,
-                    CONCEPT_REQUIRES_(!Concept<Rng>())>
+                template<typename Rng, typename C,
+                    CONCEPT_REQUIRES_(!Concept<Rng, C>())>
                 void operator()(Rng &&) const
```
CaseyCarter: The template parameter `C` is non-deduceable, so this function template won't be called for invocations that don't match `Concept` as intended. You should default `C` to `equal_to` and add a `C = {}` function parameter to mirror the interface of the "real" call operator.

include/range/v3/view/unique.hpp
tete17: Every where that I saw on the code it appear as just `C`
See `algorithm/find.hpp:45`

include/range/v3/view/unique.hpp
tete17: hahaha never give up @CaseyCarter 
You have to gice more details about why that is the **one true-path**

test/view/unique.cpp
tete17: Upss!!!
I didn't know that was not allowed here.
I guess it is because it can throw??

CaseyCarter: No, it's mostly because I despise `at` due to the horrific effects it has on optimizers.

include/range/v3/view/unique.hpp
ericniebler: @CaseyCarter is simply wrong. `int i = 0, *pi = &i;` clearly demonstrates that pointers and references should hug right.

include/range/v3/range_fwd.hpp
```diff
@@ -721,9 +721,6 @@ namespace ranges
             struct unbounded_fn;
         }
 
-        template<typename Rng>
-        using unique_view = adjacent_filter_view<Rng, not_equal_to>;
-
```
ericniebler: I'd rather we kept this around.

```c++
template<typename Rng>
using unique_view = adjacent_filter_view<Rng, logical_negate<equal_to>>;
```

tete17:
That should do it :smile:

PR #941, opened 2018-11-21
Add "enumerate" view

MikeGitb:


MikeGitb:
The error on travis seems to be unrelated to the PR but is due to a problem with installing libc++. Maybe you could re-trigger the failed build?

MikeGitb:
Done. 

Are the Unit Tests sufficient? I also didn't add anything to the documentation, as I'm not yet familiar with how that is done here.

test/view/enumerate.cpp
```diff
@@ -0,0 +1,84 @@
+// Range v3 library
+//
+//  Copyright Casey Carter 2018-present
```
CaseyCarter: Why is this copyrighted to me instead of you? I didn't write this test code ;)

CaseyCarter:
> Are the Unit Tests sufficient?

Yes: it's a composition of other view types that already have tests, so it doesn't need to be anything fancy.

> I also didn't add anything to the documentation, as I'm not yet familiar with how that is done here.

Sadly, that mostly *is* how that is done here: we add things and don't bother to document them. The docs are auto-generated with doxygen, so you could `\addtogroup group-views` and throw some doc comments on if you're motivated.

test/view/enumerate.cpp
MikeGitb: Right. Copy paste error - my mistake.
Is it ok, if I just put in my github username?

CaseyCarter: Sure.

MikeGitb:
OK, I fixed the copyright notice, added a documentation comment and rebased to current master. From my side this PR is redy to go once the CI checks pass.

PR #948, opened 2018-11-29
Implement detail::is_convertible alias template

### LINKED ISSUE #946: MSVC don't like forwarded classes
The following code does not compiles with MSVC 15.9.3
```cpp
#include <iostream>
#include <memory>
#include <vector>

#include <range/v3/view/indirect.hpp>


class Data;

struct Test {
    std::vector<Data*> m_list;

    auto list() {
        return m_list | ranges::view::indirect;
    }
};


class Data {};

int main()
{
    std::cout << Test().list().size();
}
```

With C2139 error, saying that undefined Data class can not be used with  "__is_convertible_to".
LLVM toolchain do compiles this code.

CaseyCarter:
Fixes #946.

### DIFF
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -345,6 +345,14 @@ namespace ranges
 
             template<typename T>
             using remove_rvalue_reference_t = meta::_t<remove_rvalue_reference<T>>;
+
+            // Workaround bug in the Standard Library:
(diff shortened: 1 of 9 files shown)


CaseyCarter:
@CaseyCarter: update the vcpkg reference after merging this. You've already validated it against 15.9 x86/x64 debug/release.

Issue #952, opened 2018-12-02
view::indirect does not work with view::transform


tower120:
Following doesn't compiles:

```cpp
    std::vector<int> list;
    auto a = list | ranges::view::transform([](int& i)->int* {return &i; })
                  | ranges::view::indirect;
    auto b = a.begin();
```

Compiler errors are not very helpful...
https://wandbox.org/permlink/h71MoQWuMEpyfM8V

PR #954, opened 2018-12-02
Issue 952

CaseyCarter:


CaseyCarter:
This fixes the symptom - the manifestation in #952, but not the root cause. Something is misconstrained in the `view_adaptor` machinery. Removing the exception specifications from `range_access::begin_cursor`, the `view_adaptor::begin_cursor` overloads, and `view_adaptor::begin_cursor_` also corrects the problem in #952.

CaseyCarter:
> Either something is misconstrained in the `view_adaptor` machinery

There's an order-of-instantiation dependency: changing the return type of `view_adaptor::begin_cursor` from `adaptor_cursor_t<D>` to `decltype(adaptor_cursor_t<D>{})` brings the hard error back even with the exception specifications disabled. I'm tempted to simply remove the exception specifications and call it fixed.

CaseyCarter:
I'm going to merge this, and continue investigating the `view_adaptor` weirdness as a separate issue.

ericniebler:
Thanks for this!

PR #955, opened 2018-12-04
Added example of RandomAcess iterator.

tower120:
I have added "Create Custom Iterators" section. It is mainly consists from example code...
Since there is almost no information about basic_iterator, which seems to be the basis of library... Maybe it is better to have at least something.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -293,8 +293,8 @@ now.)*
 
 ## Create Custom Iterators
 
-Here is example of Range v3 compatible RandomAccess proxy iterator.
-Iterator return key/value (aka zip).
+Here is an example of Range v3 compatible RandomAccess proxy iterator.
+The iterator returns a key/value pair, like the `zip` view.
 
 ~~~~~~~{.cpp}
     #include <range/v3/utility/basic_iterator.hpp>
@@ -309,8 +309,8 @@ Iterator return key/value (aka zip).
         // to inject things into the public interface of the iterator
         struct mixin;
 
-        // This is for dereference operator.        
-        using value_type = ranges::common_pair<Key, Value>;
+        // This is for dereference operator.      
+        using value_type = std::pair<Key, Value>; 
         ranges::common_pair<Key&, Value&> read() const {
             return { *key_iterator, *value_iterator };
         }
@@ -373,19 +373,24 @@ Iterator return key/value (aka zip).
       std::vector<Value> values = {10};
       
       iterator iter(keys.begin(), values.begin());
-      auto   pair  = *iter;
+      ranges::common_pair<Key&, Value&> pair = *iter;
       Key&   key   = pair.first;
       Value& value = pair.second;
 
+      assert(&key   == &keys[0]);
+      assert(&value == &values[0]);      
+
       auto key_iter = iter.key_iterator();
+      assert(key_iter == keys.begin());
     }
 ~~~~~~~
 
-`read()` return references. So, we implicitly specify `value_type`.  
- `ranges::common_pair` have conversions:  
-`ranges::common_pair<Key&, Value&>` <=> `ranges::common_pair<Key, Value>`  
+`read()` returns references. So, we explicitly specify `value_type`.  
+ `ranges::common_pair` has conversions:  
+`ranges::common_pair<Key&, Value&>` <=> `ranges::common_pair<Key, Value>`.  
+All `ranges::common_pair`s converts to their `std::pair` equivalents, also.
 
-For more information, see [http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0186r0.html#basic-iterators-iterators.basic](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0186r0.html#basic-iterators-iterators.basic)
+For more information, see [http://wg21.link/P0186#basic-iterators-iterators.basic](http://wg21.link/P0186#basic-iterators-iterators.basic)
 
 
 ## Constrain Functions with Concepts
--- a/doc/index.md
+++ b/doc/index.md
@@ -341,7 +341,7 @@ Iterator return key/value (aka zip).
             Value& value;
 
             // make this conversion explicit, so
-            // common_reference<KeyValue, KeyValueRef> = KeyValueRef
+            // common_type_t<KeyValue, KeyValueRef> = KeyValueRef
             explicit operator KeyValue() const {
                 return { key, value };
             }
--- a/doc/index.md
+++ b/doc/index.md
@@ -297,6 +297,9 @@ Here is example of Range v3 compatible RandomAccess proxy iterator.
 Iterator return key/value (aka zip).
 
 ~~~~~~~{.cpp}
+    #include <range/v3/utility/basic_iterator.hpp>
+    #include <range/v3/utility/common_tuple.hpp>
+
     using KeyIter   = typename std::vector<Key>::iterator;
     using ValueIter = typename std::vector<Value>::iterator;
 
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -291,6 +291,101 @@ sentinel. That is only necessary if the underlying range's sentinel type models
 BidirectionalIterator. That's a finer point that you shouldn't worry about right
 now.)*
 
+## Create Custom Iterators
+
+Here is example of Range v3 compatible RandomAcess proxy iterator.
```
```suggestion
Here is example of Range v3 compatible RandomAccess proxy iterator.
```

doc/index.md
```diff
@@ -291,6 +291,101 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+            ValueIter value_iterator() {
+                return this->get().value_iterator;
+            }
+        };
+
+        // This is for dereference operator.
+        using cursor_value_t = std::pair<Key, Value&>;
+        cursor_value_t read() const {
+            return {*key_iterator, *value_iterator};
+        }
```
ericniebler: Your value type should not have reference semantics. `std::pair<Key, Value>`. Then `read()` should return something like `ranges::common_pair<Key&, Value&>` that makes the iterator satisfy the syntactic requirements of the `Readable` concept.

Proxies are hard to get right. I hope for C++?? we get the conversion operators in `tuple` and `pair` to make this all Just Work.

doc/index.md
```diff
@@ -291,6 +291,101 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+
+        bool equal(const cursor& other) const {
+            return key_iterator == other.key_iterator;
+        }
+
+        void next() { 
+            key_iterator++;
+            value_iterator++;
+        }
```
ericniebler: Pre-increment, not post-increment here and in `prev()` below.

doc/index.md
```diff
@@ -291,6 +291,101 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+        // arrow for operator-> . It is not required.
+        struct arrow_proxy
+        {
+            cursor_value_t value;
+            cursor_value_t* operator->() { return std::addressof(value); }
+            operator cursor_value_t*()   { return std::addressof(value); }
+        };
+        arrow_proxy arrow() const {
+            return arrow_proxy{ read() };
+        }
```
ericniebler: Remove this support for `operator->`. `it->x` is supposed to be semantically equivalent to `(*it).x`, and it is not in this case.

doc/index.md
tower120: > Your value type should not have reference semantics

I see that `reference_t` is `read()`'s return type.
What is `value_type_t` for this iterator/cursor? Does `basic_iterator` infer it somehow from `cursor`? Should/could it be specified directly?

---

>  I hope for C++?? we get the conversion operators in tuple and pair to make this all Just Work.

What about custom types, like:
```
struct KeyValue{
     Key& key;
     Value& value;
};
```

doc/index.md
tower120: Why `it->x` is not equivalent to `(*it).x` ?
`iterator->value` points to the same place as `(*iterator).value` ...

doc/index.md
ericniebler: The problem of defining a proxy iterator is hard. I suggest you read [my blog post](http://ericniebler.com/2015/02/13/iterators-plus-plus-part-2/) on the subject, and then try to implement a proxy iterator _and test that it satisfies range-v3's `InputIterator` concept_.

tower120: > I suggest you read my blog post on the subject

Actually I already read it... First two times when I read it, I just couldn't get why `value_type` is needed at all. Then I saw this https://stackoverflow.com/a/33115699...

BTW, there is `view::transform`, which can return `KeyValue` struct, or pair of references ... Maybe there is need to note, that such things are not valid... _Or they are?_

> and test that it satisfies range-v3's InputIterator concept.

It already satisfies it: https://wandbox.org/permlink/1humI5P3NagLS44W
Maybe, you mean that there should be also compare operators between them KeyValue and KeyValueRef?

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+    using ValueIter = typename std::vector<Value>::iterator;
+
+    struct cursor {
+
+        // basic_iterator derives from "mixin", if present, so it can be used
+        // to inject things into the public interface of the iterator
+        struct mixin;
+
+        // This is for dereference operator.        
+        using value_type = ranges::common_pair<Key, Value>;
```
ericniebler: I'm pretty sure this (the `value_type`) can be `std::pair<Key, Value>`.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+            , value_iterator(value_iterator)
+        {}
+
+        KeyIter   key_iterator;
+        ValueIter value_iterator;
+    };
+
+    struct cursor::mixin : ranges::basic_mixin<cursor>
+    {
+      using ranges::basic_mixin<cursor>::basic_mixin;
```
ericniebler: I _think_ you also need `mixin() = default;` also.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+    };    
+
+    using iterator = ranges::basic_iterator<cursor>;
+
+    void test(){
+      std::vector<Key>   keys   = {1};
+      std::vector<Value> values = {10};
+      
+      iterator iter(keys.begin(), values.begin());
+      auto   pair  = *iter;
```
ericniebler: A comment here to the effect that `pair` is actually a `ranges::common_pair<Key&, Value&>` might be helpful.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+      Key&   key   = pair.first;
+      Value& value = pair.second;
```
ericniebler: And for clarity, you could add:
```c++
assert(&key == &keys[0]);
assert(&value == &values[0]);
```

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+
+      auto key_iter = iter.key_iterator();
+    }
+~~~~~~~
+
+`read()` return references. So, we implicitly specify `value_type`.
```
ericniebler: "returns". And the definition of `value_type` is not implicit. It is explicit on line 313.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+      auto   pair  = *iter;
+      Key&   key   = pair.first;
+      Value& value = pair.second;
+
+      auto key_iter = iter.key_iterator();
+    }
+~~~~~~~
+
+`read()` return references. So, we implicitly specify `value_type`.  
+ `ranges::common_pair` have conversions:
```
ericniebler: "has"

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+      Key&   key   = pair.first;
+      Value& value = pair.second;
+
+      auto key_iter = iter.key_iterator();
+    }
+~~~~~~~
+
+`read()` return references. So, we implicitly specify `value_type`.  
+ `ranges::common_pair` have conversions:  
+`ranges::common_pair<Key&, Value&>` <=> `ranges::common_pair<Key, Value>`
```
ericniebler: If you change `value_type` to be `std::pair`, you should change it here too.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
...
+
+      auto key_iter = iter.key_iterator();
+    }
+~~~~~~~
+
+`read()` return references. So, we implicitly specify `value_type`.  
+ `ranges::common_pair` have conversions:  
+`ranges::common_pair<Key&, Value&>` <=> `ranges::common_pair<Key, Value>`  
+
+For more information, see [http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0186r0.html#basic-iterators-iterators.basic](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0186r0.html#basic-iterators-iterators.basic)
```
ericniebler: Future-proof this link by changing it to <http://wg21.link/P0186#basic-iterators-iterators.basic>. If there is ever an updated version of P0186, this will resolve to that instead.

doc/index.md
```diff
@@ -291,6 +291,103 @@ sentinel. That is only necessary if the underlying range's sentinel type models
 BidirectionalIterator. That's a finer point that you shouldn't worry about right
 now.)*
 
+## Create Custom Iterators
+
+Here is example of Range v3 compatible RandomAccess proxy iterator.
+Iterator return key/value (aka zip).
```
ericniebler: "The iterator returns a key/value pair, like the `zip` view."

doc/index.md
tower120: At least it compiles https://wandbox.org/permlink/sKNfDGGYPjdArgWa .

doc/index.md
tower120: Ugh, I meant explicit, sorry.

ericniebler:
Thanks!

ericniebler:
Your additions are now live on https://ericniebler.github.io/range-v3

PR #958, opened 2018-12-05
[partial_sum] Fix horrible bugginess due to defaulting the binary ope…

CaseyCarter:
…ration in `bind` instead of `operator()`

NOTE: THIS CHANGE IS SOURCE-BREAKING. `view::partial_sum()` previously yielded a closure that held `plus`, despite that `operator()` did *not* default its function argument to `plus`. The new state is that `view::partial_sum()` is ill-formed and should be replaced with `view::partial_sum`, and `operator()` defaults its second argument to `plus{}`.

As a result, this program fragment:
```c++
static int const some_ints[] = {0,1,2,3,4};
auto t1 = ranges::view::partial_sum(some_ints);
auto t2 = some_ints | ranges::view::partial_sum;
::check_equal(t1, t2);
```
Now works as intended, whereas previously `ranges::view::partial_sum(some_ints)` yielded a closure object waiting for a range to sum with the "binary operation" `make_iterator_range(some_ints)`.

### DIFF
--- a/test/view/partial_sum.cpp
+++ b/test/view/partial_sum.cpp
@@ -9,13 +9,13 @@
 //
 // Project home: https://github.com/ericniebler/range-v3
 
-#include <iterator>
 #include <functional>
+#include <iterator>
 #include <range/v3/core.hpp>
-#include <range/v3/view/partial_sum.hpp>
-#include <range/v3/view/counted.hpp>
-#include <range/v3/view/reverse.hpp>
 #include <range/v3/utility/copy.hpp>
+#include <range/v3/view/counted.hpp>
+#include <range/v3/view/partial_sum.hpp>
+#include <range/v3/view/reverse.hpp>
 #include "../simple_test.hpp"
 #include "../test_utils.hpp"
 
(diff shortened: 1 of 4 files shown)


CaseyCarter:
Poking @ericniebler for review. The code is fine, but we need to decide what (if anything) to do about the potential source breakage.

ericniebler:
Ugh. I think we take the breaking change, and add a release note.

Searching github, there aren't many uses of `view::partial_sum`, but unfortunately some of them are in Advent of Code examples have have been making their way around Twitter in recent days (and I'm sure you are aware). The responsible thing to do is to follow up on those Twitter threads and tell folks about the coming breakage and what they can do to make their code forward-compatible.

CaseyCarter:
Crisis averted. I've added a `bind` overload that is called by `view::partial_sum()` that preserves the old behavior and emits a deprecation message. LMK if you'd like something different and/or more explanatory in the deprecation message.

test/view/partial_sum.cpp
```diff
@@ -9,13 +9,14 @@
...
 // Project home: https://github.com/ericniebler/range-v3
 
-#include <iterator>
 #include <functional>
+#include <iterator>
 #include <range/v3/core.hpp>
-#include <range/v3/view/partial_sum.hpp>
+#include <range/v3/utility/copy.hpp>
 #include <range/v3/view/counted.hpp>
+#include <range/v3/view/iota.hpp>
```
CaseyCarter: This header's unnecessary - you changed the repro to use an array.

PR #965, opened 2018-12-18
Allow importing from existing CMake build tree

johelegp:
This change adds an additional way to consume Range-v3 with CMake, other than installing it to use with `find_package` or somehow making it available (say, git submodule or `FetchContent`) to use it with `add_subdirectory`. It permits pointing to an existing build tree of Range-v3 by setting `range-v3_DIR` variable to that directory, so that you can then use it with `find_package` as usual.

My intention is to allow my own libraries depending on Range-v3 to be used by defining their `_DIR` variable, as mentioned above, without needing Range-v3 to be installed.


ericniebler:
I know nothing about CMake, but if you say this makes your life easier, then (y).

PR #966, opened 2018-12-19
Don't use aligned new when the probe fails

CaseyCarter:
Good thing I added a probe to guard against using aligned new when the compiler supports it but the library doesn't. Shame that I didn't have cmake actually *do* anything with the probe results.

This should fix the "clang-5.0 -std=c++1z with old libstdc++" failure on the deep integration branch.


include/range/v3/utility/memory.hpp
```diff
@@ -43,11 +43,11 @@ namespace ranges
...
                 RANGES_EXPECT(count >= 0);
                 std::size_t n = static_cast<std::size_t>(count);
-                while(n > PTRDIFF_MAX / sizeof(T))
-                    n /= 2;
+                if (n > PTRDIFF_MAX / sizeof(T))
+                    n = PTRDIFF_MAX / sizeof(T);
 
                 void *ptr = nullptr;
-                for(; ptr == nullptr && n > 0; n /= 2)
+                for (; ptr == nullptr && n > 0; n /= 2)
```
ericniebler: These changes look unrelated.

test/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ target_link_libraries(rv3-tests INTERFACE range-v3)
 if (NOT RANGE_V3_THREAD_PROBE)
   target_compile_definitions(rv3-tests INTERFACE "RANGES_CXX_THREAD=0")
 endif()
+if (NOT RANGE_V3_ALIGNED_NEW_PROBE)
+  target_compile_definitions(rv3-tests INTERFACE "RANGES_CXX_ALIGNED_NEW=0")
+endif()
```
ericniebler: Some people don't use CMake. On the off-chance that we care about such people, I did some sleuthing and updated detail/config.hpp to suss out whether the stdlib supports `std::align_val_t` by checking version macros.

CaseyCarter: IIRC, libstdc++ versions are basically timestamps and not useful for feature versioning. For example, libstdc++ 6.4 was released after 7.1, so it will have a larger version number despite not implementing the all the features that the latter version does. It's been a while since I looked, though. We should ask @jwakely.

include/range/v3/utility/memory.hpp
CaseyCarter: They're related only in that I looked at the code and saw things that need fixing.

test/CMakeLists.txt
jwakely: That's right, see the docs for `_GLIBCXX_RELEASE` and `__GLIBCXX__` at https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html

The description for `__GLIBCXX__` links to https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html#abi.versioning.__GLIBCXX__ which says:

>  This macro can be used in code to detect whether the C++ Standard Library implementation in use is libstdc++, but is not useful for detecting the libstdc++ version, nor whether particular features are supported. The macro value might be a date after a feature was added to the development trunk, but the release could be from an older branch without the feature. For example, in the 5.4.0 release the macro has the value `20160603` which is greater than the `20160427` value of the macro in the 6.1.0 release, but there are features supported in the 6.1.0 release that are not supported in 5.4.0 release. You also can't test for the exact values listed below to try and identify a release, because a snapshot taken from the `gcc-5-branch` on 2016-04-27 would have the same value for the macro as the 6.1.0 release despite being a different version. Many GNU/Linux distributions build their GCC packages from snapshots, so the macro can have dates that don't correspond to official releases.

The first libstdc++ to support `std::align_val_t` was 7.1.0 which is also the first one to define `_GLIBCXX_RELEASE` so you're in luck. You can use `#if _GLIBCXX_RELEASE >= 7` to check for `std::align_val_t`.

ericniebler: Hot diggity dog! Thanks @jwakely. 👍 

@CaseyCarter I'm curious why even after merging this PR the new-clang&old-stdlibc++ build is still failing with errors about `std::align_val_t`. Maybe some cmake weirdness where the test is being run with different settings than the tests?

Regardless, I'll use `_GLIBCXX_RELEASE` and all should be well.

CaseyCarter: The problem seems to be `add_compile_definitions`. If I replace your `add_compile_definitions(RANGES_DEEP_STL_INTEGRATION=1)` with `ranges_append_flag(RANGES_DEEP_STL_INTEGRATION -DRANGES_DEEP_STL_INTEGRATION=1)`, the other compiler flags aren't corrupted.

CaseyCarter: Also: `add_compile_definitions` is new in cmake 3.12 - we need to bump `cmake_minimum_required` if we're going to use it.

ericniebler: If I use `add_compile_options(-DRANGES_DEEP_STL_INTEGRATION=1)`, will that work for MSVC or do I need to `add_compile_options(/DRANGES_DEEP_STL_INTEGRATION=1)`?

ericniebler: Grr, still busted.

PR #968, opened 2018-12-20
cmake refactoring:

CaseyCarter:
* Remove the `rv3-tests` interface library and abandon the notion of separate flags only for the tests.

* Use `rv3_add_test` to simplify `examples/CMakeLists.txt`.


CaseyCarter:
Bah - I should have merged to deep-integration and backported instead.

PR #980, opened 2019-01-08
adds adjacent_remove_if algorithm and action (trunk)

cjdb:
The adjacent_remove_if algorithm and action are similar to
adjacent_remove_if_view in functionality.

### DIFF
--- a/include/range/v3/action/adjacent_remove_if.hpp
+++ b/include/range/v3/action/adjacent_remove_if.hpp
@@ -86,8 +86,8 @@ namespace ranges
                         "by the projection function, or of the range's value type if no projection "
                         "is specified.");
                     CONCEPT_ASSERT_MSG(Permutable<I>(),
-                        "The iterator type of the range passed to action::remove_if must allow its "
-                        "elements to be permutaed; that is, the values must be movable and the "
+                        "The iterator type of the range passed to action::adjacent_remove_if must allow its "
+                        "elements to be permuted; that is, the values must be movable and the "
                         "iterator must be mutable.");
                 }
             #endif
--- a/include/range/v3/action/adjacent_remove_if.hpp
+++ b/include/range/v3/action/adjacent_remove_if.hpp
@@ -1,0 +1,105 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Eric Niebler
+//  Copyright Christopher Di Bella
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+#ifndef RANGES_V3_ACTION_ADJACENT_REMOVE_IF_HPP
+#define RANGES_V3_ACTION_ADJACENT_REMOVE_IF_HPP
+#include <range/v3/algorithm/adjacent_remove_if.hpp>
+#include <range/v3/range_fwd.hpp>
+#include <range/v3/range_traits.hpp>
+#include <range/v3/utility/functional.hpp>
+#include <range/v3/action/action.hpp>
+#include <range/v3/action/erase.hpp>
+#include <range/v3/utility/static_const.hpp>
+#include <utility>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        /// \addtogroup group-actions
+        /// @{
+        namespace action
+        {
+            struct adjacent_remove_if_fn
+            {
+            private:
+                friend action_access;
+                template<typename Pred, typename Proj = ident, CONCEPT_REQUIRES_(!Range<Pred>())>
+                static auto bind(adjacent_remove_if_fn adjacent_remove_if, Pred pred, Proj proj = {})
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    std::bind(adjacent_remove_if, std::placeholders::_1, protect(std::move(pred)),
+                        protect(std::move(proj)))
+                )
+            public:
+                struct ConceptImpl
+                {
+                    template<typename R, typename Pred, typename Proj = ident,
+                        typename I = iterator_t<R>>
+                    auto requires_() -> decltype(
+                        concepts::valid_expr(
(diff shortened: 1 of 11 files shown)


cjdb:
I might need help with getting clang-cl to pass.

CaseyCarter:
> I might need help with getting clang-cl to pass.

`-permissive-` tells MSVC to be more conforming, including performing two-phase name lookup. clang-cl accepts and ignores `-permissive-`. (Do you see where this is headed?) The unqualified `adjacent_remove_if` call in `adjacent_remove_if_fn::operator()` is resolving to `action::adjacent_remove_if` - it's recursive!

The fix for this particular problem is obvious and simple: qualify the call as `ranges::adjacent_remove_if`. I question if we *want* to apply that fix, however: we don't support MSVC's permissive mode, why should we support clang-cl's? We test MSFTL with `clang-cl -fno-ms-compatibility -fno-delayed-template-parsing`, which notably encourages it to be more conforming and less old-MSVC-ish, I think range-v3 should use the same flags.

I'll push a cmake change.

test/algorithm/adjacent_remove_if.cpp
```diff
@@ -0,0 +1,199 @@
...
+        CHECK(r == ia + sa-3);
+        CHECK(ia[0].i == 0);
+        CHECK(ia[1].i == 1);
+        CHECK(ia[2].i == 4);
+        CHECK(ia[3].i == 2);
+        CHECK(ia[4].i == 4);
+        CHECK(ia[5].i == 2);
+    }
+
+#if 0 // Should this even be included?
```
CaseyCarter: Why not include it? (It certainly shouldn't be present-but-disabled.)

include/range/v3/action/adjacent_remove_if.hpp
```diff
@@ -0,0 +1,105 @@
...
+                    CONCEPT_ASSERT_MSG(IndirectInvocable<Proj, I>(),
+                        "The projection function must accept objects of the iterator's value type, "
+                        "reference type, and common reference type.");
+                    CONCEPT_ASSERT_MSG(IndirectRelation<Pred, projected<I, Proj>>(),
+                        "The predicate passed to action::adjacent_remove_if must accept objects returned "
+                        "by the projection function, or of the range's value type if no projection "
+                        "is specified.");
+                    CONCEPT_ASSERT_MSG(Permutable<I>(),
+                        "The iterator type of the range passed to action::remove_if must allow its "
+                        "elements to be permutaed; that is, the values must be movable and the "
```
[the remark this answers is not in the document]
CaseyCarter: > (Please double-check that this typo didn't come from the header I assume you copied this code from originally.)

Which it did. Please fix all three occurrences of "permutaed" ;)

test/algorithm/adjacent_remove_if.cpp
cjdb: > Why not include it?

I wasn't sure if range-v3 would be dropping support for rvalues like C++20.

> (It certainly shouldn't be present-but-disabled.)

Disabled to bring your attention to it (I guess I could have made a review comment).

include/range/v3/algorithm/adjacent_remove_if.hpp
```diff
@@ -0,0 +1,80 @@
...
+            template<typename I, typename S, typename Pred, typename Proj = ident,
+                CONCEPT_REQUIRES_(ForwardIterator<I>() && Sentinel<S, I>() &&
+                    IndirectRelation<Pred, projected<I, Proj>>() &&
+                    Permutable<I>())>
+            I
+            operator()(I first, S last, Pred pred = {}, Proj proj = {}) const
+            {
+                auto end = next(first, last);
+                while (first != end) {
+                    first = adjacent_find(first, end, std::ref(pred), std::ref(proj));
```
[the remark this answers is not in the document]
cjdb: FWIW `remove_if` also uses `reference_wrapper`. I can change both if you like?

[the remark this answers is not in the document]
CaseyCarter: If there's no consistency to be maintained, then we'll just leave these as-is.

ericniebler:
@CaseyCarter do you approve of the changes to this PR?

include/range/v3/algorithm/adjacent_remove_if.hpp
```diff
@@ -0,0 +1,89 @@
...
+                CONCEPT_REQUIRES_(Permutable<I>() && Sentinel<S, I>() &&
+                    IndirectRelation<Pred, projected<I, Proj>>())>
+            I
+            operator()(I first, S last, Pred pred = {}, Proj proj = {}) const
+            {
+                first = adjacent_find(std::move(first), last, std::ref(pred), std::ref(proj));
+                if (first != last)
+                {
+                    auto i = next(first);
+                    if (i != last)
```
CaseyCarter: `i != last` is always true here:
The postcondition of `adjacent_find` is that it either returns `last` or the first iterator `x` such that `pred(proj(x), proj(x+1))` is `true`. If `first != last`, then the predicate must have held, so `next(first)` must be a dereferenceable iterator and `next(first) != last` must be true.

include/range/v3/algorithm/adjacent_remove_if.hpp
```diff
@@ -0,0 +1,89 @@
...
+                    {
+                        for (auto j = next(i); j != last; ++i, (void)++j)
+                        {
+                            if (!invoke(pred, invoke(proj, *i), invoke(proj, *j)))
+                            {
+                                *first = iter_move(i);
```
CaseyCarter: Hrm. I note that all the flavors of `adjacent_remove_if` remove all but the *last* element from any run of equivalent elements, and so does `ranges::unique`, but `std::unique` and `std::ranges::unique` are specified to remove all but the *first* element from any run of equivalent elements. [We have us a bug](https://github.com/ericniebler/range-v3/issues/1003).

CaseyCarter: (I'm not expecting you to fix this - certainly not as part of this checkin. Just sharing.)

CaseyCarter:
> @CaseyCarter do you approve of the changes to this PR?

One tiny simplification to make, and it's good to go.

PR #981, opened 2019-01-10
Kill versioning namespaces

ericniebler:


CaseyCarter:
I've applied a similar change to https://github.com/ericniebler/meta/pull/60, which *cough* is awaiting review before I update the copies of meta in range-v3 & cmcstl2.

ericniebler:
Wow, that looks like a lot of work. Thanks. I've enabled AppVeyor for this repo, but you might need to push a dummy commit here to get it to test the PR.

Also, I'll just leave this here.
DON'T CROSS THE STREAMS

PR #984, opened 2019-01-12
New meta

ericniebler:


include/meta/meta.hpp
```diff
@@ -2693,14 +2693,14 @@ namespace meta
     // count_if
     namespace detail
     {
-#ifdef META_CONCEPT
+#if defined(META_CONCEPT) && META_CXX_FOLD_EXPRESSIONS
```
CaseyCarter: This doesn't use fold expressions now, Casey.

include/meta/meta_fwd.hpp
```diff
@@ -248,24 +258,29 @@ namespace meta
         typename T::type::value_type;
     }
     && Same<typename T::value_type, typename T::type::value_type>
+#if META_CXX_TRAIT_VARIABLE_TEMPLATES
```
CaseyCarter: To me, this conditional is worth the trouble given that it's cheaper to instantiate a variable template than a class template. YMMV.

ericniebler:
Something about this seems to send gcc >5 into hyperspace.

CaseyCarter:
> Something about this seems to send gcc >5 into hyperspace.

Yeah, they're hanging forever compiling utility.meta w/concepts.

ericniebler:
Is Travis misconfigured? It looks like the not-concepts builds are also failing, but I may have botched the Travis config.

CaseyCarter:
> Is Travis misconfigured? It looks like the not-concepts builds are also failing, but I may have botched the Travis config.

No, Travis is right and I am wrong: It's not concepts. All GCC builds are hanging in utility.meta on the bits annotated with:
```c++
// I'm guessing this failure is due to GCC #64970
// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64970
#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 5
```

They're fine if I comment out those two sections.

CaseyCarter:
(Note to self: unify utility.meta from range-v3 / cmcstl2 upstream.)

ericniebler:
Only the line:

```c++
static_assert(!can_invoke<lambda<_a, _b, _c, _args, defer<std::pair, _a, _a>>>::value, "");
```

is causing gcc to spin.

include/meta/meta.hpp
```diff
@@ -3286,7 +3286,7 @@ namespace meta
                 bind_back<quote_trait<subst1_>, back<As>, drop_c<Ts, size<As>{} - 2>>>>,
             list<back<As>>>;
 
-#ifdef META_CONCEPT
+#if 0//def META_CONCEPT
         template <List As, List Ts>
         requires (_v<size<Ts>> + 2 >= _v<size<As>>)
         using substitutions = substitutions_<As, Ts>;
```
ericniebler: This definition of `substitutions` in particular is what is giving gcc heartburn. I didn't investigate deeply.

The other problems below seem to be that gcc is ignoring some `requires` clauses for some reason. Also not investigated.

ericniebler:
@CaseyCarter I'm planning to merge this despite the hacky workarounds. I don't want to put off merging from master for too long. Chime in if you can think of better workarounds.

include/meta/meta_fwd.hpp
ericniebler: But `is_integral_v<T>` is required to be defined as `is_integral<T>::value`. What exactly do we gain?

ericniebler:
> we should consider inserting meta as a submodule in range-v3 / cmcstl2 to avoid divergence.

That makes it slightly harder to consume range-v3, but I suppose the v1.0 release is the time to make such a change, if ever there was one. I will also spin CPP into its own repo and add it as a submodule.

That requires additional work, though. There should be conan integration among the 3 repos, for instance. Not sure what is needed from the CI PoV.

include/meta/meta_fwd.hpp
CaseyCarter: It's not observable whether `is_foo_v` instantiates `is_foo` since `is_foo` isn't user-specializable.

CaseyCarter: Hrrm, except that it is, because objects have addresses.

CaseyCarter:
I suppose we could instead be more disciplined with `meta` sync:
* setup CI for meta that builds range-v3 and cmcstl2,
* always check changes into the `meta` repo first,
* immediately copy the new headers into both `range-v3` and `cmcstl2`.
Which is simple, but an ongoing cost rather than a one-time cost. We don't really change `meta` enough to make "upfront one-time cost with mild user annoyance" a clear winner, however - it's not like there are 5 devs each making 4 changes every day. 

Maybe avoid the submodule for now. I do think a "unified" CI setup for meta is still a great idea.

PR #987, opened 2019-01-12
semiregular: converting constructor should be conditionally explicit

### LINKED ISSUE #986: Can't move construct semiregular from non-capturing lambda
Branch: v1.0-beta
The following extract fails to compile because the storage for the lambda couldn't be constructed.
```C++
std::string_view{} | ranges::view::split([](char c) { return true; });
```
[...]/build/clang/_deps/range-v3-src/include/range/v3/view/split.hpp:265: error: no viable conversion from `typename std::remove_reference<(lambda at [...]/src/gfm.cpp:426:36) &>::type` (aka `(lambda at [...]/src/gfm.cpp:426:36)`) to `semiregular_t<(lambda at [...]/src/gfm.cpp:426:36)>` (aka `ranges::semiregular<(lambda at [...]/src/gfm.cpp:426:36)>`)

                        predicate_pred<Rng, Fun>{std::move(fun)}};
                                                 ^~~~~~~~~~~~~~

CaseyCarter:
Drive-by: parenthesize non-primary-expression in requires-clause

Fixes #986.

### DIFF
--- a/include/range/v3/utility/semiregular.hpp
+++ b/include/range/v3/utility/semiregular.hpp
@@ -123,8 +123,16 @@ namespace ranges
         template<typename U>
         explicit constexpr CPP_ctor(semiregular)(U &&u)(
             noexcept(std::is_nothrow_constructible<T, U>::value)
-                requires not defer::Same<uncvref_t<U>, semiregular> &&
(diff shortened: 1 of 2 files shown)


ericniebler:
@CaseyCarter what is going on with this PR?

PR #998, opened 2019-01-22
[WIP] Add some introductory docs

daixtrose:
- extend the set of supported compilers
- add a description about what the library provides
- give an example of how coding with range-v3 looks like 
- point to similar approaches in C# 
- point to foundation in functional programming

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -59,13 +59,13 @@ Range v3 is a generic library that augments the existing standard library with f
 
 For every STL algorithm this library provides a version that accepts a *range* instead of a pair of iterators. 
 
-All transforming and filtering functions are composable. Applying them to a range yields a so-called **lazy** representation of the result, which can either be evaluated to obtain the results or can be passed to another function.  
+Pushing this further the library distinguishes eager, mutating in-place algorithms e.g for sorting (called **actions**) and lazy, non-mutating algorithms e.g. for filtering (called **views**)
 
-This avoids temporary results while allowing a declarative and easy-to-understand programming style. Generally this is possible without sacrificing performance.  
+> @EricNiebler: What separates the [algorithms](https://github.com/ericniebler/range-v3/tree/master/include/range/v3/algorithm) from the views and actions? Only found [a distinction of views and actions](http://ericniebler.com/2014/11/23/container-algorithms/). Some of them are dimension-reducing ... 
 
-> TODO: distinguish and explain the difference of algorithms, views, and actions
+All functions are composable. E.g. applying a lazy, non-mutating view to a range does not lead to an evaluation, but yields an object representing the result, which can either be evaluated later to obtain the results or be passed to another view, action or algorithm.  
 
-Range v3 also provides in-place modifying funnctions e.g. for sorting. These are called **action**s.  
+This avoids temporaries while allowing a declarative and easy-to-understand programming style. Generally this is possible without sacrificing performance.  
 
 Range v3 is a contribution to making C++ easier to write and reason about. With Range v3 one can write code this way: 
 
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+
+Range v3 is a generic library that augments the existing standard library with facilities for working with *ranges*. This means that you do not have to fiddle with `begin` and `end` iterators directly, but pass something around, that represents both in one object. 
+
+For every STL algorithm this library provides a version that accepts a *range* instead of a pair of iterators. 
+
+All transforming and filtering algorithms are composable. This means applying an algorithm to a range yields a **lazily** evaluated representation (a so-called **view**) of the result, which can either be evaluated to obtain the results of the applied transformation or can be passed to another algorithm.  
+
+This avoids temporary results while allowing a declarative and easy-to-understand programming style. Generally this is possible without sacrificing performance.  
+
+Range v3 also provides in-place modifying algorithms e.g. for sorting. These are called **action**s.
```
ericniebler: Actions are distinct from algorithms. The important bit about actions is that they are _composable_. The algorithms are not.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
-of iterators, although they need not be implemented that way. Bundling begin/end
-iterators into a single object brings several benefits.
+
+## What This Library Provides
+
+Range v3 is a generic library that augments the existing standard library with facilities for working with *ranges*. This means that you do not have to fiddle with `begin` and `end` iterators directly, but pass something around, that represents both in one object. 
+
+For every STL algorithm this library provides a version that accepts a *range* instead of a pair of iterators. 
+
+All transforming and filtering algorithms are composable. This means applying an algorithm to a range yields a **lazily** evaluated representation (a so-called **view**) of the result, which can either be evaluated to obtain the results of the applied transformation or can be passed to another algorithm.
```
ericniebler: The views are distinct from the algorithms. Range-v3 is built on 3 pillars: algorithms, views, and actions. We need to be crisp about our terminology. If we call views algorithms we muddy the waters.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+        [](PolarCoordinate pc) { 
+          return std::isinf(pc.distance) 
+            || std::isnan(pc.distance); 
+            };
+
+    auto filteredAndInCartesianCoordinates = coordinates
+        | remove_if(outsideOfReasonableRange)
+        | remove_if(is_inf_or_nan)
+        | transform(polarToCartesian)
+        | reverse;
```
ericniebler: Please be explicit about namespaces here. `| view::remove_if(...)`, etc. And have a `using namespace ranges;` or something at the top of the example.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+~~~~~~~
+
+
+
+
+## What is a Range?
+
+It may come as a surprise, but giving a short, comprehensive definition of a *range* is much harder than one might expect. 
+
+First of all a range is something you can iterate over (*range* used to be spelled *iterable*). One possibility to describe an iterable is given in the [`IEnumerator`  interface](https://docs.microsoft.com/de-de/dotnet/api/system.collections.ienumerator) definition of the `C#` programming language which has a `bool MoveNext()` method and a `Current` property. Due to consistency with the STL and for efficiency reasons the C++ representation differs. Everything that can be passed to `begin` and `end` can be considered to be a range.
```
ericniebler: Strike "(range used to be called iterable)". Use the term "range" consistently. The comparison to the C# `IEnumerable` interface will not help those people who don't know C#. I'd rather with stick to C++. A range is anything we can call `std::begin` and `std::end` on to retrieve an iterator and a sentinel. It can be a value or a reference, a container or a view, or even a built-in C+ array.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+
+
+
+## What is a Range?
+
+It may come as a surprise, but giving a short, comprehensive definition of a *range* is much harder than one might expect. 
+
+First of all a range is something you can iterate over (*range* used to be spelled *iterable*). One possibility to describe an iterable is given in the [`IEnumerator`  interface](https://docs.microsoft.com/de-de/dotnet/api/system.collections.ienumerator) definition of the `C#` programming language which has a `bool MoveNext()` method and a `Current` property. Due to consistency with the STL and for efficiency reasons the C++ representation differs. Everything that can be passed to `begin` and `end` can be considered to be a range. 
+ 
+So a range can be loosely thought of a pair of iterators, although they need not be implemented that way.
```
ericniebler: That is closer to what a view is.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+
+A range might be a half open interval. Which means a range might have an infinite number of elements, like e.g. all positive integers starting from 42. It can be shown that in this case the iterators which represent `begin` and `end` of the interval must be of different types. See [this blogpost](http://ericniebler.com/2015/02/03/iterators-plus-plus-part-1/) and follow-ups for the details.
```
ericniebler: Not true, actually! I can implement an infinite range for which `begin` and `end` have the same type. Sentinels exist because in some cases it leads to better codegen and also because sentinels are often easier to implement.

doc/index.md
```diff
@@ -46,16 +46,65 @@ file and the CREDITS file for the licensing and acknowledgements.
...
+ 
+So a range can be loosely thought of a pair of iterators, although they need not be implemented that way.
+
+A range might be a half open interval. Which means a range might have an infinite number of elements, like e.g. all positive integers starting from 42. It can be shown that in this case the iterators which represent `begin` and `end` of the interval must be of different types. See [this blogpost](http://ericniebler.com/2015/02/03/iterators-plus-plus-part-1/) and follow-ups for the details.  
+
+## Theoretical Background
+
+C++ range and C# IEnumerable together with their corresponding algorithms share the same theoretical foundation from functional programming, especially the [concept of a monad](https://bartoszmilewski.com/2014/10/17/c-ranges-are-pure-monadic-goodness/). 
+
+All of C#'s [LINQ](https://en.wikipedia.org/wiki/Language_Integrated_Query) operations (which in fact are database queries on containers and other, similar objects)  have  counterparts in C++ ranges. E.g. `GroupBy` is called `group_by` in C++ and `SelectMany`is denoted `join`.
```
ericniebler: Again, I would leave C# out of it. In a section called "Theoretical Background" I would expect an appeal to math: namely Stepanov's Elements of Programming, which is based on Peano's axioms (iterators), and category theory. I would be fine with that. Then you can say that `join` is `bind` in the `List` monad.

You could additionally have a section describing similar functionality in other languages. There you can describe C#'s LINQ, the Haskell Prelude, python's list comprehensions and generators, and perhaps others.

Careful with `GroupBy` though. There is an expectation that a group-by operator would sort and then group. Range-v3's `group_by` does not do that.

daixtrose:
I already feared I would take more from this than I can give. Many thanks for the detailed comments. I am on a business trip the next 2 days and hopefully find time during the weekend to rephrase it. Stay tuned.

doc/index.md
daixtrose: > Strike "(range used to be called iterable)".

Well, I stole this information here (and found it useful to understand what it is)
image

daixtrose: I agree with C# comparison to be moved into an extra section about other languages

h-2:
For reference, I have written some longer introductions in the documentation to our library which makes heavy use of ranges:
http://docs.seqan.de/seqan/3.0.0-master-user/tutorial_ranges.html

Feel free to use any of that that you find useful.

agurtovoy:
For whoever stumbles upon this, the link above has moved to http://docs.seqan.de/seqan/3-master-user/tutorial_ranges.html

ericniebler:
@agurtovoy wrote:
> http://docs.seqan.de/seqan/3-master-user/tutorial_ranges.html

Very nice. Considering that this PR never quite got there, I'm closing it out. I've opened a separate issue (#1379) to mine seqan's docs for introductory material.

Issue #988, opened 2019-01-15
O(1) unordered_remove algorithm+action ?


tower120:
`action::remove_if` erase elements that satisfies predicate with O(n) complexity, and keep order.
Overwriting each erased element with the last one, have O(1) complexity, but loose order.
This unordered remove useful when using vector-alike container as fast iterable "set"/unordered list.

I would like to make a PR with this feature. But not sure, does range library actually need this.

In essence, it looks like this:
algorithm:

```cpp
    template<typename I, typename C, typename P = ranges::ident>
    I unordered_remove_if(I begin, I end, C pred, P proj = P{})
    {
        using namespace ranges;
        CONCEPT_ASSERT(BidirectionalIterator<I>());

        I i = begin;
        while(i!=end)
        {
            if((invoke(pred, invoke(proj, *i))))
            {
                --end;
                if (i==end) break;

                *i = iter_move(end);
            } else {
                ++i;
            }
        }
        return end;
    }
```

action:
```cpp
template<typename Rng, class Pred, class Proj>
Rng operator()(Rng&& rng, Pred pred, Proj proj) const{
    auto it = unordered_remove_if(ranges::begin(rng), ranges::end(rng), std::move(pred), std::move(proj));
    ranges::erase(rng, it, ranges::end(rng));
    return static_cast<Rng&&>(rng);
}
```

CaseyCarter:
I think this is a useful algorithm - it's basically `partition` except that it uses move assignment instead of swap - although I'd name it `unstable_remove_if` (by analogy to the `stable_XXX` algorithms) and implement it in terms of `find_if` and `find_if_not` (I'm a big fan of "using the library to implement the library"):
```c++
namespace ranges 
{
    inline namespace v3 
    {
        template <typename I, typename Pred, typename Proj = identity,
            CONCEPT_REQUIRES_(BidirectionalIterator<I>() &&
                Permutable<I>() &&
                IndirectUnaryPredicate<Pred, projected<I, Proj>>())>
        RANGES_CXX14_CONSTEXPR I 
        unstable_remove_if(I first, I last, Pred pred, Proj proj = {})
        {
            while(true)
            {
                first = find_if(std::move(first), last, pred, proj);
                last = find_if_not(
                    make_reverse_iterator(std::move(last)),
                    make_reverse_iterator(first),
                    pred, proj).base();
                if (first == last) return last;
                *first = iter_move(--last);
            }
        }
    }
}
```

tower120:
[...]

PR #1010, opened 2019-01-27
Implementing #988 unstable_remove_if

tower120:
Implementing https://github.com/ericniebler/range-v3/issues/988

`unstable_remove_if` algorithm's `last` must be BidirectionalIterator. Because of this, sentinel is not used. Sentinel concept does not satisfies BidirectionalIterator.

If it is allowed to require Sentinel to be BidirectionalIterator - I'll update this PR.

For example, `algorithm/reverse.hpp`, accept `end` as Sentinel, but internally casts everything to same Iterator. Thus it is semantically equivalent to BoundedRange. IOW - `algorithm\reverse` lying, saying it can accept Sentinel.


ericniebler:
If passed a sentinel, this algorithm _can_ do an O(N) probe for the end iterator. I'm on the fence about whether it should.

tower120:
>  I'm on the fence about whether it should.

It may become slower then remove algorithm, for container with huge count of pointer-sized (trivially copyable) elements. As is, unstable_remove_if should be theoretically faster (it doesn't do unnecessary memory writes) for this case.
Slow down comes from the fact, that if whole container does not fit in cache, then there will be twice more memory reads.

... From the other side... Can this action be applied to the range with sentinel? You got range with sentinel at the end - from views where end is not known beforehand. This is container mutating algorithm/action - it shouldn't be applied to the result of something like `view::remove_if`.

I am for the non-sentinel version (as here https://github.com/ericniebler/range-v3/pull/1010/commits/660564cb6585525eae76aa488a523e372e663fff - with bounded range). User can see clearly what this algorithm/action stands for just by looking at concept constraints. If user need to work with sentinel - user can probe end iterator manually.

ericniebler:
@CaseyCarter thoughts? This would be the first algorithm that refused to accept iterator/sentinel. If we are certain there is no implementation that avoids an O(_N_) probe for the end, then I _think_ I am OK with this as-is.

CaseyCarter:
> @CaseyCarter thoughts? This would be the first algorithm that refused to accept iterator/sentinel. If we are certain there is no implementation that avoids an O(_N_) probe for the end, then I _think_ I am OK with this as-is.

Seriously late response: there is an implementation that avoids the O(N) probe - the stable algorithm. Yes, it performs O(size) move assignments instead of O(#removed) assignments, but in practice I'd expect it to outperform "scan for the end iterator and then call the unstable algorithm".

I think I'd prefer for this algorithm to fallback to the stable algorithm for non-common ranges, so it could be the "I don't care about stability" form of `remove_if`.

tower120:
> this algorithm to fallback to the stable algorithm for non-common ranges

non-common ranges is all unbounded ranges? How do you know conditions when `remove_if` becomes faster?

> but in practice I'd expect it to outperform "scan for the end iterator and then call the unstable algorithm".

Here is actual benchmark.

https://wandbox.org/permlink/WHMLM20UpJR7Ps2B

```cpp
#include <iostream>

#include <range/v3/action/remove_if.hpp>
#include <range/v3/action/unstable_remove_if.hpp>
#include <chrono>
#include <random>

template<class TimeUnit = std::chrono::milliseconds, class Closure>
static auto measure(Closure&& closure){
    using namespace std::chrono;
    high_resolution_clock::time_point t1 = high_resolution_clock::now();
    closure();
    high_resolution_clock::time_point t2 = high_resolution_clock::now();

    return duration_cast<TimeUnit>(t2-t1).count();
}

template<class TimeUnit = std::chrono::milliseconds, class Closure>
static auto benchmark(unsigned int times, Closure&& closure){
    return measure<TimeUnit>([&]() {
        for (unsigned int i = 0; i < times; ++i) closure();
    });
}

template<std::size_t payload_size>
struct Data{
    bool need_remove;
    std::byte payload[payload_size];
};


int main() {
    int times = 1000;
    int size  = 10'000;
    int elements_to_delete = 1'000;

    using T = Data<32>;

    // make data
    using List = std::vector<T>;
    List src;

    std::random_device rd;
    std::mt19937 eng{rd()};
    std::uniform_int_distribution<int> distr(0, size/elements_to_delete);
    for(int i=0;i<size;++i){
        const int value = distr(eng);
        const bool need_remove = (value==0);
        src.emplace_back(T{need_remove});
    }

    const auto pred = [](const T& data) -> bool { return data.need_remove; };

    {
        auto t = benchmark(times, [&]() {
            static List list;
            list = src;
            list |= ranges::action::remove_if(pred);
        });
        std::cout << t << std::endl;
    }

    {
        auto t = benchmark(times, [&]() {
            static List list;
            list = src;
            list |= ranges::action::unstable_remove_if(pred);
        });
        std::cout << t << std::endl;
    }

    {
        auto t = benchmark(times, [&]() {
            static List list;
            list = src;

            auto iter = list.begin();
            auto sent = list.end();
            while(iter!=sent){
                ++iter;
                iter = sent;
            }

            auto last = ranges::unstable_remove_if(list.begin(), iter, pred);
            list.erase(last, list.end());
        });
        std::cout << t << std::endl;
    }

    return 0;
}
```

On my machine I have:
MinGW
remove_if : 62
unstable_remove_if : 31
loop to the end + unstable_remove_if : 46

VisualStudio:
remove_if : 35
unstable_remove_if : 26
loop to the end + unstable_remove_if : 28

It is arguable that `remove_if` faster then `O(N) probe` + `unstable_remove_if ` even for POD type. Non-trivially movable data will show even worse results for `remove_if `.

PR #1013, opened 2019-01-28
Do as the std does and require Views to be copy constructible

ericniebler:
No doubt raising the ire of @CaseyCarter.

### DIFF
--- a/include/range/v3/experimental/utility/generator.hpp
+++ b/include/range/v3/experimental/utility/generator.hpp
@@ -1,3 +1,4 @@
+/// \file
 // Range v3 library
 //
 //  Copyright Casey Carter 2017
@@ -14,15 +15,17 @@
 
 #include <range/v3/detail/config.hpp>
 #if RANGES_CXX_COROUTINES >= RANGES_CXX_COROUTINES_TS1
+#include <atomic>
 #include <cstddef>
 #include <exception>
 #include <utility>
 #include <experimental/coroutine>
 #include <meta/meta.hpp>
+#include <concepts/concepts.hpp>
+#include <range/v3/range_fwd.hpp>
 #include <range/v3/range/traits.hpp>
 #include <range/v3/view/facade.hpp>
 #include <range/v3/utility/box.hpp>
-#include <concepts/concepts.hpp>
 #include <range/v3/utility/semiregular.hpp>
 #include <range/v3/utility/swap.hpp>
 #include <range/v3/iterator/default_sentinel.hpp>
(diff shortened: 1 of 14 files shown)


include/range/v3/view/istream.hpp
```diff
@@ -96,7 +96,7 @@ namespace ranges
         template<typename Val>
         inline auto istream(std::istream &sin) ->
             CPP_ret(istream_view<Val>)(
-                requires DefaultConstructible<Val>)
+                requires CopyConstructible<Val>)
```
CaseyCarter: This should be `CopyConstructible && DefaultConstructible`.

test/debug_view.hpp
```diff
@@ -41,7 +41,7 @@ struct debug_input_view
       : debug_input_view{data, N}
     {}
     constexpr /*c++14*/
-    debug_input_view(debug_input_view &&that) noexcept
+    debug_input_view(debug_input_view const &that) noexcept
```
CaseyCarter: This copy constructor doesn't meet the semantic requirements of `CopyConstructible`.

test/debug_view.hpp
```diff
@@ -54,7 +54,7 @@ struct debug_input_view
         }
     }
     constexpr /*c++14*/
-    debug_input_view &operator=(debug_input_view &&that) noexcept
+    debug_input_view &operator=(debug_input_view const &that) noexcept
```
CaseyCarter: Ditto this doesn't meet the semantic requirements of `Copyable`.

include/range/v3/experimental/utility/generator.hpp
```diff
@@ -73,41 +83,71 @@ namespace ranges
...
+            {
+                if(*this)
+                {
+                    that.copied_.store(true, std::memory_order_relaxed);
+                    base().promise().refcount_.fetch_add(1, std::memory_order_relaxed);
+                }
+            }
+            ~coroutine_owner()
+            {
+                if(base() && (!copied_.load(std::memory_order_relaxed) ||
```
CaseyCarter: `relaxed` is too weak here to guarantee that relaxed stores from other threads happen before this load, no?

ericniebler: @lewissbaker When you are back from vacation, can you chime in here since you talked me into making these `relaxed`?

test/debug_view.hpp
ericniebler: Picky, picky.

ericniebler: How about now?

CaseyCarter: It looks right now.

PR #1021, opened 2019-01-30
partial_sum_view refactor:

### LINKED ISSUE #668: Poor performance compared to STL
See https://stackoverflow.com/questions/44413446/why-is-range-v3-slower-than-the-stl-in-this-example

CaseyCarter:
* Replace bogus `ConvertibleTo<invoke_result_t, range_value_t>` constraint
* Don't use `view_adaptor`
* `RANGES_UNIQUE_ADDRESS` all the things
* Never model `CommonRange`. There's little to be gained by making a forward range common, and doing so harms perf on the benchmark (https://stackoverflow.com/questions/44413446/why-is-range-v3-slower-than-the-stl-in-this-example).

This brings the slowdown relative to STL in the benchmark down to 1.6x, which seems to be the best we can do: my "put everything in the iterator" investigations were actually slower. I'll let you decide if this is good enough to close #668.

### DIFF
--- a/include/range/v3/view/partial_sum.hpp
+++ b/include/range/v3/view/partial_sum.hpp
@@ -14,126 +14,147 @@
 #ifndef RANGES_V3_VIEW_PARTIAL_SUM_HPP
 #define RANGES_V3_VIEW_PARTIAL_SUM_HPP
 
-#include <utility>
-#include <iterator>
 #include <functional>
+#include <iterator>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


CaseyCarter:
Note: I'm introducing divergence between master and 1.0 here: shout if you'd prefer me to make this change on master and port it over.

ericniebler:
That's ok. The master branch is not long for the world.

PR #1023, opened 2019-01-31
make semiregular copy constructor be conditionally noexcept

h-2:
I got some broken-ness through changing to v1.0-beta which this little patch resolves. I assume the conditional noexcept was just forgotten?


CaseyCarter:
> I assume the conditional noexcept was just forgotten?

That does seem to be the case.

h-2:
Thanks for the quick merge!

PR #1024, opened 2019-01-31
MSVC 1.0 work

CaseyCarter:
Workarounds for a few MSVC bugs, and various cleanup of things I noticed along the way. (See the individual commit logs for details.)


include/concepts/concepts.hpp
```diff
@@ -712,6 +735,30 @@ namespace concepts
         struct Nil
         {};
 
+#ifdef CPP_WORKAROUND_MSVC_779763
+        enum class xNil {};
+
+        struct CPP_true_t {
```
ericniebler:
```suggestion
        struct CPP_true_t
        {
```

include/range/v3/range/conversion.hpp
```diff
@@ -146,7 +146,7 @@ namespace ranges
     /// \cond
     namespace _to_
     {
-    /// \endond
+    /// \endcond
```
ericniebler: 👍

include/range/v3/view/interface.hpp
```diff
@@ -161,16 +161,16 @@ namespace ranges
         }
         /// If the size of the range is known at compile-time and finite,
         /// return it.
-        template<bool True = true>
+        template<bool True = true, int = 42>
```
ericniebler: Por què?

CaseyCarter: [[over.load]/2.2](http://eel.is/c++draft/over.load#2.2):

> Likewise, member function template declarations with the same name, the same parameter-type-list, and the same template parameter lists cannot be overloaded if any of them is a static member function template declaration.

See? MSVC is *more* conforming than clang and GCC. (Despite that the IS should be fixed to agree with what clang and GCC do here, and to allow for differing constraints)

CaseyCarter:
> Thanks for doing this work! Does the v1.0-beta branch work with MSVC now?

HAHAHAHAhahahahaaaaa... 😢. No. This is what I could workaround for now, after which every test was firing the exact same assert in the compiler trying to convert `Eval`s (and compositions thereof) to `bool` in `enable_if_t` arguments, which is probably impossible to workaround while preserving concepts syntax and short-circuiting behavior. I figured I'd push what I have and shout at the compiler team, who has already sent a fix to code review.

Now everything is failing for a different reason that I need to reduce. And the cycle continues.

CaseyCarter:
> Now everything is failing for a different reason that I need to reduce. And the cycle continues.

I spoke too soon! 13 tests are passing.

include/range/v3/view/interface.hpp
ericniebler: Wow. That's a real thing? 😕

ericniebler:
> which is probably impossible to workaround while preserving concepts syntax and short-circuiting behavior.

You probably already know this, but the CPP pseudo-concepts don't really do short-circuiting, unless you use the pseudo-concepts defined in the `defer::` namespace.

CaseyCarter:
> > which is probably impossible to workaround while preserving concepts syntax and short-circuiting behavior.
> 
> You probably already know this, but the CPP pseudo-concepts don't really do short-circuiting, unless you use the pseudo-concepts defined in the `defer::` namespace.

Yes, which reminds me that I saw a `!defer::SomeConcept</**/> && defer::OtherConcept</**/>` somewhere that needs to be fixed.

CaseyCarter:
> I saw a `!defer::SomeConcept</**/> && defer::OtherConcept</**/>` somewhere that needs to be fixed.

Wrong - the `Eval`s overload `!` specifically to allow this.

include/range/v3/view/interface.hpp
CaseyCarter: Yeah. We used to reject *any* static + non-static overloads - regardless of template-ness. I tried to convince the compiler guys to implement more relaxed rules that would allow overloads with differing constraints, but they chose to instead implement the *exact* rule as specified in [over.load]/2.2.

PR #1025, opened 2019-01-31
[WIP] iota_view experiments

ericniebler:
This implements `iota_view` from the WD with a few twists:

* `iter_difference_t<integral-type>` is now _wider_ (where possible) so as to be able to fully represent the domain.
* `iota_view<A,B>::end()` returns `unreachable_sentinel_t` if `B` is `unreachable_sentinel_t`.
* `iota_view<A,B>`'s sentinel satisfies `SizedSentinel` if `SizedSentinel<B,A>`.


include/range/v3/view/iota.hpp
```diff
@@ -19,21 +19,20 @@
...
-#include <range/v3/iterator/default_sentinel.hpp>
+#include <range/v3/utility/static_const.hpp>
 
 namespace ranges
 {
     CPP_def
     (
         template(typename I)
-        concept BidirectionalIncrementable,
+        concept Decrementable_,
```
CaseyCarter: We really should have named this `Bincrementable`.

include/range/v3/view/iota.hpp
```diff
@@ -44,399 +43,359 @@ namespace ranges
...
-    struct iota_view
-      : view_facade<iota_view<From, To>, finite>
-    {
-    private:
-        friend range_access;
-        using difference_type_ = detail::iota_difference_t<From>;
+        private:
+            friend range_access;
+            From from_ = From();
+            To to_ = To();
```
CaseyCarter: `RANGES_UNIQUE_ADDRESS` for consistency with line 145.

CaseyCarter: (I'd put it on the `From`s as well, but you know me and EBO.)

include/range/v3/view/iota.hpp
```diff
@@ -44,399 +43,359 @@ namespace ranges
...
+            (!Integral<From> || !Integral<To> ||
+                std::is_signed<From>::value == std::is_signed<To>::value))
+    closed_iota_view(From, To) -> closed_iota_view<From, To>;
+#endif
+
+    template<typename From, typename To /* = unreachable_sentinel_t*/>
+    struct RANGES_EMPTY_BASES iota_view
+      : view_facade<
+            iota_view<From, To>,
+            Same<To, unreachable_sentinel_t> ? infinite : finite>
```
CaseyCarter: `finite` here should probably be `unknown`, or `is_integeral_v<From> && is_integeral_v<To> ? finite : unknown`.

include/range/v3/view/iota.hpp
```diff
@@ -44,399 +43,359 @@ namespace ranges
...
+            CPP_member
+            auto advance(iter_difference_t<From> n) -> CPP_ret(void)(
+                requires Advanceable_<From>)
+            {
+                from_ += n;
+            }
+            CPP_member
+            auto distance_to(cursor const &that) const ->
+                CPP_ret(iter_difference_t<From>)(
+                    requires Advanceable_<From>)
```
CaseyCarter: Not strictly to spec: the relational operators of the resulting `basic_iterator` effectively require *`Advanceable`*, - since they're implemented in terms of `distance_to` - whereas in the working draft those relational operators only require `StrictTotallyOrdered<From>`.

include/range/v3/view/iota.hpp
```diff
@@ -445,45 +404,29 @@ namespace ranges
                 return iota_view<From>{std::move(value)};
             }
             template<typename From, typename To>
-            auto CPP_fun(operator())(From from, To to) (const
-                requires WeaklyIncrementable<From> && WeaklyEqualityComparableWith<From, To>)
+            auto operator()(From from, To to) const ->
+                CPP_ret(iota_view<From, To>)(
+                    requires WeaklyIncrementable<From> && Semiregular<To> &&
+                        WeaklyEqualityComparableWith<From, To> &&
```
CaseyCarter: Should we move `WeaklyEqualityComparableWith` into `detail`?

test/view/iota.cpp
```diff
@@ -33,62 +34,36 @@ struct Int
...
-
-    auto rng3 = view::closed_indices(from, to);
-    static_assert(RandomAccessRange<decltype(rng3)>, "");
-    static_assert(SizedRange<decltype(rng3)>, "");
-    models<RandomAccessViewConcept>(aux::copy(rng3));
-    models<BoundedViewConcept>(aux::copy(rng3));
-    CHECK(ranges::size(rng3) == std::size_t(6));
-    ::check_equal(rng3, {U(max - 2), U(max - 1), max, U(0), U(1), U(2)});
-}
+template<class> struct undef;
```
CaseyCarter: Put this in `simple_test.hpp` so we don't have to constantly define it (and clean it out of PRs).

ericniebler:
Thanks for the feedback. An improvement over the WD: yea or nay?

An interesting side-effect of this change is that the `difference_type` for `iota_view<char>` changes from `int` to `short`. &lt;shrug&gt;

ericniebler:
If we go the route of promoting the difference type of integral incrementables, we should also take the additional step of proving integer-specific implementations of `distance` and `advance` that do the right thing.

ericniebler:
@CaseyCarter I'm feeling really indecisive on this one. For arguments sake, consider this code:

```c++
auto ints = view::ints(INT_MIN, INT_MAX);
assert( distance(ints.begin(), ints.end()) != -1 );
```

Before this diff, the assert fires. After this diff, it doesn't. That is the context.

Here are the options I see:

1. Do nothing. The above code is user error. They overflowed the difference type, invoked UB, and now their computer is on fire. Tough luck.
2. This diff. The difference type of `int` is `long`, the above code has defined behavior, and the distance of `ints` is reported correctly. (Had they used `std::intmax_t` instead of `int`, they incur the wrath of the UB gods; see (1).)
3. We go the full monty and support user-defined integer-like types for difference type. That involves at least:
  a. Defining an `IntegerLike` concept
  b. Changing `incrementable_traits` to promote integer difference types. Those can can make do with a built-in get one; others (like `std::[u]intmax_t`) get a library-defined integer-like type.
  c. Hooking `numeric_traits` for user-defined integers
  d. Changing `is_signed`, `make_[un]signed`, and friends for user-defined integers.
  e. Probably also extending `distance` and `advance` to support integer-like types.
4. We take a baby-step in the direction of (3) and define a single extended integer-like type in the library, called `std::diffmax_t` (&lt;/bikeshed&gt;) which has range `[-(uintmax_t)-1,(uintmax_t)-1]`. (Can be trivially implemented as a literal type with a `uintmax_t` and a `bool` for magnitude and sign.)
  a. A difference type is a signed built-in integer, or `std::diffmax_t`.
  b. Changing `incrementable_traits` to promote integer difference types. Those can can make do with a built-in get one; others (like `std::[u]intmax_t`) get `std::diffmax_t`.
  c. Define a specialization for `numeric_limits<diffmax_t>`.
  d. Survey STL implementations to see if any use `is_signed`/`make_[un]signed` on a difference type. Hopefully not.
  e. Probably also extending `distance` and `advance` to support integer-like types.

The above glosses over the existence of the _extended integer types_. If we require `incrementable_traits` to promote the difference type of integers, then we need magic hand-waving to make sure that the extended integer types are handled too.

Option (4) has the obvious problem that a difference type had better be `Incrementable` and as such, should have a difference type. But here we run into an infinite regress. I don't have a good solution.

Option (3) makes me super-nervous this late in the ballgame, and requires a big push that I probably have neither the time nor the energy for. Options (2) is a possibility, but it is not a complete solution, since we still have UB in the difference of `std::[u]intmax_t`. Option (4) would be a decent stop-gap if we could somehow avoid the infinite regress.

I think I'll stop now. Hopefully after reading this you have a good idea, because I don't.

ericniebler:
See the comment [here](https://github.com/ericniebler/stl2/issues/612#issuecomment-462149757) for a potential "deep fix" for this problem and others.

PR #1028, opened 2019-02-02
[WIP] SFINAE experiments

ericniebler:
Just a test to see how compile times fare after switching most signatures back to using SFINAE-in-tparam-list.


CaseyCarter:
Our old friend "GCC can't tolerate a *requires-clause* in the template-head followed by an *attribute-specifier*" is back. I tacked on a commit that passes `-DRANGES_DISABLE_DEPRECATED_WARNINGS` in `.travis.yml` so you can get data from GCC as well.

ericniebler:
Note that with this diff, I am fixing the travis config so that GCC does not use concepts unless `CONCEPTS=ON` is specified. Previously there was a bug that GCC always used concepts if the compiler supported them. That will skew the numbers to make GCC without concepts look slower.

CaseyCarter:
GCC8/Release/1z/concepts is notably *slower* than GCC8/Release/1z/no concepts (disclaimer: there are a great many uncontrolled variables in Travis builds, this result isn't statistically significant). This seems to corroborate our experience with cmcstl2 which suggests that subsumption is expensive, and `if constexpr` should be preferred for closed overload sets.

ericniebler:
Interestingly, this does appear to make the Travis build measurably slower: ~8h45m vs ~8h30m. I'm rerunning some builds to be sure, but I think I'll abandon this experiment and stick with SFINAE-in-return-type where possible.

PR #1029, opened 2019-02-02
More MSVC 1.0 work

CaseyCarter:
* [NFC] Remove unused `_concepts_int_` type alias
* Workaround MSVC's aversion to `explicit operator bool` in constant expressions.

This raises the bar from 6% tests passing to 32% (With VS2019 + not-yet-released bugfix)


ericniebler:
By now you have realized why the `_concepts_int_` typedef was there. :-)

PR #1034, opened 2019-02-04
Msvc beta

CaseyCarter:
More bits of WIP. 40% passing.

### DIFF
--- a/include/range/v3/view/zip_with.hpp
+++ b/include/range/v3/view/zip_with.hpp
@@ -206,7 +206,7 @@ namespace ranges
             {}
             template<bool Other>
             CPP_ctor(cursor)(cursor<Other> that)(
-                requires Const && (!Other))
+                requires (!Other && Const))
               : fun_(std::move(that.fun_)), its_(std::move(that.its_))
             {}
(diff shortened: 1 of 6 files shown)


include/range/v3/view/zip_with.hpp
```diff
@@ -206,7 +206,7 @@ namespace ranges
             {}
             template<bool Other>
             CPP_ctor(cursor)(cursor<Other> that)(
-                requires Const && (!Other))
+                requires (!Other) && Const)
```
CaseyCarter: Note: This is still ill-formed NDR when `Const` is `false` after being transformed into:
```c++
template<bool Other>
cursor(cursor<Other>, enable_if_t<(!Other) && Const, /**/> = {});
```
Since there is no valid specialization of this template for any choice of `Other`. The reordering here is specifically to trick MSVC, which diagnoses `false && dependent-stuff` but not `dependent-stuff && false`.

CaseyCarter: I should probably `CPP_template(bool foo = Const)` all of these converting constructors?

PR #1039, opened 2019-02-07
MOAR MSVC 1.0 work

CaseyCarter:
84% tests passing.

### DIFF
--- a/include/range/v3/view/transform.hpp
+++ b/include/range/v3/view/transform.hpp
@@ -166,13 +166,13 @@ namespace ranges
         constexpr /*c++14*/ auto CPP_fun(size)() (
             requires SizedRange<Rng>)
         {
-                return static_cast<range_size_type_t<Rng>>(ranges::size(this->base()));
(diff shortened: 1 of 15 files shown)


include/range/v3/view/transform.hpp
```diff
@@ -166,13 +166,13 @@ namespace ranges
         constexpr /*c++14*/ auto CPP_fun(size)() (
             requires SizedRange<Rng>)
         {
-                return static_cast<range_size_type_t<Rng>>(ranges::size(this->base()));
+                return ranges::size(this->base());
```
ericniebler: This looks suspiciously like a greater than 4 space indent to me. :-? Preexisting of course, but since you're changing that line anyway.... (Below, also.)

CaseyCarter:
I snuck in one more commit at the end there, which documents the "MSVC allows array rvalues to bind to lvalue references" bug and disables test cases that fail due to `std::move(array)` satisfying *`forwarding-range`*.

PR #1048, opened 2019-02-08
zip_with: Deal with underlying ranges with size types of differing si…

CaseyCarter:
…gnedness

Drive-by: "fix" `detail::min_` and `detail::max_`


CaseyCarter:
Note: the fix on v1.0 is better, drop the change in `size` when merging.

PR #1052, opened 2019-02-11
Improve logic controlling the enabling of testing.

dpacbach:
When range-v3 is being used as a standalone project (i.e., not a
subproject) then behavior should stay the same, in that the building
of tests/examples/etc. is enabled by default, but still settable
by the user using the same variables.

However, if range-v3 is a subproject of a parent then we will attempt
to respect the parent's settings regarding testing.  In that situation:

  * Testing/header checks are disabled by default.
  * Examples/docs/perf targets are disabled by default.
  * Enable testing/header checks if the parent has already enabled
    testing.

Effectively this causes the range-v3 subproject to mirror the testing
status of its parent, but while still allowing customization using
the usual variables.

This change was tested with CMake 3.13 in the following configurations:

  1) range-v3 is a standalone project, verified that tests/examples
     are built by default.
  2) range-v3 is a subproject of a parent with testing disabled.
     Verified that range-v3 tests are not built.
  3) rnage-v3 is a subproject of a parent with testing enabled.
     Verified that range-v3 tests are built and runnable as part
     of the global collection of tests.

For reference: https://cmake.org/cmake/help/v3.0/module/CMakeDependentOption.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,13 @@
 cmake_minimum_required(VERSION 3.6)
 get_directory_property(is_subproject PARENT_DIRECTORY)
 
+# Having both `is_subproject` and `is_standalone` improves readability.
+if(NOT is_subproject)
+    set(is_standalone YES)
+else()
+    set(is_standalone NO)
+endif()
+
 project(Range-v3 CXX)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
@@ -29,7 +36,6 @@ function(rv3_add_test TESTNAME EXENAME FIRSTSOURCE)
   add_test(${TESTNAME} ${EXENAME})
 endfunction(rv3_add_test)
 
-include(CTest) # invokes enable_testing() and defines BUILD_TESTING variable, defaulting to ON
 include(TestHeaders)
 include(ranges_options)
 include(ranges_env)
@@ -40,6 +46,9 @@ if(RANGE_V3_DOCS)
 endif()
 
 if(RANGE_V3_TESTS)
+  # The CTest module invokes enable_testing() and defines the
+  # BUILD_TESTING variable, defaulting to ON.
+  include(CTest)
   add_subdirectory(test)
 endif()
 
--- a/cmake/ranges_options.cmake
+++ b/cmake/ranges_options.cmake
@@ -30,22 +30,22 @@ endif()
 
 CMAKE_DEPENDENT_OPTION(RANGE_V3_TESTS
   "Build the Range-v3 tests and integrate with ctest"
-  "${BUILD_TESTING}" "${is_subproject}" OFF)
+  ON "${is_standalone}" "${BUILD_TESTING}")
 
 CMAKE_DEPENDENT_OPTION(RANGE_V3_HEADER_CHECKS
   "Build the Range-v3 header checks and integrate with ctest"
-  "${BUILD_TESTING}" "${is_subproject}" OFF)
+  ON "${is_standalone}" "${BUILD_TESTING}")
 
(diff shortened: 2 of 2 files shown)


dpacbach:
Question: is the build system supposed to build examples by default when in standlone build mode?  My changes should build them, but I wasn't totally clear on what the intention was.

ericniebler:
Looks good to me, thanks.

johelegp:
Now it seems that I can't have my own tests built without Range v3 building its own. Setting the cache variable `RANGE_V3_TESTS` to `OFF` isn't working. Any suggestion, @dpacbach?

The documentation you linked to suggests that setting `RANGE_V3_TESTS` would require `is_standalone` to be `TRUE`, otherwise it will always default to `BUILD_TESTING` without the possibility of setting it.

This also makes me wonder if it was possible at all to have Range v3 build its tests when used as a subproject.

dpacbach:
What I do in my top-level CMake file is the following:

`add_subdirectory( extern/range-v3 EXCLUDE_FROM_ALL )`

I believe that excludes all targets from the subproject unless they are explicitly depended upon by the parent project.  In this case, since presumably your tests do not depend on the range-v3 tests, those would not be built.  If this doesn't work for you and you want to use the `RANGE_V3_TESTS` variable then see below.

From my reading of the CMake code I think what you are seeing is expected:
  1) You run CMake in your top-level project with testing enabled
  2) The range-v3 cmake script sees that `is_standalone` is false and therefore sets
      `RANGE_V3_TESTS` to `${BUILD_TESTING}` which is `ON`.
  3) Then in range-v3 top-level CMake script it adds the testing folder, causing tests
       to be built even if you later set the the cached variable to OFF.

So perhaps you can try to clean the CMake cache, and run CMake specifying the:

`<cmake-cmd>  ... -DRANGE_V3_TESTS=OFF  ...`

on the commandline.  Hopefully this will intercept #2 above and force that variable to `OFF` so that the tests will not be included.

dpacbach:
Another solution would potentially be to have range-v3 tests off by default when `is_standalone` is false (as opposed to defaulting to `${BUILD_TESTING}`) which I suppose would work for people:
```CMake
CMAKE_DEPENDENT_OPTION(RANGE_V3_TESTS
  "Build the Range-v3 tests and integrate with ctest"
  ON "${is_standalone}" OFF)
```
and would still allow overriding the value on the commandline on a clean cmake run.

johelegp:
Thank you. `EXCLUDE_FROM_ALL` does the job gracefully.

david8dixon:
@ johelegp check to see if the range-v3 tests are still registered if you do this:

```
include(CTest)
...
...
...
add_subdirectory(range-v3 EXCLUDE_FROM_ALL)
```

I suspect that they will be. We are trying to resolve this in another issue, but I think the best solution for now is:

```
CMAKE_DEPENDENT_OPTION(RANGE_V3_TESTS
  "Build the Range-v3 tests and integrate with ctest"
  ON "${is_standalone}" OFF)
```

From my perspective, I don't want subproject tests coupled to setting tests on/off in my composing project. In projects with many dependencies it becomes hard to manage where the first `include(CTest)` is used and `BUILD_TESTING` appears to propagate in ways I wouldn't expect (need to verify this).

In our projects, we do the following:

```
CMAKE_DEPENDENT_OPTION(my.tests
  "Build the tests and integrate with ctest"
  ON "BUILD_TESTING; NOT subproject" OFF)
```

johelegp:
> @ johelegp check to see if the range-v3 tests are still registered if you do this:
> [...]

They aren't. At least, they don't build. That said, I call `add_subdirectory(${range-v3_SOURCE_DIR} ${range-v3_BINARY_DIR} EXCLUDE_FROM_ALL)` before my own `include(CTest)`, but I'm using the `FetchContent` module.

PR #1059, opened 2019-02-13
MSVC 1.0 again

CaseyCarter:
99% passing; act.split, generator, and iter.iterator (specifically the deep integration tests) are failing.


include/range/v3/view/subrange.hpp
```diff
@@ -409,7 +409,7 @@ namespace ranges
         subrange<iterator_t<R>, sentinel_t<R>, subrange_kind::sized>;
 #endif
 
-    // in leu of deduction guides, use make_subrange
+    // in lieu of deduction guides, use make_subrange
```
CaseyCarter: Fun fact: "leu" is "the basic monetary unit of Romania, equal to 100 bani."

PR #1061, opened 2019-02-15
[CMake] Give all message(...) invocations explicit modes.

dpacbach:
For reference: https://cmake.org/cmake/help/latest/command/message.html

There were various invocations of the message() function without
explicit modes being specified.  This causes the message to go to
stderr (as does every other mode except for STATUS).

This commit makes all message() invocations have an explicit mode: most
of them get "STATUS" (which is akin to info-level logging and goes
to stdout) while the ones that had "[range-v3 warning]" were given
the "WARNING" mode so that they will continue going to stderr.

This should make the output of the range-v3 CMake scripts more
consistent and organized so that wrapper scripts/projects can more
easily distill warnings from standard messages.


cmake/ranges_diagnostics.cmake
```diff
@@ -40,5 +40,5 @@ if (RANGES_CXX_COMPILER_GCC)
 endif()
 
 if (RANGES_VERBOSE_BUILD)
-  message("[range-v3]: test C++ flags: ${CMAKE_CXX_FLAGS}")
+    message(STATUS "[range-v3]: test C++ flags: ${CMAKE_CXX_FLAGS}")
```
CaseyCarter:
```suggestion
  message(STATUS "[range-v3]: test C++ flags: ${CMAKE_CXX_FLAGS}")
```

CaseyCarter:
Note: We need to audit for other occurrences when merging this to v1.0-beta.

dpacbach:
are you asking me to submit a PR to merge this into v1.0-beta as well?

CaseyCarter:
> are you asking me to submit a PR to merge this into v1.0-beta as well?

No, I am not - that's a note to remind me that I should merge this to 1.0 before forgetting that there's extra work to do besides simply resolving merge conflicts.

CaseyCarter:
Oh, and thanks!

dpacbach:
ok, thank you

CaseyCarter:
(FWIW, there were no other occurrences of `message` on 1.0)

Issue #1015, opened 2019-01-28
No action for remove (without if)?


tower120:
Is there a particular reason, why there is no `action::remove` / `view::remove` ?

ericniebler:
No reason, short of a lack of time. Can I interest you in submitting a PR?

tower120:
I'm trying to implement `action::remove` as wrapper around `action::remove_if` .

I think, I'm bumped into the same thing as with views. Are actions **must** be Semiregular too?

Following wan't compile:
https://wandbox.org/permlink/2cqBMYc1lTAoJ5w2

```cpp
#include <iostream>
#include <vector>

#include <range/v3/action/remove_if.hpp>


struct Pred{    
    int i = 2;
        
    Pred(const Pred&) = delete;
    Pred(Pred&&) = default;
    Pred() = default;
    
    bool operator()(int other) const {
        return other == i;
    }
};

void test_value_capture(){
    using List = std::vector<int>;
    List list{};
    list |= ranges::action::remove_if(Pred());

    std::cout << list.empty();
}

int main()
{
    test_value_capture();
}
```

PR #1063, opened 2019-02-17
Added action::remove (#1015)

tower120:
Implementing https://github.com/ericniebler/range-v3/issues/1015


include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,82 @@
...
+        /// \addtogroup group-actions
+        /// @{
+        namespace action
+        {
+            struct remove_fn
+            {
+            private:
+                friend action_access;
+                template<typename V, typename P = ident, CONCEPT_REQUIRES_(!Range<V>())>
+                static auto bind(remove_fn remove, const V& value, P proj = P{})
```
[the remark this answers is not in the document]
CaseyCarter: Lifetime: rvalues can bind to `const V&`, which could leave `std::cref(value)` dangling (`auto foo = action::remove(string{"abc"}); container |= foo;`) I suggest taking `V &&value` and adding a `std::is_lvalue_reference<V>::value` term to the requirements.

include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+        /// \addtogroup group-actions
+        /// @{
+        namespace action
+        {
+            struct remove_fn
+            {
+            private:
+                friend action_access;
+                template<typename V, typename P = ident, CONCEPT_REQUIRES_(!Range<V>())>
```
CaseyCarter: `!Range<V>()` seems limiting here - I might want to use this action on a range of ranges (e.g. `vector<string>`). Could we split this into two overloads?
```c++
template<typename V>
static auto bind(remove_fn f, V const& v);
template<typename V, typename P,
    CONCEPT_REQUIRES_(!(Range<V>() && EqualityComparable<range_reference_t<V>, P>()))>
static auto bind(remove_fn f, V const& v, P proj);
```

include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,82 @@
...
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    std::bind(remove,
+                        std::placeholders::_1,
+                        std::cref(value),
+                        protect(std::move(proj))
+                    )
+                )
+
+                template<class Value>
```
CaseyCarter: range-v3 refuses to accept the self-evident truth that `class` is superior to `typename` for template type parameters:
```suggestion
                template<typename Value>
```

include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,82 @@
...
+                        return other_value == value;
+                    }
+                };
+            public:
+
+                template<typename Rng, typename Value, typename P = ident,
+                    CONCEPT_REQUIRES_(remove_if_fn::Concept<Rng, pred<Value>, P>())>
+                Rng operator()(Rng&& rng, const Value& value, P proj = P{}) const
+                {
+                    return remove_if_fn{}(
```
CaseyCarter: Why implement this in terms of `action::remove_if` - which necessitates the making a predicate - rather than implementing it in the style of `action::remove_if` (i.e., call `ranges::remove` and `action::erase`)?

test/action/remove.cpp
```diff
@@ -0,0 +1,57 @@
...
+//
+// Project home: https://github.com/ericniebler/range-v3
+
+#include <vector>
+#include <range/v3/action/remove.hpp>
+
+#include "../simple_test.hpp"
+#include "../test_utils.hpp"
+
+struct Data{
```
CaseyCarter:
```suggestion
struct Data
{
```

test/action/remove.cpp
```diff
@@ -0,0 +1,57 @@
...
+    int i;
+
+    Data() = default;
+    explicit Data(int i) : i(i) {}
+
+    Data(const Data&) = delete;
```
CaseyCarter: It seems anomalous to explicitly `delete` one of the copy operations and not the other - it makes me thing something weird is going on here. Could we either explicitly delete both, or allow both to be implicitly deleted (which is the effect of defaulting the move operations)?

test/action/remove.cpp
```diff
@@ -0,0 +1,57 @@
...
+    list.emplace_back(Data{4});
+
+
+    list |= action::remove(Data{2});
+    check_equal(list, {Data{1}, Data{3}, Data{4}});
+
+    list |= action::remove(3, &Data::i);
+    check_equal(list, {Data{1}, Data{4}});
+
+
```
CaseyCarter:
```suggestion
```

test/action/remove.cpp
```diff
@@ -0,0 +1,57 @@
...
+    bool operator==(const Data& other) const {
+        return other.i == i;
+    }
+    bool operator!=(const Data& other) const {
+        return other.i != i;
+    }
+};
+
+
```
CaseyCarter:
```suggestion
```

include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,82 @@
...
+            {
+            private:
+                friend action_access;
+                template<typename V, typename P = ident, CONCEPT_REQUIRES_(!Range<V>())>
+                static auto bind(remove_fn remove, const V& value, P proj = P{})
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    std::bind(remove,
+                        std::placeholders::_1,
+                        std::cref(value),
```
CaseyCarter: I'm vaguely uncomfortable with this "capture a reference" formulation, where range-v3 would typically require `value` to be `CopyConstructible` and store it in a `semiregular_t`.

@ericniebler, thoughts? Do you think it's cleaner to enable use of move-only types by capturing references or by relaxing the `Semiregular` requirement for actions?

tower120: No, you're right. 
I thought about actions as something immobile and applied immediately.
But since     
```cpp
const auto remove_data = action::remove(Data{2});
list |= remove_data;
```
is possible, it is incorrect to stored value as reference.

include/range/v3/action/remove.hpp
tower120: Somehow I was sure, there is no `ranges::remove` algorithm (only `remove_if`)... I stand corrected.

include/range/v3/action/remove.hpp
tower120: I will replace `std::cref` with `bind_forward`.

But somewhere internally, inside utility/functional.hpp binded function copied (passed by value). I suppose its somewhere in `pipeable::operator|(...)`

At least this prevents move-only types from being used as arguments for actions. _Besides `Semiregular` requirement for actions._

include/range/v3/action/remove.hpp
tower120:
```
EqualityComparable<range_reference_t<V>, P>()
```
It never compiles. Probably, compiler proceed processing code even in cases when V is not Range.

include/range/v3/action/remove.hpp
ericniebler: I find that C++'s addition of a `class` keyword in the 80's was largely pointless and a cowing to the OO religion of the day. `struct` was doing just fine and didn't need replacement. In range-v3 I voted with my feet (er, fingers) and type `c-l-a-s-s` hardly ever.

include/range/v3/action/remove.hpp
```diff
@@ -0,0 +1,94 @@
...
+        /// \addtogroup group-actions
+        /// @{
+        namespace action
+        {
+            struct remove_fn
+            {
+            private:
+                friend action_access;
+                template<typename V, typename P,
+                    CONCEPT_REQUIRES_(!(Range<V>() /*&& EqualityComparable<range_reference_t<P>, P>()*/))>
```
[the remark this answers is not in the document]
tower120: That's from @CaseyCarter original suggestion: https://github.com/ericniebler/range-v3/pull/1063#discussion_r257518621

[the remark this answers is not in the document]
ericniebler: That isn't what Casey suggested. He suggested `range_reference_t<V>` but you wrote `range_reference_t<P>`.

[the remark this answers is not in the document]
ericniebler: I have pushed a change to your branch to properly constrain the `remove_fn::bind` function.

tower120:
Read this https://github.com/ericniebler/range-v3/issues/1136 before applying .

ericniebler:
Please rebase this on master. I'm having a hard time reviewing this diff with all the other commits gumming it up.

Issue #1030, opened 2019-02-03
view::addressof


tower120:
There is `view::indirect` which makes references from pointers of all forms(dereference), but there is no reverse operation.
What about `view::addressof`?
Something as simple as : 
```
    const auto addressof = ranges::view::transform([](auto& element) -> auto* {
        return std::addressof(element);
    });

```

CaseyCarter:
This seems useful to me, although it needs some thought on what the constraints should be. Obviously we want to require the underlying range to have a reference type that's actually an lvalue reference, but there's also some subtlety around iterator category as well. We require that references obtained from `Forward` iterators must be valid while the iterator's range is valid, but there's no such requirement for single-pass iterators. Someone needs to investigate whether allowing `view::addressof` to accept single-pass ranges could lead to badness.

tower120:
At first: truly Single-pass range, will probably deference to value, or something like shared_ptr...
At second: if range deferences to lvalue reference, it is expected that we can actually do something with that reference. And it not becomes invalid immediately after read, and will live until some predictable point in time. And until that reference alive - it's pointer/address valid.

PR #1064, opened 2019-02-18
 implementing view::addressof (#1030) 

tower120:


include/range/v3/view/addressof.hpp
```diff
@@ -0,0 +1,72 @@
...
+                    template<class V>
+                    constexpr V *operator()(V &value) const noexcept
+                    {
+                        return std::addressof(value);
+                    }
+                    template<class V>
+                    constexpr V const *operator()(V const &value) const noexcept
+                    {
+                        return std::addressof(value);
+                    }
```
ericniebler: This second overload is unnecessary.

test/view/addressof.cpp
```diff
@@ -0,0 +1,36 @@
...
+using namespace ranges;
+
+void simple_test()
+{
+    std::vector<int> list = {1,2,3};
+
+    auto out = list | view::addressof;
+
+    check_equal(out, {&list[0], &list[1], &list[2]});
+}
```
ericniebler: Please also add tests for an input (not forward) range, and test also that you cannot pipe ranges with prvalue or xvalue reference types to `view::addressof`.

ericniebler:
Looks good. One last thing: please add `#include <range/v3/view/addressof.hpp>` to `<range/v3/view.hpp>`. Then I'll merge.

tower120:
>  One last thing: please add #include <range/v3/view/addressof.hpp> to <range/v3/view.hpp>. 

I'm sorry to ask, but - what for?
`addressof.hpp` includes `transform.hpp`. That's a ton of potentially unnecessary code for each `#include <range/v3/view.hpp>`.
Besides, `addressof.hpp` depends on `view.hpp`. This could easily cause circular dependency at some point (it does not now - due to deferred template instantiation, but such things fragile).

CaseyCarter:
> > One last thing: please add #include <range/v3/view/addressof.hpp> to <range/v3/view.hpp>.
> 
> I'm sorry to ask, but - what for?
> `addressof.hpp` includes `transform.hpp`. That's a ton of potentially unnecessary code for each `#include <range/v3/view.hpp>`.
> Besides, `addressof.hpp` depends on `view.hpp`. This could easily cause circular dependency at some point (it does not now - due to deferred template instantiation, but such things fragile).

`<range/v3/view.hpp>` is the header that includes the definitions of all the views; `<range/v3/view/view.hpp>` is the view machinery upon which all of the view definitions (including addressof) depend.

tower120:
@CaseyCarter Oh I see. Thanks.
@ericniebler please, update documentation as well. Online version is already differs from repository's.

PR #1068, opened 2019-02-25
testing is disabled when subproject

david8dixon:
Removed some comments and changed testing defaults in ranges_options.cmake from `BUILD_TESTING` to `OFF`. This ensures that when range-v3 is a subproject and consumed with `add_subdirectory` that the tests will not build or register with ctest.


dpacbach:
Thanks -- this is probably what I should I have done from the beginning now that I think about it.

ericniebler:
Thanks all!

PR #1073, opened 2019-03-05
delimit_view: don't decay arrays into pointers

CaseyCarter:


include/range/v3/view/delimit.hpp
```diff
@@ -82,7 +82,8 @@ namespace ranges
             public:
                 template<typename Rng, typename Val>
                 using Concept = meta::and_<
-                    Range<Rng>,
+                    InputRange<Rng>,
+                    SemiRegular<Val>,
                     EqualityComparable<Val, range_common_reference_t<Rng>>>;
```
ericniebler: Obviously pre-existing, but this constraint is highly suspect.

CaseyCarter: It wasn't really my intent to overhaul the requirements, so I was willing to turn a blind eye. Given that I largely overhauled the requirements anyway, and even duplicated this bogus constraint for the iterator overload on 120, I should fix it.

PR #1079, opened 2019-03-10
Added Conan.io version

cjdb:
Makes it possible to export range-v3. I think it's just BAU to upload to bintray from here.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -14,6 +14,7 @@ from conans import ConanFile, CMake
 
 class Rangev3Conan(ConanFile):
     name = "range-v3"
+    version = "v1.0-beta"
     license = "Boost Software License - Version 1.0 - August 17th, 2003"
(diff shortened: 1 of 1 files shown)


ericniebler:
I don't know enough to say what the impact of this change is. Can you say a little more about it?

cjdb:
Right now it's not possible to produce a Conan package because the version is missing. With this, anyone can create a package and upload it to an artifact repository. To the best of my knowledge, adding this doesn't break anything.

For example, if you were to upload a `v1.0-beta` package to Bintray.com, one could install it into their Conan-based project using the package name `range-v3/v1.0-beta@ericniebler/experimental`.

cjdb:
I should add that I was able to get a package uploaded to `range-v3/v1.0-beta@cjdb/stable` without any troubles.

PR #1080, opened 2019-03-11
mixin for adaptor (view_adaptor)

tower120:
This PR adds ability to use mixin in adaptor.
Interface is following (same as with cursor's mixin, but have additional base_iterator template parameter):

```cpp
class adaptor : public ::ranges::adaptor_base{
     template<class base_iter_t /* can be iterator_t<base_rng> or sentinel_t<base_rng> */>
     struct mixin : ranges::basic_adaptor_mixin<adaptor, base_iter_t>{
          mixin() = default;
          using ranges::basic_adaptor_mixin<adaptor, base_iter_t>::basic_adaptor_mixin;
     };
};
```

See `/test/view_adaptor.cpp` for example.

Accessing adaptor guts through `adaptor`, not `get`.

### DIFF
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -203,6 +203,8 @@ namespace ranges
         {
         private:
             friend range_access;
+            friend basic_adaptor_mixin<Adapt, BaseIter>;
+
             using base_t = detail::adaptor_value_type_<BaseIter, Adapt>;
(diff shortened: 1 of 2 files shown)


tower120:
This PR works only for cursor (not sentinel) adaptor. 
Closing this in favor of https://github.com/ericniebler/range-v3/pull/1086

PR #1086, opened 2019-03-12
mixin for adaptor (view_adaptor) v2

tower120:
This PR adds ability to use mixin in adaptor.
Interface:

```cpp
class adaptor : public ::ranges::adaptor_base
{
     template<class base_mixin>
     struct mixin : base_mixin
     {
          mixin() = default;
          using base_mixin::base_mixin;

          decltype(auto) base_value() const
          {
              return *base.value();
          }

          int get_i() const
          {
              return this->adaptor().i;
          }
     };

     int i = 100;
};
```

`adaptor()` is only visible from inside mixin.
`base()` - visible everywhere.

Work for both - cursor and sentinel adaptor versions.

### DIFF
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -203,6 +203,8 @@ namespace ranges
         {
         private:
             friend range_access;
+            friend basic_adaptor_mixin<Adapt, BaseIter>;
+
             using base_t = detail::adaptor_value_type_<BaseIter, Adapt>;
             using single_pass = meta::or_<
                 range_access::single_pass_t<Adapt>,
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -260,7 +260,7 @@ namespace ranges
             static meta::id<typename Adapt_::template mixin<basic_adaptor_mixin>> basic_adaptor_mixin_2_(int);
 
             struct basic_adaptor_mixin_
-                : decltype(basic_adaptor_mixin_2_<Adapt>(42))
+              : decltype(basic_adaptor_mixin_2_<Adapt>(42))
             {};
 
             using mixin = meta::_t<basic_adaptor_mixin_>;
--- a/include/range/v3/view_adaptor.hpp
+++ b/include/range/v3/view_adaptor.hpp
@@ -213,32 +213,6 @@ namespace ranges
             using detail::adaptor_sentinel_<BaseSent, Adapt>::adaptor_sentinel_;
         };
 
-        template<typename Adapt, typename BaseIter>
-        struct basic_adaptor_mixin
-            : basic_mixin<adaptor_cursor<BaseIter, Adapt>>
-        {
-            basic_adaptor_mixin() = default;
-            using basic_mixin<adaptor_cursor<BaseIter, Adapt>>::basic_mixin;
-
-            // All iterators into adapted ranges have a base() member for fetching
-            // the underlying iterator.
-            BaseIter base() const
-            {
-                return this->get().first();
-            }
-
-        protected:
-            // Adaptor accessor
-            Adapt& adaptor()
-            {
-                return this->get().second();
-            }
-            const Adapt& adaptor() const
-            {
-                return this->get().second();
-            }
-        };
-
         // Build a cursor out of an iterator into the adapted range, and an
         // adaptor that customizes behavior.
         template<typename BaseIter, typename Adapt>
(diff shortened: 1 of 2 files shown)


include/range/v3/view_adaptor.hpp
```diff
@@ -179,6 +179,38 @@ namespace ranges
...
+            {
+                return second();
+            }
+            const Adapt& adaptor() const
+            {
+                return second();
+            }
+        };
+
+        namespace detail {
```
ericniebler:
```suggestion
        namespace detail
        {
```

include/range/v3/view_adaptor.hpp
```diff
@@ -179,6 +179,38 @@ namespace ranges
...
+            template<typename BaseSent, typename Adapt>
+            meta::id<base_adaptor_sentinel<BaseSent, Adapt>> base_adaptor_sentinel_2_(long);
+
+            template<typename BaseSent, typename Adapt>
+            meta::id<typename Adapt::template mixin<base_adaptor_sentinel<BaseSent, Adapt>>> base_adaptor_sentinel_2_(int);
+
+            template<typename BaseSent, typename Adapt>
+            struct base_adaptor_sentinel_
+                : decltype(base_adaptor_sentinel_2_<BaseSent, Adapt>(42))
```
ericniebler:
```suggestion
              : decltype(base_adaptor_sentinel_2_<BaseSent, Adapt>(42))
```

include/range/v3/view_adaptor.hpp
```diff
@@ -179,6 +179,38 @@ namespace ranges
...
+            {};
+
+            template<typename BaseSent, typename Adapt>
+            using adaptor_sentinel_ = meta::_t<base_adaptor_sentinel_<BaseSent, Adapt>>;
+        }
+
+        template<typename BaseSent, typename Adapt>
+        struct adaptor_sentinel : detail::adaptor_sentinel_<BaseSent, Adapt>{
```
ericniebler:
```suggestion
        struct adaptor_sentinel
          : detail::adaptor_sentinel_<BaseSent, Adapt>
        {
```

include/range/v3/view_adaptor.hpp
```diff
@@ -193,19 +225,43 @@ namespace ranges
             using single_pass = meta::or_<
                 range_access::single_pass_t<Adapt>,
                 SinglePass<BaseIter>>;
-            struct mixin
-              : basic_mixin<adaptor_cursor>
+
+            struct basic_adaptor_mixin
+                : basic_mixin<adaptor_cursor<BaseIter, Adapt>>
```
ericniebler:
```suggestion
              : basic_mixin<adaptor_cursor>
```

include/range/v3/view_adaptor.hpp
```diff
@@ -193,19 +225,43 @@ namespace ranges
...
+                }
             };
 
+            template<typename Adapt_>
+            static meta::id<basic_adaptor_mixin> basic_adaptor_mixin_2_(long);
+            template<typename Adapt_>
+            static meta::id<typename Adapt_::template mixin<basic_adaptor_mixin>> basic_adaptor_mixin_2_(int);
+
+            struct basic_adaptor_mixin_
+                : decltype(basic_adaptor_mixin_2_<Adapt>(42))
```
ericniebler:
```suggestion
              : decltype(basic_adaptor_mixin_2_<Adapt>(42))
```

test/view_adaptor.cpp
```diff
@@ -81,6 +98,34 @@ struct my_delimited_range
         ranges::delimit_view<ranges::istream_range<int>, int>>
 {
     using view_adaptor::view_adaptor;
+
+
```
ericniebler:
```suggestion
```

test/view_adaptor.cpp
```diff
@@ -81,6 +98,34 @@ struct my_delimited_range
...
+            }
+        };
+
+        int t = 20;
+    };
+    adaptor begin_adaptor() const
+    {
+        return {};
+    }
+    adaptor end_adaptor() const {
```
ericniebler:
```suggestion
    adaptor end_adaptor() const
    {
```

tower120:
There is some inconsistency between adaptor mixin and iterator mixin. 
Iterator mixin use `get()` to access to iterator, while (for technical reasons) adaptor call it `adaptor()`.

I think now I can make it `get()` everywhere. But this require small changes to `basic_iterator` implementation. I'll make new separate PR to show you.

tower120:
Here it is https://github.com/ericniebler/range-v3/pull/1099 . It seems working. Look if that change in `basic_iterator` does not break anything conceptually.

include/range/v3/view_adaptor.hpp
tower120: Context dependent indent? 3 spaces, not 4, in class parents list?

tower120:
Why not this version https://github.com/ericniebler/range-v3/pull/1099 ?

tower120:
May I ask why you preferred this mixin interface, instead of https://github.com/ericniebler/range-v3/pull/1099 ?

PR #1091, opened 2019-03-18
view::transform iterator size optimisation.

tower120:
Currently, each `view::transform` in composition chain increase iterator size on 1 ptr:
```cpp
auto t1 = vec | view::transform(Op{});
auto t2 = t1  | view::transform(Op{});
auto t3 = t2  | view::transform(Op{});

assert( sizeof(t1.begin) == 16);
assert( sizeof(t2.begin) == 24);
assert( sizeof(t3.begin) == 32);
```
This is **too** wasteful. This may be forgivable for views, but not for iterators itself - since they copied/constructed in tight loops, and designed to be small.

This PR eliminates size grow of iterators of `transform`s with stateless functors (empty sized + default constructible):
```cpp
assert( sizeof(t1.begin) == 8);
assert( sizeof(t2.begin) == 8);
assert( sizeof(t3.begin) == 8);
```
---
Pay attention, that empty base optimization / [[no_unique_address]] is not used here. This is due to the fact, that only distinct types may occupy the same "memory location" (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88865). Which means that with EBO we'd have this:
```cpp
auto t1 = vec | view::transform(Op{});   // 8 - ok
auto t2 = t1  | view::transform(Op{});   // 16 - growing
```
At least I got that in [my version with EBO](https://github.com/tower120/range-v3/tree/view_transform_size_optimisation_w_ebo) (ranges::box) [see diff](https://github.com/tower120/range-v3/compare/view_transform_size_optimisation_w_ebo..view_transform_size_optimisation?diff=split)

If you know better, and EBO should work here - please let me know.
Another example with "non-working" size optimization https://wandbox.org/permlink/xrVxwlq4JsLG9abz

UPDATE

That's what happens under the hood:
```cpp
struct payload{ void* v;};
struct base_adaptor{};
struct A: base_adaptor, private box<Op, Op> {};    // without base_adaptor - all ok
struct B: base_adaptor, private box<Op2, Op2> {};  // or if replace box with any other type

using P  = compressed_pair<A, payload>;    // sizeof(P) == 8
using P2 = compressed_pair<B, P>;          // sizeof(P2) == 16
```
It seems that at the end - `compressed_pair` does not play nicely with the `box` (being box itself)... 
Linked issue https://github.com/ericniebler/range-v3/issues/1093

---

The same optimization must be done to the view.

### DIFF
--- a/include/range/v3/view/transform.hpp
+++ b/include/range/v3/view/transform.hpp
@@ -65,36 +65,87 @@ namespace ranges
             using use_sentinel_t =
                 meta::or_<meta::not_<BoundedRange<Rng>>, SinglePass<iterator_t<Rng>>>;
 
+            struct adaptor_fun;
+
+            // Do not use empty base class optimisation here. Return Fun{} each time.
+            // Stateless lambda is default_constructible since c++20.
+            // Empty base class optimisation will work only for distinct types.
+            // Some discussion/examples :
+            // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89745 ; https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88865
+            // This means, that vec | transform(Fun{}) | transform(Fun{}) can not be optimised with EBO.
+            template<class fun_ref_>
+            struct adaptor_fun<true, fun_ref_>
+            {
+            public:
+                using fun_t  = Fun;
+
+                fun_t fun_() const noexcept
+                {
+                    return Fun();
+                }
+
+                adaptor_fun() = default;
+                adaptor_fun(fun_t&&) noexcept {}
+            };
+
+            template<class fun_ref_>
+            struct adaptor_fun<false, fun_ref_>
+            {
+            public:
+                using fun_t  = fun_ref_;
+            private:
+                fun_t fun;
+            public:
+                const fun_t& fun_() const noexcept
+                {
+                    return fun;
+                }
+
+                adaptor_fun() = default;
+                adaptor_fun(fun_t&& fun) noexcept : fun(std::move(fun)) {}
+            };
+
             template<bool IsConst>
-            struct adaptor : adaptor_base
+            using adaptor_fun_ = adaptor_fun<
+                std::is_empty<Fun>::value && std::is_default_constructible<Fun>::value,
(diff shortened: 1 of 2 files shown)


ericniebler:
> This may be forgivable for views, but not for iterators itself - since they copied/constructed in tight loops, and designed to be small.

Is the concern that copying the extra empty space is too expensive at runtime? I would need to see proof of that.

Or is the concern that iterators should be small because large iterators are causing memory pressure? Is this a problem you have experienced in practice, or is it hypothetical?

tower120:
> Is the concern that copying the extra empty space is too expensive at runtime?

Yes this is my concern.
`view::transform` iterator currently holds pointer to the view. It's not empty-sized. 

> I would need to see proof of that.

https://github.com/tower120/big_iterators_benchmark

> Is this a problem you have experienced in practice, or is it hypothetical?

I found this, when I measured size impact of my views on iterators.
This [view chain](https://github.com/tower120/range_tagged_view#all-together) iterator has size 156 bytes, if memory serves. Common middle average view iterator is around 48-64 bytes.

Iterators copied all around in algorithms. They treated as cheap pointers.

tower120:
I think more general solution is possible for most view iterators. Most iterators store pointer to the view. 
If there would only one pointer for all chain - it would cut size tremendously.

Views and view iterators (implemented with view_adaptor) basically formed as chain of inheritance (well, not exactly, but). By knowing address of The Base View it should be possible to get address of required view. All view chain is known at compile time, and transformations needed to get from Base -> Required view can be passed to iterators (it could be template tuple). So only the very first one (e.g. view::all) iterator would need to store pointer to the base view, all other could calculate pointer to their respective view, by calculating it.

ericniebler:
I can't tell if your big_iterators_benchmark is measuring the cost of big iterators or whether it is measuring the compiler's ability to inline. It also doesn't show that this diff improves the situation.

I'm super-not excited about taking this special-case hack for the transform view. I'm closing this. I'm open to considering a more general fix that keeps iterators of adapted views small, but I'm still not convinced this is a real problem that needs solving. I'd wager that "most people" only apply one or two view adaptors in real code.

PR #1095, opened 2019-03-20
[FIX-#1092] view::take has problems with GCC+ConceptsTS

### LINKED ISSUE #1092: small bug with view::take
the following tiny example does not work with 1.0-beta branch:

```cpp
#include <list>
#include <iostream>

#include <range/v3/view/take.hpp>
#include <range/v3/view/drop.hpp>

int main()
{
    std::list l{1, 2, 3, 4, 5, 6, 7, 8, 9, 0};

    auto v = l | ranges::view::drop(1) | ranges::view::take(10)
               | ranges::view::drop(1) | ranges::view::take(10);
    std::cout << *v.begin();
}
```

```
g++8 -std=c++17 -fconcepts -I ~/devel/seqan3/submodules/range-v3/include test.cpp
In file included from test.cpp:4:
/home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/view/take.hpp: In instantiation of 'constexpr auto ranges::take_view<Rng>::begin() requires !(simple_view<Rng>)() [with Rng = ranges::drop_view<ranges::take_view<ranges::drop_view<ranges::ref_view<std::__cxx11::list<int, std::allocator<int> > > > > >]':
test.cpp:13:27:   required from here
/home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/view/take.hpp:155:30: error: no matching function for call to 'ranges::take_view<ranges::drop_view<ranges::take_view<ranges::drop_view<ranges::ref_view<std::__cxx11::list<int, std::allocator<int> > > > > > >::size()'
                     auto s = static_cast<range_difference_t<Rng>>(size());
                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/range/concepts.hpp:21,
                 from /home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/view/take.hpp:19,
                 from test.cpp:4:
/home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/view/take.hpp:233:32: note: candidate: 'constexpr auto ranges::take_view<Rng>::size() requires  SizedRange<Rng> [with Rng = ranges::drop_view<ranges::take_view<ranges::drop_view<ranges::ref_view<std::__cxx11::list<int, std::allocator<int> > > > > >]'
         constexpr auto CPP_fun(size)() (requires SizedRange<Rng>)
                                ^~~~
/home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/concepts/concepts.hpp:496:20: note: in definition of macro 'CPP_fun'
 #define CPP_fun(X) X CPP_FUN_IMPL_1_
                    ^
/home/mi/h4nn3s/devel/seqan3/submodules/range-v3/include/range/v3/view/take.hpp:233:32: note:   constraints not satisfied
         constexpr auto CPP_fun(size)() (requires SizedRange<Rng>)
```

The peculiar thing is that take.hpp:155 is inside a `if constexpr(SizedRange<Rng>)`, block, but it claims it cannot use either of the size overloads because the requirement `SizedRange<Rng>` is not met. So it might be a compiler issue, too.

Changing the line to use the free size() function on the base object "fixes it", but I still don't know why the direct invocation of the member fails:
```diff
-                    auto s = static_cast<range_difference_t<Rng>>(size());
+                    auto s = static_cast<range_difference_t<Rng>>(ranges::size(base_));
```
(note that this is not a real fix, as the it does not contain the std::min-logic of the member function).

h-2:
This fixes #1092 for me. The workaround would always work, but I have just activated it for ConceptsTS, because I don't know what your general policy for platform specific workarounds is.

### DIFF
--- a/include/range/v3/view/take.hpp
+++ b/include/range/v3/view/take.hpp
@@ -164,7 +164,12 @@ namespace ranges
                     return ranges::begin(base_);
                 else
                 {
+#if defined(__cpp_concepts) && __cpp_concepts <= 201507 // cannot always delegate to size() member on GCC with ConceptsTS
+                    auto s = ranges::min(static_cast<range_difference_t<Rng>>(count_),
+                                         static_cast<range_difference_t<Rng>>(ranges::size(base_)));
+#else
                     auto s = static_cast<range_difference_t<Rng>>(size());
+#endif
                     return make_counted_iterator(ranges::begin(base_), s);
                 }
             else


ericniebler:
Thanks!

Issue #1093, opened 2019-03-19
compressed_pair does not play nicely with box (size grows)


tower120:
https://wandbox.org/permlink/gPLp7Z3znfM7ZIbR
```cpp
struct Op{};
struct Op2{};    
    
struct payload{ void* v;};
struct base_adaptor{};
    
struct A: base_adaptor, private box<Op, Op> {};    // without base_adaptor - all ok
struct B: base_adaptor, private box<Op2, Op2> {};  // or if replace box with any other type

using P  = compressed_pair<A, payload>;    // sizeof(P) == 8
using P2 = compressed_pair<B, P>;          // sizeof(P2) == 16    
```
Replacing box with 
```cpp
template<class T, class Tag>
class box2 : T{};
```
Solves the problem. 
Removing multiple inheritance solves the problem too.

tower120:
The problem comes down to:
```cpp
    struct base_adaptor{};    struct Op{};     struct Op2{};
    struct A: base_adaptor, box<Op, Op> {};
    struct B: base_adaptor, box<Op2, Op2> {};

    struct P2 : box<A, A>
              , box<B, B>
    {}; // sizeof(P2) == 2
```
If box use in "coalesce" mode, problem solves:
```cpp
    struct base_adaptor{};    struct Op{};     struct Op2{};
    struct A: base_adaptor, box<Op, Op> {};
    struct B: base_adaptor, box<Op2, Op2> {};

    struct P2 : box<A, A, detail::box_compress::coalesce>
              , box<B, B, detail::box_compress::coalesce>
    {}; // sizeof(P2) == 1
```
Which comes down to the fact that EBO is allowed only for distinct types:
```cpp
    struct base_adaptor{};
    struct A: base_adaptor{};
    struct B: base_adaptor{};  // with base_adaptor2 EBO would work

    struct P2 : box<A, A, detail::box_compress::ebo>
              , box<B, B, detail::box_compress::ebo>
    {}; // sizeof(P2) == 2
```

I don't know is it safe to use "coalesce" box mode in general. 
But if so - maybe use it by default? This would solve the problem with compressed_pair/tuple.

_By the way - for default_constructible types there is probably one more box mode -> return T() on `get()`._

tower120:
Hmmm... - what now preventing "coalesce" box mode, is that this type is not trivial (!) :
```cpp
struct A: base_adaptor, box<Op> {};
```
Because `box<Op>` is not trivial.
[...]

PR #1096, opened 2019-03-20
make box trivially_default_constructible

tower120:
This PR makes box trivially_default_constructible if possible.
Fix https://github.com/ericniebler/range-v3/issues/1093


tower120:
Argh. It's just too much code for such a simple thing.

Probably it would be more reasonable just to remove `std::is_trivial<T>` from `box_compress::coalesce` requirements.

ericniebler:
IIRC @CaseyCarter is the one that added the `coalesce` strategy to `box`. Maybe he could cast an eye over this.

CaseyCarter:
Ack, I $&#@$@-ed up the branch with a bunch of commits from `master` trying to add bda19f7. Sorry about that.

In any case: I've removed all of the `box_ebo_` complexity: trivial default construction isn't a design goal for `box`, since it must value-initialize its `Element`. The only change that's needed for #1093 is the defaulted default constructor for the `coalesce` specialization of `box` (I have no idea why I didn't default it in the first place).

tower120:
@CaseyCarter Can trivial-constructivity requirement for `coalesce` be lifted too?

```cpp
    struct Op {
        Op() noexcept(some_condition) {};   // !!! this will prevent coalesce strategy now.
    };
    struct Op2 {};

    struct payload { void* v; };
    struct base_adaptor {};

    struct RANGES_EMPTY_BASES A : base_adaptor, private box<Op, A> {};
    struct RANGES_EMPTY_BASES B : base_adaptor, private box<Op2, B> {};

    using P  = compressed_pair<A, payload>;
    using P2 = compressed_pair<B, P>;

    // checks will fail
    CHECK(sizeof(P) == sizeof(payload));
    CHECK(sizeof(P2) == sizeof(P));
```

CaseyCarter:
> @CaseyCarter Can trivial-constructivity requirement for `coalesce` be lifted too?

That would be problematic for a couple of reasons: 
1. An empty type with non-trivial initialization is likely engaging in some kind of funny business that would break if we coalesce multiple instances into a single instance.
2. We *really* don't want to invisibly introduce the Static Initialization Order fiasco into a program by defining a static storage duration variable with dynamic initialization. I won't assert that there's no way to determine that a type has `constexpr` default-initialization, but I can't think of one of the top of my head.

Issue #1094, opened 2019-03-19
Can't avoid __builtin_readcyclecounter


mikezackles:
When building for aarch64 with the Android NDK r19c version of clang, `__builtin_readcyclecounter` fails with an illegal instruction at runtime:
`mrs x0, pmccntr_e10`

I was going to file a bug upstream, but I noticed this from the clang docs:
"Query for this feature with __has_builtin(__builtin_readcyclecounter). Note that even if present, its use may depend on run-time privilege or other OS controlled state."

So it seems like it might be a bug to assume that this is going to work if `__has_builtin` succeeds.

Perhaps it would be sufficient to allow the user to define RANGES_CPU_ENTROPY manually to disable this?

CaseyCarter:
I'm going to burninate the "entropy-gathering" code from `utility/random.hpp`; it's not worth the bother to try to provide non-deterministic seeding for platforms with a broken `std::random_device`.

> Perhaps it would be sufficient to allow the user to define `RANGES_CPU_ENTROPY` manually to disable this?

I have absolutely no idea what I was thinking when I didn't put the `__builtin_readcyclecounter` detection inside `#ifndef RANGES_CPU_ENTROPY`. 🤷‍♂️ I suppose it won't matter after it's gone.

mikezackles:
:dragon:

CaseyCarter:
Closed by #1097.

mikezackles:
Many thanks for the quick fix!

CaseyCarter:
> Many thanks for the quick fix!

It wasn't that quick, considering that I said I was going to make this change in #769 in January of 2018. But you're welcome ;)

PR #1097, opened 2019-03-20
Burninate the "local entropy" machinery in random.hpp

### LINKED ISSUE #769: Remove usage of <thread> in <range/v3/utility/random.hpp>
image

  
And the `RANGES_CXX_THREAD` configuration, including the CMake test.

CaseyCarter:
If your `std::random_device` is broken, *don't* use the default-provided random engine.

Also disables the despicable unsigned-integer-overflow sanitizer, which triggers in `std::uniform_int_distribution` causing builds to fail. (ATTN: @gnzlbg)

Fixes #769 and #1094.

Drive-by: libc++ trunk is now version 9.


gnzlbg:
>  which triggers in std::uniform_int_distribution causing builds to fail. (ATTN: @gnzlbg)

With which standard library is this ? That USAn check should only be enabled with `libc++`, which is tested with this check enabled, so it shouldn't fail there (otherwise that might be a bug worth reporting).

CaseyCarter:
> With which standard library is this?

libc++ 5.0.2: https://travis-ci.org/CaseyCarter/range-v3/builds/508703168

PR #1099, opened 2019-03-22
mixin for adaptor (view_adaptor) v3

tower120:
Same as v2 https://github.com/ericniebler/range-v3/pull/1086 but `adaptor()` called `get()`.


ericniebler:
Closing this since I already merged the earlier diff. Please make a new diff if you would like to make changes.

PR #1108, opened 2019-03-27
view::generate_n cache update optimisation

tower120:
This small change makes generate_n as fast as plain for-loop. 8x speed up with generated monotonically increasing int sequence (in compare to current version).

`emplace` has less branches then =. Because operator= also supports move assign value. This will never happens in `generate_n`.


ericniebler:
Nice!

Issue #1110, opened 2019-03-28
RandomAccess ranges::advance does not work with n = 0


tower120:
This triggers assertation error:
https://wandbox.org/permlink/Z6Xr5TZGgjx7Zry4
```cpp
    std::vector<int> v = {1,2,3};
    auto e = ranges::end(v);
    ranges::advance(e, 0, ranges::begin(v));
```

All advance-dependent functions affected (prev/next/etc.).

Issue #979, opened 2019-01-08
view::drop_last?


tower120:
There is `view::drop`, which "removes" **first** elements from sequence.
Is it possible to have the same, but for **last** elements?

cjdb:
I think this is solved by `view::take_exactly(size(r) - n)`.

tower120:
You can't take_exactly with pipe :)

With "drop_last" you could:
```
list | drop_last(1)
```

ericniebler:
> You can't take_exactly with pipe :)

Says who? https://godbolt.org/z/zWQwgt You can also use slice for this, but take and take_exactly are the better answers.

lewissbaker:
> I think this is solved by `view::take_exactly(size(r) - n)`.

This would prevent usage with input ranges or be inefficient for ranges where `size(r)` is not O(1), though.

ericniebler:
How would you drop the last N elements of an input range? When would you know how to stop taking elements if you didn't know the size? And size(r) is required to be O(1).

ericniebler:
That's a very "react" answer. :-) The views generally avoid copying/storing the elements. However, for a forward range (not input), a `drop_last` view could have iterators that stored two iterators into the underlying sequence, one N elements ahead of the other. (`begin()` would have to advance one iterator and cache it in the view to avoid incurring quadratic behavior in downstream adaptors). Deref would be from the "slow" iterator and comparison with the end would check the "fast" one. Increment would advance them both, obviously.

CaseyCarter:
FWIW: That's how `view::sliding` works. I suspect there's some useful commonality there that could be factored out.

tower120:
Just to note. For BidirectionalIterator case, it seems it's doable without caching iterators. [drop_last sentinel can be calculated on creation - by just `prev(end)` n times and comparing with `begin`]
So there would be 3 versions/specializations: RandomAccess, Bidirectional and Forward...

ericniebler:
> drop_last sentinel can be calculated on creation

This is a tempting and flawed implementation strategy. For one, it makes view creation O(N), which is poor form. But more seriously, there is a high chance that the sentinel will dangle after the view has been copied or moved, which means that it must be recomputed (at O(N) cost) on copy/move. Views are required to have O(1) copy/move.

tower120:
> there is a high chance that the sentinel will dangle after the view has been copied or moved

How sentinel can dangle/invalidate on copy/move? Not that I argue with this, just don't understand.

[...]

PR #1112, opened 2019-03-28
view::drop_last #979

tower120:
This PR implements https://github.com/ericniebler/range-v3/issues/979

It depends on fixing this https://github.com/ericniebler/range-v3/issues/1110 with (presumably) this https://github.com/ericniebler/range-v3/pull/1111 .
So, apply solution, then re-run tests for this PR. Without it - tests will fail.


include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+              : drop_last_forward_view::view_adaptor(std::move(rng))
+              , n(n)
+            {}
+        };
+
+        template<class Rng>
+        using drop_last_view =
+            meta::if_<BidirectionalRange<Rng>
+                , drop_last_bidirectional_view<Rng>
+                , drop_last_forward_view<Rng>>;
```
ericniebler: Prefer to define a `drop_last_view` struct rather than an alias template. The alias template prevents us from optionally defining deduction guides.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <range/v3/view_adaptor.hpp>
+#include <range/v3/utility/functional.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        template<typename Rng>
+        struct drop_last_bidirectional_view
+          : view_interface<drop_last_bidirectional_view<Rng>, range_cardinality<Rng>::value>
```
ericniebler: If the range cardinality is greater than or equal to zero, then it is _exactly_ the number of elements in the range. Obviously, if we are dropping elements of the range, that is no longer true.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+        {
+            CONCEPT_ASSERT(ranges::BidirectionalRange<Rng>());
```
ericniebler: `BidirectionalView` here.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+            {
+                return rng;
+            }
+        };
+
+        template<typename Rng>
+        struct drop_last_forward_view
+          : view_adaptor<drop_last_forward_view<Rng>, Rng>
+        {
+            CONCEPT_ASSERT(ranges::ForwardRange<Rng>());
```
ericniebler: `ForwardView` here.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+            }
+            Rng const & base() const
+            {
+                return rng;
+            }
+        };
+
+        template<typename Rng>
+        struct drop_last_forward_view
+          : view_adaptor<drop_last_forward_view<Rng>, Rng>
```
ericniebler: You need to similarly fix up the cardinality here.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+                void next(iterator_t<Rng>& it)
+                {
+                    ++it;
+                    ++probe;
+                }
+            };
+
+            struct sentinel_adaptor :  adaptor_base
+            {
+                bool empty(const iterator_t<Rng>&, const adaptor& ia, const sentinel_t<Rng>& s) const
```
ericniebler: East `const`, please.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    return ia.probe == s;
+                }
+            };
+
+            adaptor begin_adaptor() const { return {this->base(), n}; }
+            sentinel_adaptor end_adaptor() const { return {}; }
```
ericniebler: Some views are not `const`-iterable (e.g. `view::filter`). If you want to be able to `drop_last` such a view, you need non-`const`-qualified `(begin|end)_adaptor` members.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+            iterator_t<Rng> begin()
+            {
+                return ranges::begin(rng);
+            }
+            sentinel_t<Rng> end()
+            {
+                update_sentinel();
+                return *sentinel;
+            }
```
ericniebler: You'll need `const`-qualified `begin`/`end` also.

test/view/drop_last.cpp
```diff
@@ -0,0 +1,98 @@
...
+    static_assert(
+        std::is_same<
+            drop_last_view<Src>, drop_last_forward_view<Src>
+        >::value
+        , "Must have correct view.");
+
+    Src src = {1,2,3,4};
+
+    test_range(src);
+}
```
ericniebler: Please add tests for non-`const`-iterable forward (not-bidi) and bidi ranges.

test/view/drop_last.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+int main()
+{
+    random_acccess_test();
+    /*bidirectional_test();
+    forward_test();*/
```
ericniebler: Why are these commented out?

include/range/v3/view/drop_last.hpp
tower120: Why it must be View, not Range?

include/range/v3/view/drop_last.hpp
tower120: What about enforcing format with clang-format, as final CI step? Something like this https://github.com/Sarcasm/run-clang-format

As a bonus, many IDE's understand .clang-format, and will highlight/reformat problems.

johelegp: Unfortunately, that's not possible. See https://github.com/ericniebler/range-v3/issues/802#issuecomment-450400841.
Clang-Format does provide ways to mark range-for emulating macros (`ForEachMacros`), block macros (`MacroBlockBegin`, `MacroBlockEnd`) and statement macros (`StatementMacros`), but they're not enough for the concept emulation macros.

tower120: @johelegp can't it be just opt-outed for macroses entirely? 

In any case, I get it - it's already in TODO list.

include/range/v3/view/drop_last.hpp
ericniebler: The views all aggregate views, not ranges (which are not requires to be movable or copyable). The view adaptors are all responsible for turning into Views those Ranges that can be safely converted _before_ instantiating the view templates.

include/range/v3/view/drop_last.hpp
johelegp: There's no option for doing that. As the linked comment says, you'd have to splash these `// ...`-disabling comments everywhere.

include/range/v3/view/drop_last.hpp
tower120: I didn't work with deduction guides. Does simple wrapper around that alias, will do the job?
```cpp
       template<class Rng>
       using drop_last_view_ =
           meta::if_<BidirectionalRange<Rng>
               , drop_last_bidirectional_view<Rng>
               , drop_last_forward_view<Rng>>;

        template<class Rng>
        struct drop_last_view : drop_last_view_<Rng>{
            using drop_last_view_<Rng>::drop_last_view_;
        };
```
Or I need to implement it as template specializations for `drop_last_view<class Rng, bool is_bidi>`?

_BTW, aren't this solved in C++20?_

johelegp: > I didn't work with deduction guides. Does simple wrapper around that alias, will do the job?

No. CTAD only works when the name of the template is a class template.
> _BTW, aren't this solved in C++20?_

Apparently, the CTAD wording for alias templates is complex. Hopefully it makes it in time for C++20. I wonder if they couldn't merge what they have and iterate on it until C++20 is published. Even then, there are the defect reports.

include/range/v3/view/drop_last.hpp
tower120: I looked at `view::remove_if` and `drop_while` implementations. It seems that all iterators/adaptors that have "underlying cache" does not have const version. Should I follow that?

That seems weird to me. That's interior mutability. Unrelated to data in range. Caching iterator does not mutate underlying value in any way.

include/range/v3/view/drop_last.hpp
tower120: I can't - because of cached sentinel. Same as with `view::filter` .

tower120: From the other side - I'll just add on the fly computation for RandomAccessRange const end .

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,237 @@
...
+                CONCEPT_REQUIRES_(RandomAccessRange<CRng>())>
+            iterator_t<CRng> begin() const
+            {
+                return ranges::begin(rng);
+            }
+            template<typename CRng = Rng const,
+                CONCEPT_REQUIRES_(RandomAccessRange<CRng>())>
+            sentinel_t<CRng> end() const
+            {
+                return ranges::prev(ranges::end(rng), n, ranges::begin(rng));
```
ericniebler: This doesn't work if `ranges::end(rng)` returns a sentinel and not an iterator. You need to additionally require `CommonRange<CRng>()`. `begin() const` should too for consistency.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+                template<bool OtherConst,
+                    CONCEPT_REQUIRES_(IsConst && !OtherConst)>
+                adaptor(adaptor<OtherConst> other)
+                    : probe(other.probe)
+                {}
+
+                adaptor(CRng &rng, difference_t n)
+                {
+                    probe = ranges::next(ranges::begin(rng), n, ranges::end(rng));
```
ericniebler: This makes `begin()` O(N) always. The probe should be stored in a `non_propagating_cache` in `drop_last_view` and initialized on first access.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    return {all(static_cast<Rng&&>(rng)), n};
+                }
+            public:
+                template<typename Rng,
+                    CONCEPT_REQUIRES_(ForwardRange<Rng>())>
+                auto operator()(Rng &&rng, range_difference_type_t<Rng> n) const
+                RANGES_DECLTYPE_AUTO_RETURN
+                (
+                    drop_last_fn::invoke_(static_cast<Rng&&>(rng), n)
```
ericniebler: Why not just return `drop_last_view<all_t<Rng>>{all(static_cast<Rng&&>(rng)), n}` ?

test/view/drop_last.cpp
```diff
@@ -0,0 +1,145 @@
...
+    friend ranges::range_access;
+
+    ranges::adaptor_base begin_adaptor() { return {}; }
+    ranges::adaptor_base end_adaptor()   { return {}; }
+public:
+    using view_non_const_only::view_adaptor::view_adaptor;
+};
+
+template<class Rng>
+view_non_const_only<view::all_t<Rng>> non_const_only(Rng &&rng){
```
ericniebler: Open-curlies on newlines, please.

include/range/v3/view/drop_last.hpp
tower120: Instead of `drop_last_fn::invoke_(static_cast<Rng&&>(rng), n)` ?

include/range/v3/view/drop_last.hpp
tower120: Hmm if so... Shouldn't  non-const begin/end be CommonRange too? And since all begins/ends require CommonRange (both Forward and Bidi), maybe CommonRange should be required at view level?

BTW - `view::drop` need this "fix" too.

include/range/v3/view/drop_last.hpp
tower120: Good point. Though only `probe_begin` should be stored in view. `probe` should still live in iterator (but now inited from `probe_begin`), since it need to be incremented on each `next()` step. Otherwise, each ++it will increase view's probe. 

And now view becomes mut only.

tower120:
Now it support any SizedRange (mode::sized).
`mode::sized` is faster then `mode::forward` and `mode::bidi`(for Bidi). But result range becomes unbound, and it works with non-original iterators.
In mode dispatcher `aux::drop_last::get_mode()`, currently presented two strategies. They differs in how sized bibi handled.

include/range/v3/view/drop_last.hpp
```diff
@@ -0,0 +1,343 @@
...
+#include <range/v3/view_interface.hpp>
+#include <range/v3/view_adaptor.hpp>
+#include <range/v3/utility/functional.hpp>
+#include <range/v3/utility/counted_iterator.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        namespace aux{ namespace drop_last_view {
```
ericniebler: Every `{` on its own line, every namespace causes a 4-space indent, please use `detail` instead of `aux`, and wrap all internal details in `/// \cond` / `/// \endcond` to hide them from Doxygen. Here and elsewhere. Otherwise, looks good.

tower120:
Emmm.... What about `aux::drop_last::get_mode()` https://github.com/ericniebler/range-v3/pull/1112#issuecomment-485285652 ? Which strategy to use?

PR #1117, opened 2019-04-02
MSVC 1.0 updates (again)

CaseyCarter:
* iter.iterator:
  * remove extraneous typename disambiguators applied to non-dependent types
  * don't redundantly test that contiguous iterator is random access

* Use `__has_include` to detect `<string_view>` in `range/access.hpp`

* Workaround VSO#792338 "Failure to match specialization enabled via call to constexpr function"
* Workaround VSO#793042 in `<iterator>`
* Workaround VSO#779708 again

* VS2019: Disable unneeded MSVC workarounds
* Update README

* Disable C5105 "macro expansion producing 'defined' has undefined behavior"

### DIFF
--- a/test/range/conversion.cpp
+++ b/test/range/conversion.cpp
@@ -48,7 +48,11 @@ template<
 void test_zip_to_map(Rng &&rng, int)
 {
     using namespace ranges;
+#ifdef RANGES_WORKAROUND_MSVC_779708
+    auto m = static_cast<Rng &&>(rng) | to<std::map>();
+#else // ^^^ workaround / no workaround vvv
     auto m = static_cast<Rng &&>(rng) | to<std::map>;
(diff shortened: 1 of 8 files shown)


CaseyCarter:
With this change, everything is passing with the exception of the generator test. Our generator support is "experimental" anyway, so I can live with releasing with it broken.

test/range/conversion.cpp
```diff
@@ -69,7 +73,7 @@ int main()
         ::check_equal(lst0, {0,1,4,9,16,25,36,49,64,81});
     }
 
-#ifndef RANGES_WORKAROUND_MSVC_779708 // "workaround" is a misnomer - there's no workaround.
```
ericniebler: This is deeply satisfying to my inner grammar geek. Thank you.

PR #1126, opened 2019-04-09
More detailed view_adaptor doc

tower120:
By the way, why `empty()` is needed? `equal()` could accept sentinel instead...

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -478,7 +478,7 @@ Sometimes, you can use the same adaptor for both `begin_adaptor` and `end_adapto
     adaptor end_adaptor()   const { return {100}; }
 ~~~~~~~
 
-Note that all the data that you store in the adaptor will become part of the iterator.
+Note, that all the data, that you store in the adaptor, will become part of the iterator.
 
 If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor, default ones will be used.
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -480,7 +480,7 @@ Sometimes, you can use the same adaptor for both `begin_adaptor` and `end_adapto
 
 Pay attentention that all data, that you'll store in adaptor - will become part of iterator payload.
 
-If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor - default ones will be used.
+If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor, default ones will be used.
 
 ## Create Custom Iterators
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -478,7 +478,7 @@ Sometimes, you can use the same adaptor for both `begin_adaptor` and `end_adapto
     adaptor end_adaptor()   const { return {100}; }
 ~~~~~~~
 
-Pay attentention that all data, that you'll store in adaptor - will become part of iterator payload.
+Note that all the data that you store in the adaptor will become part of the iterator.
 
 If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor, default ones will be used.
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -316,7 +316,7 @@ to inject things into the public interface of the iterator:
               // everything inside this class will be accessible from iterator
               using base_mixin::base_mixin;
 
-              decltype(auto) base_value() const
+              auto& base_value() const
               {
                   return *this->base();
               }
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -331,6 +331,157 @@ to inject things into the public interface of the iterator:
     };
 ~~~~~~~
 
+From within mixin you can call:
+* `get()` - to access adaptor internals 
+* `base()` - to access adopted iterator
```
ericniebler: adopted?

doc/index.md
```diff
@@ -331,6 +331,157 @@ to inject things into the public interface of the iterator:
...
+
+Iterator/sentinel adaptor may "override" following members:
+~~~~~~~{.cpp}
+    class adaptor : public ranges::adaptor_base
+    {
+        // !For begin_adaptor only!
+        template<typename Rng>
+        constexpr decltype(auto) begin(Rng &rng)
```
ericniebler: s/`decltype(auto)`/`auto`/

doc/index.md
```diff
@@ -331,6 +331,157 @@ to inject things into the public interface of the iterator:
...
+        {
+            return *this.n == that_adapt.n;
+        }
+    };
+
+    adaptor begin_adaptor() const { return {}; }
+    adaptor end_adaptor()   const { return {100}; }
+~~~~~~~
+
+Pay attentention that all data, that you'll store in adaptor - will become part of iterator payload.
```
```suggestion
Note that all the data that you store in the adaptor will become part of the iterator.
```

doc/index.md
```diff
@@ -331,6 +331,157 @@ to inject things into the public interface of the iterator:
...
+
+If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor - default ones will be used.
```
```suggestion
If you will not "override" `begin_adaptor()` or/and `end_adaptor()` in your view_adaptor, default ones will be used.
```

ericniebler:
> By the way, why `empty()` is needed? `equal()` could accept sentinel instead...

I don't remember. 😳

doc/index.md
tower120: adapted

doc/index.md
tower120: ??

doc/index.md
tower120: Applying, but I disagree that commas are better than dashes.

tower120:
> I don't remember. 

I assume, that was needed for the case, when you compared [sentinel with sentinel (?)], **and** [sentinel with iterator]. Probably, with separate `empty()`, dispatching looked more naturally in initial versions...

That all would be just a matter of taste, if not this particular case (not that rare BTW): https://github.com/ericniebler/range-v3/issues/1127

_... Maybe it could be removed... Some day? Refactoring of all views using `empty()` looks trivial..._

doc/index.md
ericniebler: Yes, really.

tower120: `??` = don't understand you completely.

----

`s/decltype(auto)/auto/` how this translates to English?

ericniebler: Oh! I'm so sorry. I was speaking sed. /facepalm. Let me try English. Substitute ("s/") `auto` for `decltype(auto)`. In other words:

```suggestion
        constexpr auto begin(Rng &rng)
```

ericniebler:
In all honesty, I'm happier with the design of `view_facade` than I am of `view_adaptor`. I've considered dropping `view_adaptor` entirely and just using facade. Before I do that, I'll probably want to see if could redesign the adaptor so that I'm happier with it.

TL;DR: This all needs thought.

tower120:
>  I've considered dropping view_adaptor entirely and just using facade. 

You need machinery to override iterator/sentinel "customization points". 

I think there is need in `cursor_adaptor` (not the current one), which would have the same interface as cursor (without begin/end/empty/etc.), and would just forward all cursor calls to underlying iterator:

```cpp

template<class BaseIterator, class Derived = void>
struct cursor_adaptor{
    BaseIterator base;

    cursor_adaptor() = default;
    cursor_adaptor(BaseIterator base) : base(base) {}

     // enable_if Derived != void   ( optionally hide base() )
    struct mixin : ranges::basic_mixin<Derived>
    {
        using ranges::basic_mixin::basic_mixin;

        BaseIterator&  base();
    };  

    // forward cursor interface whenever possible
    void next() { 
        ++base;
    }
};

struct my_cursor : cursor_adaptor<BaseIterator, my_cursor>{
    void next() { 
        base += 2;
    }    

    // optional
    struct mixin : my_cursor:::cursor_adaptor::mixin
    {
        // add anything you need.
    };  

};

basic_iterator<my_cursor> my_iter(vec.begin());
```

On top of that you can build everything.

PR #1130, opened 2019-04-10
Update TestHeaders CMake module

johelegp:
This prevents the massive amount of generated targets from appearing in graphical contexts that overshadow the local project's targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,7 @@ else()
 endif()
 if(RANGE_V3_TESTS)
   add_header_test(test.rangev3.headers HEADERS ${RANGE_V3_PUBLIC_HEADERS})
-  target_link_dependencies(test.rangev3.headers PRIVATE range-v3)
+  target_link_libraries(test.rangev3.headers PRIVATE range-v3)
 endif()
 set_target_properties(headers
     PROPERTIES FOLDER "test/header"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,7 +85,8 @@ else()
   add_custom_target(headers SOURCES ${RANGE_V3_PUBLIC_HEADERS_ABSOLUTE})
 endif()
 if(RANGE_V3_TESTS)
-  add_header_test(range_v3_headers HEADERS ${RANGE_V3_PUBLIC_HEADERS})
+  add_header_test(test.rangev3.headers HEADERS ${RANGE_V3_PUBLIC_HEADERS})
+  target_link_dependencies(test.rangev3.headers PRIVATE range-v3)
 endif()
 set_target_properties(headers
     PROPERTIES FOLDER "test/header"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,6 @@ function(rv3_add_test TESTNAME EXENAME FIRSTSOURCE)
   add_test(${TESTNAME} ${EXENAME})
 endfunction(rv3_add_test)
 
-include(TestHeaders)
 include(ranges_options)
 include(ranges_env)
 include(ranges_flags)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,8 +82,20 @@ if(RANGE_V3_HEADER_CHECKS)
                     "${CMAKE_CURRENT_SOURCE_DIR}/include/*.hpp")
   # This header is not meant to be included directly:
   list(REMOVE_ITEM RANGE_V3_PUBLIC_HEADERS std/detail/associated_types.hpp)
+  # Deprecated headers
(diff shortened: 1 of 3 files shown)


johelegp:
An example of QtCreator. When Ranges v3 adds/renames a header, it is set to be run, and you have go on a scrolling quest to reset your target again.
scroll_quest

johelegp:
As a better alternative, we could backport the [updates to `TestHeaders.cmake`](https://github.com/boostorg/hana/commits/master/cmake/TestHeaders.cmake). In particular, the latest change (https://github.com/boostorg/hana/commit/126cdb8e0d83a4d24bcea50e485d0c3181ed9643#diff-582339976497cc3d8cfa5261b99d3ad1, committed on Nov 25, 2017) would solve my problem. @ldionne How could we go about doing this? If licenses allow, I could update this PR to be such backport.

CaseyCarter:
> As a better alternative, we could backport the updates to TestHeaders.cmake.

Oh, that's nice. The licenses are both Boost - range-v3's `TestHeaders.cmake` came from @ldionne in the first place - so I'm sure he'd be happy for us to take an update.

And apologies for not responding to the PR yet. I took a glance at it, and I've been debating with myself whether we should simply get rid of the header tests altogether. They've found 0 bugs in the last 2-3 years.

I'd very much appreciate a PR with an update of `TestHeaders`. I'd also appreciate if `RANGE_V3_HEADER_CHECKS` was `OFF` by default and therefore only run in CI, but @ericniebler may differ on that point.

johelegp:
> I'd also appreciate if `RANGE_V3_HEADER_CHECKS` was `OFF` by default and therefore only run in CI, but @ericniebler may differ on that point.

That's how it works already. It's just that the targets were created nevertheless, since the call to `generate_standalone_header_tests` wasn't guarded. They weren't build, though, as `EXCLUDE_FROM_ALL` was set.
https://github.com/ericniebler/range-v3/blob/6d7885c8471bef73c038501e5665b6b69d6baa17/CMakeLists.txt#L88

johelegp:
> They weren't build, though, as `EXCLUDE_FROM_ALL` was set.

For dependent projects, of course. I still guarded the call because I had to remove `EXCLUDE_FROM_ALL` given than `MASTER_TARGET` was removed.

CaseyCarter:
> For dependent projects, of course. 

Yes, that's the case. I'd like `RANGE_V3_HEADER_CHECKS` to default to `OFF` instead of `${is_standalone}`; it's annoying to always have to explicitly turn it off for every non-standalone config. I rarely want to test the headers when I'm testing the rest of range-v3.

But this is my problem, not yours, and I should handle it in a separate change rather than trying to get you to do it for me.

CMakeLists.txt
```diff
@@ -85,7 +84,10 @@ if(RANGE_V3_HEADER_CHECKS)
 else()
   add_custom_target(headers SOURCES ${RANGE_V3_PUBLIC_HEADERS_ABSOLUTE})
 endif()
-generate_standalone_header_tests(EXCLUDE_FROM_ALL MASTER_TARGET headers HEADERS ${RANGE_V3_PUBLIC_HEADERS})
```
CaseyCarter: There's no reason now for `headers` to be attached to `all`; it doesn't build anything. We can remove lines 82-84 and 86 and just leave the fake target so IDEs know the header files are all part of the project.

Also, do you see any reason not to move the globs and include (lines 70 - 81) into the `if(RANGE_V3_TESTS)` on 87?

johelegp:
Ah, I had misunderstood you since it defaulted to `OFF` from my perspective. I'll leave that to you as it seems to require a more global view of the project.

CMakeLists.txt
johelegp: The glob one for the IDEs was left outside without the `test/` prefix for its `FOLDER` property.

johelegp:
> I'd like `RANGE_V3_HEADER_CHECKS` to default to `OFF` instead of `${is_standalone}`

I did it. I thought I'd require to touch the CI config files, but they're well setup.

Now, it seems that for GCC, I'd have to remove all headers that include `RANGES_DEPRECATED_HEADER`.

johelegp:
I should have tested locally way earlier. Everything should be fine now.

CMakeLists.txt
CaseyCarter: > The glob one for the IDEs was left outside without the `test/` prefix for its `FOLDER` property.

Oh, obviously.

CaseyCarter:
I'm pulling the trigger on this PR. @ericniebler, complain if you don't like the bit that makes `RANGE_V3_HEADER_CHECKS` purely opt-in and I'll revert it.

PR #1132, opened 2019-04-10
fix libcpp aligned new detection for iOS

mtak-:
Currently range-v3/master does not build when targeting iOS, due to iOS lacking support for aligned new.

This PR disables aligned new when `_LIBCPP_HAS_NO_ALIGNED_ALLOCATION` is defined which allows range-v3 to build for iOS targets.


ericniebler:
Thank you!

PR #1133, opened 2019-04-11
Bring box::coalesce to the land of constexpr!

tower120:
This PR:
* Adds `is_constexpr_default_cosntructible` probe.
* Replaces `box::coalesce` `is_trivial` requirement with `is_constexpr_default_cosntructible`.
* Removes potentially unnecessary `is_trivially_copyable` requirement, since underlying static value actually never copied.

Previous discussion here: https://github.com/ericniebler/range-v3/pull/1096#issuecomment-481380161


tower120:
I can't make `is_constexpr_default_cosntructible` work reliably in clang without `is_literal_type`. And that thing is deprecated and removed.

Closing this.

PR #1135, opened 2019-04-14
documentation: view constness

tower120:
Short explanation of view constness and validity.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -133,7 +133,7 @@ Because of the cache, many views lack a `const`-qualified `begin()`/`end()`.
 When `const` versions of `begin()`/`end()` are provided, they are truly `const`; that is, thread-safe.
 But since they can't cache - const versions usually slower then non-const.
 
-Constness of view is not constness of iterators. Non-const view may hold const iterators.
+As with pointers and iterators (an `int* const` is a `const` pointer to a mutable `int`), the `const`-ness of a view is not related in any way to the `const`-ness of the underlying data.
 
 Use non-const views whenever possible. If you need thread-safety - work with view copies in threads, don't share.
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -139,7 +139,7 @@ Use non-const views whenever possible. If you need thread-safety - work with vie
 
 ### View validity
 
-View is guaranteed to be valid, only until it adopted range remains constant all over its span.
+Any operation on the underlying range that invalidates its iterators or sentinels will also invalidate any view that refers to any part of that range. Some views (_e.g._, `view::filter`) additionally are invalidated when the underlying elements of the range are mutated. It is best to recreate a view after any operation that may have mutated the underlying range.
 Iterator/sentinel validity alone is not enough.
 
 ~~~~~~~{.cpp}
--- a/doc/index.md
+++ b/doc/index.md
@@ -141,15 +141,6 @@ Use non-const views whenever possible. If you need thread-safety - work with vie
 
 Any operation on the underlying range, that invalidates its iterators or sentinels, will also invalidate any view that refers to any part of that range. Additionally, some views (_e.g._, `view::filter`), are invalidated when the underlying elements of the range are mutated. It is best to recreate a view after any operation, that may have mutated the underlying range.
 
-~~~~~~~{.cpp}
-    using namespace ranges;
-    std::forward_list<int> list = {1,2,3,4};
-    auto v = list | view::drop_last(1);
-    iter_swap(advance(list.begin(), 1), advance(list.begin(), 2));
-    // v is invalid now.
-~~~~~~~
-
-
 #### Actions
 
 When you want to mutate a container in-place, or forward it through a chain of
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -125,6 +125,33 @@ vector with it:
...
+### View validity
+
+View is guaranteed to be valid, only until it adopted range remains constant all over its span.
+Iterator/sentinel validity alone is not enough.
+
+~~~~~~~{.cpp}
+    using namespace ranges;
+    std::forward_list<int> list = {1,2,3,4};
+    auto v = list | view::drop_last(1);
+    iter_swap(advance(list.begin(), 1), advance(list.begin(), 2));
```
ericniebler: This code is invalid. You probably meant `next`, not `advance`. And I don't think this code actually invalidates the view `v`. Before the `iter_swap`, `v` refers to the range [1,2,3] and afterward to [1,3,2]. Or have I misunderstood?

ericniebler: This code is invalid. You probably meant `next`, not `advance`.

doc/index.md
```diff
@@ -125,6 +125,33 @@ vector with it:
...
+Const version of `begin()`/`end()` (where provided) - are truly const, hence - tread-safe.
+But since they can't cache - const versions usually slower then non-const.
+
+Constness of view is not constness of iterators. Non-const view may hold const iterators.
+
+Use non-const views whenever possible. If you need thread-safety - work with view copies in threads, don't share.
+
+### View validity
+
+View is guaranteed to be valid, only until it adopted range remains constant all over its span.
```
```suggestion
Any operation on the underlying range that invalidates its iterators or sentinels will also invalidate any view that refers to any part of that range. Some views (_e.g._, `view::filter`) additionally are invalidated when the underlying elements of the range are mutated. It is best to recreate a view after any operation that may have mutated the underlying range.
```

doc/index.md
```diff
@@ -125,6 +125,33 @@ vector with it:
...
+Iterator/sentinel validity alone is not enough.
```
```suggestion
```

doc/index.md
```diff
@@ -125,6 +125,33 @@ vector with it:
...
+
+~~~~~~~{.cpp}
+    using namespace ranges;
+    std::forward_list<int> list = {1,2,3,4};
+    auto v = list | view::drop_last(1);
+    iter_swap(advance(list.begin(), 1), advance(list.begin(), 2));
+    // v is invalid now.
+~~~~~~~
+
+Though, sometimes view may remain valid even after mutations in adopted range, it is not safe to assume so.
```
```suggestion
```

doc/index.md
tower120: > You probably meant next, not advance

Right.

> And I don't think this code actually invalidates the view v ...

You're right, I think, I imagined myself that on `iter_swap` iterators will be invalidated. 

I'll try change example.

doc/index.md
```diff
@@ -125,6 +125,33 @@ vector with it:
     // vi == {1,2,2,3,3,3,4,4,4,4,5,5,5,5,5,...}
 ~~~~~~~
 
+### View constness
+
+View is a range, pair of iterators. In order to work, and work fast, many views need to cache some data.
+In order to keep iterators small sized - this cached data stored in the view itself, and iterators holds only pointers to their view.
+Because of cache, many views have non-const only `begin()`/`end()` (though they may return const-iterators).
+Const version of `begin()`/`end()` (where provided) - are truly const, hence - tread-safe.
+But since they can't cache - const versions usually slower then non-const.
```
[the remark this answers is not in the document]
johelegp: If `begin()` or `end()` needs to cache, they can't have a `const` overload because the `Range` concept imposes on them an amortized constant complexity requirement in http://wg21.link/range.range#3.2. So such slower versions are non-existent.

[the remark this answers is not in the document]
tower120: @johelegp But it not says that const and non-const versions must have **exact** complexity and performance characteristics.

Imagine following.

`drop_exactly`. Currently, RandomAccess const and non-const versions are essentially the same. But if non-const version would cache (like !RandomAccess do now) - it would be faster for `std::deque`. 

```cpp
            // this is measurably faster for std::deque
            iterator_t<Rng> begin()
            {
                // copy-pasted from !RA version
                if(!begin_)
                    begin_ = next(ranges::begin(rng_), n_);
                return *begin_;
            }
            // fallback to slower version
            iterator_t<CRng> begin() const
            {
            }
```

Both versions have amortized constant complexity.

doc/index.md
tower120: I changed it a bit...

doc/index.md
tower120: You know, from the other side, it looks good without example. I remove it.

doc/index.md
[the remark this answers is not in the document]
ericniebler: In the docs we have a lot to talk about, and the developer's time is precious. I'd like the docs to talk about complexity guarantees because they are important. I don't want the docs to lose that focus or make needless distinctions. What you say is not wrong, but I would leave it out in the interest of keeping the docs tightly focused.

tower120:
I think you forgot this suggestion https://github.com/ericniebler/range-v3/pull/1135#discussion_r276431575 of yourself.

PR #1137, opened 2019-04-16
Add trim view

johelegp:
I hope this is within scope. I thought I'd try getting this in so that it's available for the broader public rather than stuffing it in a library of mine. I'm looking forward to the review comments.

The current implementation is simple thanks to the bidirectional range requirement. Looking at #1112, it should be possible to do with a forward range. But I believe it'd be simpler to add a complementing `drop_last_while` and implement `view::trim` in terms of it and `drop_while`. This possibility makes me wonder if this is okay as is.

### DIFF
--- a/include/range/v3/view/trim.hpp
+++ b/include/range/v3/view/trim.hpp
@@ -50,7 +50,7 @@ namespace ranges
     public:
         CPP_assert(BidirectionalView<Rng> &&
             IndirectUnaryPredicate<Pred, iterator_t<Rng>> &&
-            std::is_object<Pred>::value);
+            CommonRange<Rng>);
 
         trim_view() = default;
         trim_view(Rng rng, Pred pred)
@@ -112,7 +112,7 @@ namespace ranges
                 CPP_ret(trim_view<all_t<Rng>, Pred>)(
                     requires ViewableRange<Rng> && BidirectionalRange<Rng> &&
                         IndirectUnaryPredicate<Pred, iterator_t<Rng>> &&
-                        std::is_object<Pred>::value)
+                        CommonRange<Rng>)
             {
                 return {all(static_cast<Rng &&>(rng)), std::move(pred)};
             }
@@ -121,7 +121,7 @@ namespace ranges
                 CPP_ret(trim_view<all_t<Rng>, composed<Pred, Proj>>)(
                     requires ViewableRange<Rng> && BidirectionalRange<Rng> &&
                         IndirectUnaryPredicate<composed<Pred, Proj>, iterator_t<Rng>> &&
-                        std::is_object<Pred>::value && std::is_object<Proj>::value)
+                        CommonRange<Rng>)
             {
                 return {
                     all(static_cast<Rng &&>(rng)),
--- a/include/range/v3/view/trim.hpp
+++ b/include/range/v3/view/trim.hpp
@@ -1,0 +1,147 @@
+/// \file
+// Range v3 library
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGES_V3_VIEW_TRIM_HPP
+#define RANGES_V3_VIEW_TRIM_HPP
+
+#include <functional>
+#include <type_traits>
+#include <utility>
+#include <concepts/concepts.hpp>
+#include <range/v3/algorithm/find_if_not.hpp>
+#include <range/v3/detail/config.hpp>
+#include <range/v3/iterator/concepts.hpp>
+#include <range/v3/functional/bind.hpp>
+#include <range/v3/functional/compose.hpp>
+#include <range/v3/functional/pipeable.hpp>
+#include <range/v3/range/access.hpp>
+#include <range/v3/range/concepts.hpp>
+#include <range/v3/range/primitives.hpp>
+#include <range/v3/utility/optional.hpp>
+#include <range/v3/utility/semiregular.hpp>
+#include <range/v3/view/all.hpp>
+#include <range/v3/view/interface.hpp>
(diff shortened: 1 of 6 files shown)


include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
+/// \file
+// Range v3 library
+//
+//  Use, modification and distribution is subject to the
```
ericniebler: You wrote this code. You should have your copyright in it.

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+        trim_view() = default;
+        trim_view(Rng rng, Pred pred)
+          : rng_(std::move(rng)), pred_(std::move(pred))
+        {}
+        CPP_template(typename R)(
+            requires BidirectionalRange<R> && ViewableRange<R> &&
+                Constructible<Rng, view::all_t<R>>)
+        trim_view(R &&rng, Pred pred)
+          : rng_(view::all(static_cast<R &&>(rng))), pred_(std::move(pred))
+        {}
```
ericniebler: Nuke this constructor, please. Perhaps you saw similar constructors on the views in C++20. Those constructors are similarly a mistake and will be removed (I hope).

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+        }
+    };
+
+#if RANGES_CXX_DEDUCTION_GUIDES >= RANGES_CXX_DEDUCTION_GUIDES_17
+    template<typename Rng, typename Pred>
+    trim_view(Rng &&, Pred) -> trim_view<view::all_t<Rng>, Pred>;
+#endif
+
+    template<typename Rng, typename Pred>
+    RANGES_INLINE_VAR constexpr bool disable_sized_range<trim_view<Rng, Pred>> = true;
```
ericniebler: This specialization isn't necessary.

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+    private:
+        Rng rng_;
+        semiregular_t<Pred> pred_;
+        detail::non_propagating_cache<iterator_t<Rng>> begin_;
+        detail::non_propagating_cache<iterator_t<Rng>> end_;
+
+    public:
+        CPP_assert(BidirectionalView<Rng> &&
+            IndirectUnaryPredicate<Pred, iterator_t<Rng>> &&
+            std::is_object<Pred>::value);
```
ericniebler:
```suggestion
            CommonRange<Rng>);
```

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+                    std::bind(trim, std::placeholders::_1, protect(std::move(pred)),
+                    protect(std::move(proj))));
+            }
+        public:
+            template<typename Rng, typename Pred>
+            auto operator()(Rng &&rng, Pred pred) const ->
+                CPP_ret(trim_view<all_t<Rng>, Pred>)(
+                    requires ViewableRange<Rng> && BidirectionalRange<Rng> &&
+                        IndirectUnaryPredicate<Pred, iterator_t<Rng>> &&
+                        std::is_object<Pred>::value)
```
ericniebler:
```suggestion
                        CommonRange<Rng>)
```

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+            {
+                return {all(static_cast<Rng &&>(rng)), std::move(pred)};
+            }
+            template<typename Rng, typename Pred, typename Proj>
+            auto operator()(Rng &&rng, Pred pred, Proj proj) const ->
+                CPP_ret(trim_view<all_t<Rng>, composed<Pred, Proj>>)(
+                    requires ViewableRange<Rng> && BidirectionalRange<Rng> &&
+                        IndirectUnaryPredicate<composed<Pred, Proj>, iterator_t<Rng>> &&
+                        std::is_object<Pred>::value && std::is_object<Proj>::value)
```
ericniebler:
```suggestion
                        CommonRange<Rng>)
```

include/range/v3/view/trim.hpp
```diff
@@ -0,0 +1,147 @@
...
+            if(!begin_)
+                begin_ = find_if_not(rng_, std::ref(pred_));
+            return *begin_;
+        }
+        iterator_t<Rng> end()
+        {
+            if(!end_)
+            {
+                auto rbase = view::reverse(rng_);
+                end_= find_if_not(rbase, std::ref(pred_)).base();
```
ericniebler: I have a mild preference for this to be implemented directly rather than with `view::reverse`. Reverse views are slightly inefficient.

include/range/v3/view/trim.hpp
johelegp: Aren't the intrinsics the same as in the standard? If left alone, `ranges::size` would return `v.begin() - v.end()`. That'd violate the non-amortized constant complexity of `ranges::size`.

johelegp: That'd be for the case where `Rng` models `RandomAccessRange`.

johelegp: Correction: It is `SizedRange` that has a complexity requirement on `ranges::size`.

johelegp:
The latter part of d28764f's message shouldn't matter as the range concepts assign no semantics at all to default initialization.

johelegp:
I guess default-initialized views are to be treated as the moved-from state (like being unable to get their begin/end). I have reverted the attempt to assert whether `pred_` was not in its default-initialized state.

johelegp:
Is it alright to mark `find_if_not` as `constexpr /*c++14*/` so that `trim_view` can be `constexpr` too? I thought that you were holding back the algorithms from being `constexpr` since their standard version weren't, but I noticed that `equal` is `constexpr`.

ericniebler:
> Is it alright to mark find_if_not as `constexpr /*c++14*/` so that `trim_view` can be `constexpr` too? I thought that you were holding back the algorithms from being `constexpr` since their standard version weren't, but I noticed that `equal` is `constexpr`.

Fine by me.

PR #1139, opened 2019-04-17
fully support user-defined integral types for iterator difference_type, refs #591

ericniebler:


ericniebler:
This makes it possible for range-v3 itself to define and use UDT difference types, but users of range-v3 are not given that flexibility yet. I haven't yet defined the concept such types must satisfy. Conversions and promotions between such types will be a particularly sticky issue.

include/range/v3/utility/memory.hpp
```diff
@@ -37,7 +37,7 @@ namespace ranges
     namespace detail
     {
         template<typename T>
-        std::pair<T *, std::ptrdiff_t> get_temporary_buffer(std::ptrdiff_t count) noexcept
+        std::pair<T *, std::ptrdiff_t> get_temporary_buffer_impl(std::size_t count) noexcept
```
marehr: here was the change to `std::size_t`

PR #1143, opened 2019-04-20
partially fix primary expressions in requires clause (#1138)

### LINKED ISSUE #1138: [concept] requires only allows primary expressions
I'm currently trying to get a work-in-progress version of clang which supports concepts (see https://github.com/saarraz/clang-concepts) to work with the `v1.0-beta` version of ranges-v3.

After a bit of trial and error I encountered the following issue:

> https://en.cppreference.com/w/cpp/language/constraints#Requires_clauses
> The expression must have one of the following forms:
>    * a primary expression, e.g. Swappable<T>, std::is_integral<T>::value, (std::is_object_v<Args> && ...), or any parenthesized expression
>    * a sequence of primary expressions joined with the operator &&
>    * a sequence of aforementioned expressions joined with the operator ||

The ranges library is currently violating this on multiple occasions:

https://github.com/ericniebler/range-v3/blob/36c651faeb5c81bc28ac9ffcd97ad0602f2fade4/include/range/v3/iterator/common_iterator.hpp#L176-L182
`not ForwardIterator<I>` is not a primary expression, but `(not ForwardIterator<I>)` would be

https://github.com/ericniebler/range-v3/blob/36c651faeb5c81bc28ac9ffcd97ad0602f2fade4/include/range/v3/iterator/basic_iterator.hpp#L574-L582
same here

https://github.com/ericniebler/range-v3/blob/36c651faeb5c81bc28ac9ffcd97ad0602f2fade4/include/range/v3/detail/variant.hpp#L140-L145

`sizeof...(Ts) != 0` is not a primary expression

https://github.com/ericniebler/range-v3/blob/36c651faeb5c81bc28ac9ffcd97ad0602f2fade4/include/range/v3/detail/variant.hpp#L681-L686

`all_convertible_to<Args...>(0)` is not a primary expression

marehr:
partially fixes; #1138

### DIFF
--- a/include/range/v3/action/action.hpp
+++ b/include/range/v3/action/action.hpp
@@ -71,7 +71,7 @@ namespace ranges
                 CPP_ret(invoke_result_t<Action &, Rng>)(
                     requires Range<Rng> &&
                         Invocable<Action &, Rng> &&
-                        !std::is_reference<Rng>::value)
+                        (!std::is_reference<Rng>::value))
             {
                 return invoke(act.action_, detail::move(rng));
             }
--- a/include/range/v3/action/drop_while.hpp
+++ b/include/range/v3/action/drop_while.hpp
@@ -35,7 +35,7 @@ namespace ranges
             friend action_access;
             template<typename Fun>
             static auto CPP_fun(bind)(drop_while_fn drop_while, Fun fun)(
-                requires !Range<Fun>)
+                requires (!Range<Fun>))
             {
                 return std::bind(drop_while, std::placeholders::_1, std::move(fun));
             }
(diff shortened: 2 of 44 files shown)


CaseyCarter:
You may want to submit these in smaller batches; reviewing large mostly-mechanical changes is not fun.

marehr:
I'm sorry for this. I wasn't sure which granularity you prefer. Per file commits, per subfolder commits. Multiple PRs for each commit? It would have been fine for me if you had rejected. Thank you for the effort!

CaseyCarter:
> I wasn't sure which granularity you prefer.

Something roughly 1/4 to 1/3 of this size that we can review in a few minutes would be ideal. Also, I sound like a bit of an asshat above - sorry for that. Thank you very much for this.

PR #1148, opened 2019-04-21
Add ends_with algorithm

### LINKED ISSUE #1144: Add ends_with as an algorithm
C++20 adds `starts_with` and `ends_with` as member functions to `basic_string`. I think `ends_with` would be a useful refactoring of `mismatch(x | view::reverse, y | view::reverse).in2.base() == end(y)`.

johelegp:
Resolves #1144.

What do you think about the requirements? As it is now, it requires sized ranges or iterator/sentinel pairs, the first one being random access. This allows implementing the lowest complexity.


cjdb:
This is much simpler than my implementation, which doesn't use `drop`. If it can support forward ranges, that'd be great 😄 

Also, I forgot to mention an interface in #1144, but `ends_with(r1, r2, comp, proj1, proj2)` is exactly what I was thinking too.

Great work so far!

johelegp:
Thank you. Perhaps, like `equal`, `ends_with` should only require input ranges and do its best effort on algorithmic complexity based on the refined concepts modelled by the ranges.

johelegp:
Should `ends_with` return a result type? When `false` is returned, you loss the information as to which iterators from the ranges did not satisfy the predicate.

cjdb:
I don't know, I think this falls into the same category as `equal`, which is just a nice wrapper for `mismatch`. Having said that, `ends_with(x, y).result` doesn't affect readability by much.

You might also like to consider an interface where you don't have a sized range, but do have a sized sentinel.

johelegp:
The `.result` isn't needed if it has an `explicit operator bool()`.

include/range/v3/algorithm/ends_with.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    struct ends_with_fn
+    {
+    private:
+        template<typename I0, typename S0, typename I1, typename S1,
+            typename C, typename P0, typename P1>
+        constexpr /*c++14*/ ends_with_result<I0, I1> nocheck(
+            I0 begin0, S0 end0, I1 begin1, S1, C pred, P0 proj0, P1 proj1) const
+        {
+            for(; begin0 != end0; ++begin0, ++begin1)
```
johelegp: Note that I removed the bounds check for the second range in this code I got from `equal.hpp`. Is that okay? The ranges are supposed to be of the same size.

include/range/v3/algorithm/ends_with.hpp
```diff
@@ -0,0 +1,92 @@
...
+                if(!invoke(pred, invoke(proj0, *begin0), invoke(proj1, *begin1)))
+                    return {false, std::move(begin0), std::move(begin1)};
+            return {true, std::move(begin0), std::move(begin1)};
```
johelegp: Similarly here, `true` is unconditionally returned rather than confirming that the bounds were reached.

include/range/v3/algorithm/ends_with.hpp
cjdb: I'd be much more comfortable if we checked both ranges.

include/range/v3/algorithm/result_types.hpp
```diff
@@ -51,6 +51,39 @@ namespace ranges
             }                                                                                   \
             /**/
 
+        #define RANGES_ALGO_RESULT_AUX_BOOL_2(C, T1, M1, T2, M2)                                \
+            constexpr explicit operator bool() const noexcept                                   \
+            {                                                                                   \
+                return value;                                                                   \
+            }                                                                                   \
```
ericniebler: I would prefer not to have result structs that are implicitly convertible to their actual result. I appreciate the attempt to support the simple common use case while also returning all the computed information, but in this case, it might be best for `ends_with` to simply return `bool`.

Maybe there should be a `find_at_end` that returns all the information, and `ends_with` that just returns the `bool` result. Just spitballing here.

Issue #1147, opened 2019-04-20
view_interface empty() for Sized && !ForwardRange


tower120:
`view_interface` does not provide `empty()` for Sized InputRange, if that range does not provide `empty()` itself.

https://wandbox.org/permlink/iWcyjq3fE9RFUezD
```cpp
    auto g = view::generate_n([](){ return 1;}, 4);
    g.size();  // ok.
    g.empty(); // error.
```

`view_interface` can provide `empty()` for this cases, by
addding empty() overload for SizedRange && !ForwardRange into view_interface:
```cpp
bool empty(){
   return ranges::size(derived()) == 0;
}
```

PR #1149, opened 2019-04-21
view_interface empty() for SizedRange

tower120:
Implement https://github.com/ericniebler/range-v3/issues/1147

### DIFF
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -105,7 +105,7 @@ namespace ranges
                     Cardinality < 0 && Cardinality != infinite &&
                     !ForwardRange<D const>() &&
                     SizedRange<D const>())>
-            RANGES_CXX14_CONSTEXPR bool empty() const
+            constexpr bool empty() const
                 noexcept(noexcept(bool(ranges::size(std::declval<D const &>()) == 0)))
             {
                 return ranges::size(derived()) == 0;
--- a/include/range/v3/view_interface.hpp
+++ b/include/range/v3/view_interface.hpp
@@ -91,6 +91,26 @@ namespace ranges
                 return Cardinality == 0;
             }
             template<typename D = Derived,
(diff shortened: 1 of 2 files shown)


include/range/v3/view_interface.hpp
```diff
@@ -90,6 +90,26 @@ namespace ranges
             {
                 return Cardinality == 0;
             }
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(
+                    Cardinality < 0 && Cardinality != infinite &&
+                    !ForwardRange<D>() &&
+                    SizedRange<D>())>
+            RANGES_CXX14_CONSTEXPR bool empty()
```
johelegp: This qualifies for C++11 `constexpr`.

include/range/v3/view_interface.hpp
```diff
@@ -90,6 +90,26 @@ namespace ranges
...
+                noexcept(noexcept(bool(ranges::size(std::declval<D &>()) == 0)))
+            {
+                return ranges::size(derived()) == 0;
+            }
+            template<typename D = Derived,
+                CONCEPT_REQUIRES_(
+                    Cardinality < 0 && Cardinality != infinite &&
+                    !ForwardRange<D const>() &&
+                    SizedRange<D const>())>
+            RANGES_CXX14_CONSTEXPR bool empty() const
```
johelegp: And this too.

include/range/v3/view_interface.hpp
johelegp: Actually, not this one due to the lack of `const`.

tower120: ??

johelegp: The other `empty` overload you added can be `constexpr` in C++11, so you should replace `RANGES_CXX14_CONSTEXPR` for `constexpr`. Not this one, though.

Issue #932, opened 2018-11-02
Computing result type with type helpers for views.


tower120:
Need some way to **PRE**compute result view type. 

Having concrete (not auto) type as function return, allow to move implementation from header to source file (this is important for breaking circular dependency). 
Example:

```cpp
// CarTypes.h
class CarTypes {
protected:
    using m_wheels_t = std::vector<std::unique_ptr<Wheel>>;
public:
    // this not work!
    // using wheels_t = ranges::indirect_view< m_wheels_t>;  

    // here, need helper ranges::indirect_view_t, instead of this:
    using wheels_t = ranges::indirect_view< ranges::view::all_t<m_wheels_t > >;
}
```

```cpp
// Car.h
class Car : public CarTypes {
public:
    wheels_t wheels() {  // return concrete type, not auto. Now it is possible to move wheels() to .cpp file
	return m_wheels | ranges::view::indirect;     // "return {m_wheels};"  this would be even nicer
    }  
private:
  m_wheels_t m_wheels;
};
```

```cpp
// CarsHelperTypes.h
// imagine that we can't include Car.h here, due to cyclic dependency
class Car;

// separating all "interface" types to separate class and header can make life sometimes easier.
// This allows to work with types (#include "xxxxxTypes.h") without need of including implementation.
struct CarsHelperTypes{
       const auto fn = [](auto* car){
          return yield_from(car->wheels());
      };
     using cars_t = std::vector<Car*>;
     using all_wheels_t = for_each_view_t<cars_t , decltype(fn)>;
};

// CarsHelper.h
struct CarsHelper : CarsHelperTypes{
   cars_t cars;
   all_wheels_t all_wheels();
}
```

```cpp
// CarsHelper.cpp
#include " CarsHelper.h"

// including implementation now!
#include "Car.h"

CarsHelperTypes::all_wheels_t CarsHelper::all_wheels(){
      return view::for_each(cars, fn);
}
```

It is not always possible to separate return types from implementation like this (thou I think often enough to have practical use), but when it is - we can avoid slowness of  `any_view`.

So, to summarize:
 * It would be good, if each view function return exactly corresponding type; like: `ranges::view::indirect -> ranges::indirect_view`  (maybe do decaying/other type conversion directly in indirect_view? Utilize c++17 class template argument deduction to get rid of functions-makers at all? *Or whatever reason for these helper functions to exists - pipeability?*)
[...]

PR #1155, opened 2019-04-23
View type composition

tower120:
This PR implements this https://github.com/ericniebler/range-v3/issues/932#issuecomment-483304960

Interface is following:
```cpp
using List = std::vector<std::unique_ptr<int>>;
List list;

// all views in chain unary constructible
using Get = compose_view<List, indirect_view, const_view >;
Get get(){
    return list | view::indirect | view::const_;
}
Get get2(){
    return Get{list};
}
```
non-unary constructible:
```cpp
struct Fn{
    const int& operator()(const int& i) const { return i; }
};
using Get = compose_view<List, compose_bind<transform_view, Fn>::type >;
Get get(){
    return list | view::transform(Fn{});
}
/*Get get2(){
    return Get{list};  // impossible, for now
}*/
```


tower120:
View composition without data source attached (aka `view::indirect | view::tail`) possible with this https://github.com/ericniebler/range-v3/issues/1158 implemented :

```cpp
template<class Rng>
struct filter_view : compose_view<Rng, drop_view, const_view>{
    filter_view() = default;
    explicit filter_view(Rng&& rng, int n)
        : filter_view::compose_view(std::move(rng) | view::drop(n) | view::const_) {}
};

using Get = compose_view<List, filter_view>;
Get get(){
    return list | pipe<filter_view>(1);
}
```

johelegp:
What problems do you perceive with regards to the types of `view::X` differing from `X_view`? I also find it unfortunate that `range_fwd.hpp` does not have declarations of all views, which would see increased use with this feature.

tower120:
> What problems do you perceive with regards to the types of view::X differing from X_view? 

I'm not sure I understand you. `view::X` is not a type.

>  I also find it unfortunate that range_fwd.hpp does not have declarations of all views

Don't understand, either...

----

Motivation example for `compose_view` is following:
```cpp
// .hpp
class Widget;   // forward declare only

using List = std::vector<std::unique_ptr<Widget>>;
List list;

using Get = compose_view<List, indirect_view, const_view >;
Get get();  // it is impossible to have auto return type function with implementation in .cpp !!

// --------------------------------------------------

// .cpp
#include "Widget.h"   // now include concrete class

Get get(){
   return Get{};
}
```

- breaking cyclic "include" dependency.
- moving implementation to .cpp . Have sense with parametric (non-unary) views drop / drop_last / etc..

johelegp:
Yeah. I understand the benefits and would like to have such functionality available.

> > What problems do you perceive with regards to the types of view::X differing from X_view?
> 
> I'm not sure I understand you. `view::X` is not a type.

I mean, the composition should use `view::X_fn` rather than `X_view` so that even when the return type of `view::X` isn't a specialization of `X_view`, things work as expected.

> > I also find it unfortunate that range_fwd.hpp does not have declarations of all views
> 
> Don't understand, either...

Just some wrong thought. Since a benefit of this feature is improved compile times for users, I thought about using it with forward declarations of the views for further improvement. But then the header wouldn't be self-contained. Please, dismiss this.

tower120:
> I mean, the composition should use view::X_fn rather than X_view so that even when the return type of view::X isn't a specialization of X_view, things work as expected.

So... you ask how it works (`drop_view` in type, while `drop` in construction)?

johelegp:
No. Consider your example [above](https://github.com/ericniebler/range-v3/pull/1155#issuecomment-487009531).

> ```c++
> template<class Rng>
> struct filter_view : compose_view<Rng, drop_view, const_view>{
>     filter_view() = default;
>     filter_view(Rng&& rng, int n)
>       : filter_view::compose_view(std::forward<Rng>(rng) | view::drop(n) | view::const_) {}
> };
> 
> using Get = compose_view<List, filter_view>;
> Get get(){
>     return list | pipe<filter_view>(1);
> }
> ```

If `Rng` is `std::string_view`, `view::drop(n)` returns `subrange<iterator_t<std::string_view>, sentinel_t<std::string_view>>` and not `drop_view<all_t<std::string_view>>`. That would result in a compile-time error.

tower120:
@johelegp Now looking at `view/drop.hpp ` I see what you mean. There is special case - and that special case handled not by view, but by "adaptor closure object" (drop_fn). I have feeling that is wrong... Adaptor closure object must be view "wrapper"  - and nothing more... All that logic for special cases must be handled by view itself... It's counter-intuitive after all! How it come that `drop_view(my_string_view, 2)` and `my_string_view | drop(2)` is a different things?

_But in case - if that is not view implementation mistake - compose_view can work with X_fn instead of X_view..._

johelegp:
> _But in case - if that is not view implementation mistake - compose_view can work with X_fn instead of X_view..._

Yes, that's allowed. So that change would be for the best. Look at the description of the various `std::ranges::view`s to see why it can be useful for the types to differ.

tower120:
@johelegp at a first glance - looks like a pure madness to me! I don't see why http://www.eelis.net/c++draft/range.adaptors#range.all can't have `all_view` which would do all that dispatching. why `/*undefined type*/` range adaptor object do all that, instead?

I need to reconsider `compose_view` design...

That's not that pretty now:
```cpp
using Get = compose_view<List, decltype(view::indirect), decltype(view::const_) >;
```

tower120:
There is probably more "sane", standard compatible (I hope!) way of doing this.

```cpp
using Get = compose_view<List, view::indirect, view::const_ >;
```

By working with `template<auto&...>` http://coliru.stacked-crooked.com/a/d07c1f442bb1b90f

johelegp:
That's great! But it doesn't work on C++14. If you're going to require `decltype`, might as well do it on the outside so that it's only needed once.
```C++
using Get = compose_view<List, decltype(view::indirect), decltype(view::const_) >;
using Get = decltype(compose_view<List>(view::indirect, view::const_));
```

tower120:
@johelegp Mmmm... Can't it be C++17 only?

johelegp:
You'd have to ask @ericniebler.

tower120:
Сontinued here https://github.com/ericniebler/range-v3/pull/1172

PR #1160, opened 2019-04-25
Merge from origin/master

ericniebler:


include/range/v3/iterator/operations.hpp
```diff
@@ -87,7 +87,9 @@ namespace ranges
         {
             if constexpr(SizedSentinel<S, I>)
             {
-                const auto d = bound - i;
+                if(0 == n)
+                    return 0;
+               const auto d = bound - i;
```
CaseyCarter: I think this is missing a space of indent.

PR #1163, opened 2019-04-27
Improve check of -Wno-* compiler flags

mymedia2:
The GCC documentation says that unrecognized warning options in the `-Wno-*` form do not cause errors unless other diagnostics are issued. And thus unsupported flags may be appended. This results in a failed
build when there are some false positive warnings.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

### DIFF
--- a/cmake/ranges_flags.cmake
+++ b/cmake/ranges_flags.cmake
@@ -8,10 +8,14 @@
 
 # Compilation flags
 include(CheckCXXCompilerFlag)
-macro(ranges_append_flag testname flag ${ARGN})
-    check_cxx_compiler_flag("${flag} ${ARGN}" ${testname})
+macro(ranges_append_flag testname flag)
+    # As -Wno-* flags do not lead to build failure when there are no other
+    # diagnostics, we check positive option to determine their applicability.
+    # Of course, we set the original flag that is requested in the parameters.
+    string(REGEX REPLACE "^-Wno-" "-W" alt ${flag})
+    check_cxx_compiler_flag(${alt} ${testname})
     if (${testname})
-        add_compile_options(${flag} ${ARGN})
+        add_compile_options(${flag})
     endif()
 endmacro()
 


cmake/ranges_flags.cmake
```diff
@@ -8,10 +8,14 @@
 
 # Compilation flags
 include(CheckCXXCompilerFlag)
-macro(ranges_append_flag testname flag ${ARGN})
-    check_cxx_compiler_flag("${flag} ${ARGN}" ${testname})
```
ericniebler: Looks like we lost support for optional additional arguments. (I'm assuming that's what the `$(ARGN)` is doing here, but I'm Not A CMake Expert). Was that intentional?

mymedia2: To be honest, I have not figured out how this works. I thought there was the ranges_append_flag macro with three parameters `testname`, `flag` and third one had name consisted of empty string as the ARGN variable was not defined. This third parameter was not used within the macro because `${ARGN}` was referenced directly, without nesting.
If I understand something wrong, I can change the string back, but I do not know what the test with the alt variable would look like.
In fact, everywhere the macro is called with two arguments.

ericniebler: The `${ARGN}` parameter comes from @gnzlbg in 577f2d62. I wonder if he would be willing to chime in here.

gnzlbg: IIRC some flags take optional arguments, e.g., [clang `-X` flag](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-x), and the arguments need to be part of the test for the flag.

mymedia2: Well, but there are other flags that accept their values with equal sign. Examples: `-std=c++11`, `-DLEVEL=3`.
In any case, it seems that definition of the ranges_append_flag macro was slightly incorrect. New variables are created without braces so the first line of the macro should look like

    macro(ranges_append_flag testname flag ARGN)

gnzlbg: > Well, but there are other flags that accept their values with equal sign.

Sure, but from this macro POV, those just take a single argument. It's the flags that take multiple space-separated arguments that `${ARGN}` was _probably_ intended to support.

> In any case, it seems that definition of the ranges_append_flag macro was slightly incorrect. 

IIRC `${ARGN}` expands to optional arguments inside a macro, so I don't think it needs to be in the `macro()` definition at all. Changing it to `ARGN` means that only a single extra argument would be supported, but `${ARGN}` is "variadic". 

In any case, if you don't need to support any of these flags, the best would probably be to just drop the `${ARGN}` thing completely. If you need space-separated flags, maybe `ranges_append_flag(... "-X ...")` would work anyways.

mymedia2: Thank you for the explanation. I think we can leave the macro having no extra variadic parameters because, as you mentioned, for space-separated values one can use single argument. Like this:

    ranges_append_flag(TEST_NAME "-X arg1 arg2 arg3")

Moreover, `${ARGN}` list was expanded with a semicolon in call of the check_cxx_compiler_flag function.

One thing I can improve on is to add to the commit description that the ranges_append_flag macro no longer accepts optional arguments, and it is desirable to use a single argument with explicit delimiters.

ericniebler: > One thing I can improve on is to add to the commit description that the ranges_append_flag macro no longer accepts optional arguments, and it is desirable to use a single argument with explicit delimiters.

Should I continue waiting for this, or should I merge now?

ericniebler:
Thank you!

PR #1166, opened 2019-04-28
use modern concept definition

marehr:
The standard only allows the following syntax 

```c++
template <typename t>
concept AnyType = true; // gcc >= 9 and clang
```

but ranges currently uses 

```c++
template <typename t>
concept bool AnyType = true; // only way to define concepts gcc<9
```

-----

clang will not allow the additional `bool`.

This PR will add the concept definition without the `bool` and uses it in the same manner as it is already handled in the meta library, see

https://github.com/ericniebler/range-v3/blob/75d96cfeabf6366012090594382bda635b6cfb38/include/meta/meta_fwd.hpp#L138-L153

Open Questions:
* Would it also be okay to use `META_CONCEPT` instead if redefining it? Or do you want to make the definitions of the meta library independent of the ones in ranges?
* Do you have a better name for `CPP_CXX_CONCEPT`? Or is it okay?

### DIFF
--- a/include/concepts/concepts.hpp
+++ b/include/concepts/concepts.hpp
@@ -224,13 +224,13 @@ CPP_PP_IGNORE_CXX2A_COMPAT_BEGIN
 #ifdef CPP_DOXYGEN_INVOKED
 #define CPP_PP_DECL_DEF_IMPL(TPARAM, NAME, ARGS, ...)                           \
     CPP_PP_CAT(CPP_PP_DEF_, TPARAM)                                             \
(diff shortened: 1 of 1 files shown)


CaseyCarter:
`CPP_CXX_CONCEPT` is a bit awkward given that there's already a definition for `CPP_CXX_CONCEPTS`. In any case, I suggest simply using `META_CONCEPT` directly. `concepts/concepts.hpp` includes `meta` via `concepts/type_traits.hpp`, so there's already a dependency between the libraries.

marehr:
@CaseyCarter Thank you! I updated the PR to use `META_CONCEPT` directly.

ericniebler:
Thank you!

PR #1170, opened 2019-04-30
Add cpp20::ref_view

johelegp:


ericniebler:
The valgrind failures are troubling. Clearly they are unrelated to your changes. Bears some investigation. I'll get to it as I can.

ericniebler:
@CaseyCarter investigated and reports that the valgrind failures are due to a well-known bug in valgrind: https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1501545. Unclear why we're only seeing it now.

include/range/v3/view/ref.hpp
```diff
@@ -135,6 +135,13 @@ namespace ranges
         /// \ingroup group-views
         RANGES_INLINE_VARIABLE(view<ref_fn>, ref)
     }
+
+    namespace cpp20
+    {
+        CPP_template(typename Rng)(
+            requires std::is_object<Rng>::value)
+        using ref_view = ranges::ref_view<Rng>;
```
CaseyCarter: Why not simply `using ranges::ref_view;`? A *using-declaration* here - instead of an *alias-declaration* - would preserve the capability to use CTAD.

johelegp: Good idea. I just followed what other headers did. The only difference seems to be the addition of requires clauses. Why not put those on the classes themselves and do as @CaseyCarter says, @ericniebler?

johelegp: Well, it's not like that'd do anything without concepts in the language. I'd guess that Eric chose having simulation of requires clauses over CTAD.

ericniebler: Yes, CTAD and concept emulation are at odds. I haven't quite worked out what the general policy should be. Perhaps @johelegp is right and the `requires` clauses should be on the classes themselves. Needs thought. For now, it's probably best to be consistent with what the rest of the library does -- as you're doing here -- and I can fix it all at once.

PR #1172, opened 2019-04-30
View type composition v2

tower120:
This is continuation of https://github.com/ericniebler/range-v3/pull/1155 . Work with adaptor closure objects, instead of view classes.
C++17 only. Requires `template<auto&...>`

Add `compose_view<class, auto&...>`, `compose_bind<auto&, class...>` and `compose_bind_static<auto&, auto...>`

Type composition:
```cpp
using Get = compose_view<
    List,
    view::indirect,
    compose_bind<view::slice, int, int>,
    compose_bind<view::transform, Fn>,
    compose_bind_static<view::drop, 1>>;  // can accept trivial classes with C++20
```

Construction:
```cpp
Get get_a(){
    // TODO: reject conversion constructor if compose_view contain compose_bind_static?
    return list | view::indirect | view::slice(1, 3) | view::transform(Fn{}) | view::drop(1);
}
Get get_b(){
    return Get{list, view::slice(1, 3), view::transform(Fn{})};
}
Get get_c(){
    return Get{list, 1, 3, Fn{}};
}
```

Bonus:
```cpp
inline static auto fn = view::transform(Fn{}) | view::drop(1);
using Get = compose_view<List, view::indirect, fn>;
Get get(){
      return Get{list};
}
```


ericniebler:
I don't feel the need for this functionality, and this is a rather complex interface that only meta-programmers would be comfortable with. `decltype` works fine for this.

PR #1174, opened 2019-04-30
Reorder and split some views' documentation

johelegp:


doc/index.md
```diff
@@ -588,7 +588,9 @@ provides, and a blurb about how each is intended to be used.
...
 <DT>\link ranges::view::split_fn `view::split`\endlink</DT>
-  <DD>Given a source range and a delimiter specifier, split the source range into a range of ranges using the delimiter specifier to find the boundaries. The delimiter specifier can be a value, a subrange, a predicate, or a function. The predicate should take an single argument of the range's reference type and return true if and only if the element is part of a delimiter. The function should accept current/end iterators into the source range and return `make_pair(true, iterator_past_the_delimiter)` if the current position is a boundary; otherwise, `make_pair(false, cur)`. The delimiter character(s) are excluded from the resulting range of ranges.</DD>
+  <DD>Given a source range and a delimiter specifier, split the source range into a range of ranges using the delimiter specifier to find the boundaries. The delimiter specifier can be a value or a subrange. The delimiter character(s) are excluded from the resulting range of ranges.</DD>
```
CaseyCarter: * "a value or a subrange" was slightly confusing *before* we had a library component named `subrange`. "An element or a range of elements" seems better to me.
* "The delimiter character(s)" could more precisely be "The elements matching the delimiter."

doc/index.md
```diff
@@ -588,7 +588,9 @@ provides, and a blurb about how each is intended to be used.
...
+<DT>\link ranges::view::split_when_fn `view::split_when`\endlink</DT>
+  <DD>Given a source range and a delimiter specifier, split the source range into a range of ranges using the delimiter specifier to find the boundaries. The delimiter specifier can be a predicate or a function. The predicate should take a single argument of the range's reference type and return true if and only if the element is part of a delimiter. The function should accept current/end iterators into the source range and return `make_pair(true, iterator_past_the_delimiter)` if the current position is a boundary; otherwise, `make_pair(false, cur)`. The delimiter character(s) are excluded from the resulting range of ranges.</DD>
```
CaseyCarter: * "should return true" => "should return `true`"
* "The function should accept current/end iterators" is wrong, `split_view` passes a sentinel rather than an end iterator. "otherwise, `make_pair(false, cur)`" is somewhat misleading: the iterator in the returned `pair` is ignored when the `bool` is `false`. (`split_view` *should* expect an `optional<iterator_t<R>>`, but that's a design change and not really germane to this review.) I think we should reformulate this sentence as:
  > The function should accept an iterator and sentinel indicating the current position and end of the source range and return `std::make_pair(true, iterator_past_the_delimiter)` if the current position is a boundary; otherwise `std::make_pair(false, ignored_iterator_value)`.
* Ditto "delimiter character(s)"

PR #1175, opened 2019-05-01
Workaround valgrind Travis failures

CaseyCarter:
`valgrind` doesn't grok `RDRAND` (https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1501545), so we must explicitly initialize `std::random_device{"/dev/urandom"}` to avoid using the instruction in valgrind runs on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -251,8 +251,13 @@ install:
   # Workaround for valgrind bug: https://bugs.kde.org/show_bug.cgi?id=326469.
   # It is fixed in valgrind 3.10 so this won't be necessary if someone
   # replaces the current valgrind (3.7) with valgrind-3.10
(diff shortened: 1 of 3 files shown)


ericniebler:
Oh, Travis. _\<shakes head sadly\>_

```
The following packages have unmet dependencies:
 clang-5.0 : Depends: libjsoncpp0 (>= 0.6.0~rc2) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

CaseyCarter:
Weirdness. I pushed a commit to fix this hours ago. It's on my branch (https://github.com/CaseyCarter/range-v3/tree/rdrand) but didn't show up in this PR. New PR incoming.

Issue #1158, opened 2019-04-24
basic adaptor closure object for user views


tower120:
Views are not pipeable by themselves. Block of code, which make view pipeable is quite verbose too...

What about having some `pipe<View>` helper:

```cpp
    template<template<class...> class V>
    struct wrap_view_{
        using Self = wrap_view_<V>;

        template<class ...Args>
        static auto bind(Self self, Args&&...args)
        RANGES_DECLTYPE_AUTO_RETURN
        (
            make_pipeable(std::bind(self, std::placeholders::_1, args...))
        )

        template<typename Rng, class ...Args>
        auto operator()(Rng&& rng, Args&& ...args) const
        {
            // >=C++17 Class template argument deduction
            return V(view::all(std::forward<Rng>(rng)), std::forward<Args>(args)...);
        }
    };

    template<template<class> class V>
    using pipe_t = view::view<wrap_view_<V>>;

    // >=C++14 only
    template<template<class> class V>
    pipe_t<V> pipe;
```

Usage:

```cpp
   list | pipe<my_view>(1);
```
```cpp
   pipe<my_view> my;
   ...
   list | my(1);
```

This will fit not all, but yet many views...

PR #1177, opened 2019-05-04
basic_range_adaptor

tower120:
Adds `view::basic_range_adaptor`, which makes range adaptor object for any CTAD friendly view. Range must be first constructor parameter, for sake of pipeability.
```cpp
template<class Rng>
struct filter_view{
    filter_view(Rng rng, int value)
};
auto filter = view::basic_range_adaptor<filter_view>;
// or 
view::basic_range_adaptor_t<filter_view> filter;
```
Usage:
```cpp
list | view::basic_range_adaptor<filter_view>(1);
list | filter(1);
```
Requires C++17 class template argument deduction.

Implements https://github.com/ericniebler/range-v3/issues/1158 .


tower120:
Maybe it should be called `view_range_adaptor`... Because similar helper can be done to simplify making range adaptors from closures:
```cpp
auto trim = 
view::range_adaptor([](auto&& rng /* pre-wrap in view::all ?*/, int front, int back){
   return view::drop(rng, front) | view::drop_last(back);
});
```

ericniebler:
Recently, I have tended to shy away from extremely generic library helpers like this because they are simply unable to provide useful diagnostics when folks try to misuse them.

Yes, spamming out view adaptor and view adaptor closure types is verbose and annoying, but each one can have custom `requires` clauses and give helpful error messages for types that don't satisfy the requirements.

PR #1179, opened 2019-05-06
ranges::base

tower120:
### Abstract

This PR introduces `ranges::base` standalone function. It provides machinery for working with iterator/range base() chains (`base().base().base()....`).

### Motivation

```cpp
auto list = vec | view::transform(&Data::i) | view::take_exactly(3);
list.begin().base().base() != vec.begin();
```
You have to "guess" count of base() calls to the desired layer of iterator adaptor.
```cpp
auto list = vec | view::transform(&Data::i) | view::take_exactly(3);
base<iterator_t<Vec>>(list.begin()) == vec.begin();
```
Usage with algorithms:
```cpp
auto list  = vec | view::transform(&Data::i) | view::take_exactly(3);
iterator i = base<iterator>(find(list, 10));
```

### Interface

```cpp
// iterator

ragnes::base(iter);                     // iter.base()
ragnes::base<3>(iter);                  // iter.base().base().base()
ragnes::base<iterator_t<Range>>(iter);  // iter.base()...
ragnes::base<Range>(iter);              // ragnes::base<iterator_t<Range>>(iter)
ragnes::base(iter, range);              // ragnes::base<decltype(range)>(iter)

// range

ragnes::base(range);
ragnes::base<3>(range);
ragnes::base<List>(range);

```

### DIFF
--- a/include/range/v3/base.hpp
+++ b/include/range/v3/base.hpp
@@ -1,0 +1,144 @@
+/// \file
+// Range v3 library
+//
+//  Copyright Andrey Diduh 2019
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+
+#ifndef RANGE_V3_BASE_HPP
+#define RANGE_V3_BASE_HPP
+
+#include <range/v3/range_concepts.hpp>
+#include <range/v3/view/all.hpp>
+
+namespace ranges
+{
+    inline namespace v3
+    {
+        namespace details
+        {
+            template<class V>
+            RANGES_CXX14_CONSTEXPR auto base_loop(V&& v, std::integral_constant<int, 1>)
+            RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+            (
+                v.base()
+            )
+            template<int N, class V>
+            RANGES_CXX14_CONSTEXPR auto base_loop(V&& v, std::integral_constant<int, N>)
+            RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+            (
+                base_loop( v.base(), std::integral_constant<int, N-1>{})
+            )
+            template<int N, class V>
+            RANGES_CXX14_CONSTEXPR auto base_loop(V&& v)
+            RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+            (
+                base_loop(std::forward<V>(v), std::integral_constant<int, N>{})
+            )
+
+            template<typename Target, bool found>
+            struct loop_to_t;
+
+            template<typename Target, class V>
+            RANGES_CXX14_CONSTEXPR auto loop_to(V&& v)
+            RANGES_DECLTYPE_AUTO_RETURN_NOEXCEPT
+            (
+                loop_to_t<
+                    Target,
+                    std::is_same< detail::decay_t<V>, detail::decay_t<Target> >::value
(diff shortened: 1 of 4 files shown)


ericniebler:
I'm not a fan of this functionality. I don't see enough of a need, and I never want to call `.base()` an arbitrary number of times.

tower120:
>  I don't see enough of a need

The main motivation for this - algorithms:
  1) There are things, that close to impossible to do with projections.
      ```cpp
       // Data has following structure
       // | Header (16 bytes) | Payload(128byte) | ....
       // Seeking payload with specific header.
      using Iterator = iterator_t< std::vector<char> >;
      std::optional<Iterator> lookup(
         const std::vector<char>& raw_data, const std::vector<char>& required_header)
      {
         auto chunks  = raw_data | view::sliding(16+128);
         auto headers = 
           chunks 
           | view::forward_range
           | view::transform([](auto&& rng){ return rng | view::take(16); });

         auto founded = find_if(headers, [&](auto&& rng){ return equal(rng, required_header); });
         if (founded == headers.end()) return {};
         return {base<Iterator>(founded) + 16};
     }
      ```
  2) You already have a view chain that do transformation , and you don't want to repeat yourself.
      ```cpp
      Events = std::list<IEvent>;
      Events events;

      struct SyncPoint : IEvent{
          SyncFence& sync_fence();
          const SyncFence& sync_fence() const;
      };

     struct SyncFence{
          std::string text;
     }

      // This part already exists and in use.
      // -------------
      template<class T>
      auto get_only = 
         view::transform([](IEvent& e){ return dynamic_cast<T*>(&e); }) 
         | view::remove(nullptr)
         | view::indirect;

     auto get_sync_fences = 
         get_only<SyncPoint> 
         | view::transform([](SyncPoint& sync_point) -> SyncFence& {
              return sync_point.sync_fence();
          });
      // -------------

      // with base<>
      auto sync_fences = events | get_sync_fences ;
      auto founded = base<Widgets>(find(sync_fences, "red", &SyncFence::text));

     // with projection
     // It does not look much longer, but all that transformation chain
     // has substantial cognitive burden. While writing that transformer -
     // you'll forgot what exactly you wanted to search.
     auto founded = find_if(widgets, [](const IEvent& e) -> bool {
        const SyncPoint* p = dynamic_cast<const SyncPoint*>(&e);
        if (!p) return false;
        const SyncFence& fence = p->sync_fence();
        return (fence.text == "red");
      });
      ```

> I never want to call .base() an arbitrary number of times.

The core part of this PR is only:
```cpp
ranges::base<Iterator>(iter);
```
Everything else actually is auxiliary, and can be hided/removed.

tower120:
added example 3.

ericniebler:
This PR is now badly out of date. I never was able to convince myself that this problem justifies any solution, or that this solution is it. Do you still want this, @tower120?

tower120:
My last line of defense is "`ranges::base` = `better projections` for lookup algorithms".

That's probably main use-case for `ranges::base()`. _Previously I thought that `ranges::base()` can be used inside mutable algorithms/actions (like `actions::remove( rng1, view::not_unique)`), but abandoned that idea._

 Important property of projection is that it can't skip elements, nor change their order. While view can.
```cpp
using Pair =std::pair<int, int>; 
std::vector<Pair> pairs;

auto iter1 = ranges::find(pairs | view::sliding(256) | view::each_range(view::take(4)).base(pairs);
```
You can't do that at all with projections, and you can't do that efficiently with predicates. And you definitely can't have that elegant solution with manual loop.

N.B.
```cpp
auto each_range = [](auto view){
   // view::forward_iter = view::iota(view.begin(), view.end())    kinda...
   return view::forward_iter | view::transform([](auto&& rng){ return *rng | view; });
};
```

I still think this is useful. But if you think it has too narrow application, close it.

Issue #1164, opened 2019-04-27
about-to-be-released GCC9 ICEs on 1.0-branch


h-2:
This simple example ICEs on GCC9 with `-fconcepts` (works on 8.3 though):
```cpp
#include <range/v3/view/reverse.hpp>
#include <vector>
#include <iostream>

int main()
{
    std::vector v{0, 1, 2, 3, 4};
    for (auto e : v | ranges::view::reverse)
    {
        std::cout << e << '\n';
    }
    return 0;
}
```

Here is the GCC problem report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89953

Maybe you can underline the importance of the problem there? Otherwise the 1.0 release will be broken on upcoming GCC (or need elaborate workarounds).

Thanks!

CaseyCarter:
> Maybe you can underline the importance of the problem there?

I suggest putting `[9 regression]` at the beginning of the issue title; that should raise visibility and light a fire under them.

ericniebler:
Did #1185 fix this for you? (I'm having trouble building gcc-9 on my Macbook so I can't easily test this.)

h-2:
> Did #1185 fix this for you? (I'm having trouble building gcc-9 on my Macbook so I can't easily test this.)

Yep, that's why I made the PR!

PR #1185, opened 2019-05-10
Make v1.0-branch build with GCC9

h-2:
The first commit makes the test suite build with gcc9 (independent of `-fconcepts`). I think the changes *shouldn't* be necessary and that it is probably a GCC bug, but it only affects test suite code and not library code.

The second commit is a workaround for #1164. Note that although the example in #1164 seems rather trivial, the issue was not detected by the test suite. One should maybe create a test case, but I thought it was more important to get a fix in now since we have a bunch of other ICEs in GCC9 that we can't tackle without #1164 being out of the way.

P.S: I thought about creating a MACRO for the workaround, but since it involves code inside the other macros, it would involve more code than necessary. I can still change this or replace the alias template entirely with the variable template since one doesn't really need both. But the solution below was the one with the smallest diff that I could come up with.

### DIFF
--- a/include/range/v3/range/access.hpp
+++ b/include/range/v3/range/access.hpp
@@ -90,6 +90,11 @@ namespace ranges
             template<typename R>
             using impl = impl_<HasMemberBegin<R>>;
 
+#ifdef RANGES_WORKAROUND_GCC_89953
+            template<typename R>
+            static constexpr impl<R> impl_v{};
+#endif
+
         public:
             template<typename R, std::size_t N>
             void operator()(R (&&)[N]) const = delete;
@@ -101,9 +106,15 @@ namespace ranges
             }
 
             template<typename R>
+#ifdef RANGES_WORKAROUND_GCC_89953
+            constexpr auto CPP_fun(operator())(R &&r) (const
+                noexcept(noexcept(impl_v<R>((R &&) r)))
+                requires (HasMemberBegin<R> || HasNonMemberBegin<R>))
+#else
             constexpr auto CPP_fun(operator())(R &&r) (const
                 noexcept(noexcept(impl<R>{}((R &&) r)))
                 requires (HasMemberBegin<R> || HasNonMemberBegin<R>))
+#endif
             {
                 return impl<R>{}((R &&) r);
             }
@@ -223,6 +234,11 @@ namespace ranges
             template<typename R>
             using impl = impl_<HasMemberEnd<R>>;
 
+#ifdef RANGES_WORKAROUND_GCC_89953
+            template<typename R>
+            static constexpr impl<R> impl_v{};
+#endif
+
         public:
             template<typename R, std::size_t N>
             void operator()(R (&&)[N]) const = delete;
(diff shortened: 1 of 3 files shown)


test/test_utils.hpp
```diff
@@ -162,7 +162,7 @@ struct test_range_algo_1
         using S = meta::_t<sentinel_type<I>>;
         using R = decltype(algo_(begin, end, rest...));
         Algo algo = algo_;
-        auto check_algo = [=](function_ref<void(R)> const & check)
+        auto check_algo = [algo, begin, end, rest...](function_ref<void(R)> const & check)
```
CaseyCarter: Should we init-capture `algo = algo_` here and remove line 164? We don't need to support C++11 on v1.0-beta.

test/test_utils.hpp
```diff
@@ -196,7 +196,7 @@ struct test_range_algo_2
         using S2 = meta::_t<sentinel_type<I2>>;
         using R = decltype(algo_(begin1, end1, begin2, end2, rest...));
         Algo algo = algo_;
-        return checker<R>{[=](function_ref<void(R)> const & check)
+        return checker<R>{[algo, begin1, end1, begin2, end2, rest...](function_ref<void(R)> const & check)
```
CaseyCarter: Ditto init-capture.

include/range/v3/range/access.hpp
```diff
@@ -102,7 +105,7 @@ namespace ranges
 
             template<typename R>
             constexpr auto CPP_fun(operator())(R &&r) (const
-                noexcept(noexcept(impl<R>{}((R &&) r)))
+                noexcept(noexcept(impl_v<R>((R &&) r)))
```
CaseyCarter: This change pessimizes compile time. I'd prefer to see it conditional on `#ifdef RANGES_WORKAROUND_GCC_FOO`, where `FOO` is either a GCC bug number or `UNKNOWN1` (with a corresponding range-v3 issue to minimize and file a bug). `RANGES_WORKAROUND_GCC_FOO` should be defined in `detail/config.hpp` only when `__GNUC__ >=9 && defined(__cpp_concepts)`.

h-2:
I have updated the PR!

h-2:
on an unrelated note: you can setup travis so that it cancels ongoing builds if new builds are triggered by a new push to the same branch.

CaseyCarter:
> you can setup travis so that it cancels ongoing builds if new builds are triggered by a new push to the same branch.

Yes, we have [auto cancellation](https://travis-ci.org/ericniebler/range-v3/settings) enabled. IIRC it only cancels builds that haven't yet started? I don't recall the conditions exactly, but it's something more complicated than simply killing builds when there's a push.

h-2:
>  IIRC it only cancels builds that haven't yet started? I don't recall the conditions exactly, but it's something more complicated than simply killing builds when there's a push.

Yeah, I am not sure either. I just wondered why it built all the 30 builds of the previous push, although I updated like 3min after pushing originally, but oh well :man_shrugging:

ericniebler:
Yes, it only cancels builds that haven't started yet.

CaseyCarter:
I've found force-pushing to be more effective in general. Travis may not cancel the build, but individual jobs that start after the force push will fail to check out the old commit - since it no longer exists - and die quickly.

h-2:
I did a force-push, though :thinking: Well, whatever. Not too important I guess.

PR #1186, opened 2019-05-11
Small change to basic_iterator documentation.

tower120:
It is not clear, why we must specify `value_type`. Explanation added.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -441,7 +441,7 @@ The iterator returns a key/value pair, like the `zip` view.
     }
 ~~~~~~~
 
-`read()` returns references. So, we explicitly specify `value_type`.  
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -441,7 +441,7 @@ The iterator returns a key/value pair, like the `zip` view.
     }
 ~~~~~~~
 
-`read()` returns references. So, we explicitly specify `value_type`.  
+`read()` returns references. Default `value_type = decay_t<decltype(read())> = common_pair<Key&, Value&>`, which is not correct in our case. It should be `pair<Key, Value>`, so we explicitly specify `value_type`.
```
ericniebler: IIRC, it's incorrect to suggest there is a default for value type. There isn't. It must be specified always.

tower120: Here http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0186r0.html#cursors-cursor.intro says that `cursor` do have default `value_type`.

ericniebler: Ah, you're right. Never mind.

PR #1188, opened 2019-05-11
[FIX] iota-distance computation too strict

h-2:
The changes in c9d18a1e405c8d588859c430951eefb5fe5ae1f0 made iota hard fail for us, because it could not find an appropriate overload for some of our types (that do model the respective concepts). I think the overload below was just too strict by requiring SizedSentinel.

### DIFF
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -129,12 +129,12 @@ namespace ranges
                 i -= static_cast<Int>(-n);
         }
 
-        template<typename I, typename S>
-        auto iota_distance_(I const &i, S const &s) ->
+        template<typename I>
+        auto iota_distance_(I const &i, I const &s) ->
             CPP_ret(iota_difference_t<I>)(
-                requires SizedSentinel<S, I>)
+                requires Advanceable_<I> && (!Integral<I>))
         {
-            return s - i;
+            return static_cast<iota_difference_t<I>>(s - i);
         }
 
         template<typename Int>


h-2:
edit: pushed a little change to make sure the returned type is the `iota_difference_t`.

include/range/v3/view/iota.hpp
```diff
@@ -131,10 +131,9 @@ namespace ranges
 
         template<typename I, typename S>
         auto iota_distance_(I const &i, S const &s) ->
-            CPP_ret(iota_difference_t<I>)(
-                requires SizedSentinel<S, I>)
+            decltype((s - i, iota_difference_t<I>{}))
```
ericniebler:
```suggestion
            CPP_ret(iota_difference_t<I>)(
                requires Advanceable_<I> && (!Integral<I>))
```

include/range/v3/view/iota.hpp
```diff
@@ -131,10 +131,9 @@ namespace ranges
 
         template<typename I, typename S>
```
ericniebler:
```suggestion
        template<typename I>
```

include/range/v3/view/iota.hpp
```diff
@@ -131,10 +131,9 @@ namespace ranges
...
         auto iota_distance_(I const &i, S const &s) ->
```
ericniebler:
```suggestion
         auto iota_distance_(I const &i, I const &s) ->
```

h-2:
I have updated the PR and included your suggested changes. Some particular builds have failed, but I think they are just hiccups by travis. Can you restart those tests individually?

PR #1189, opened 2019-05-12
Fix to_container.hpp:declaration of 'size' hides global declaration

robert-andrzejuk:
…claration

MSVC 2019:
1>\range\v3\to_container.hpp(76,1): warning C4459:  declaration of 'size' hides global declaration
1>\range-v3\include\range\v3\to_container.hpp(76,1): warning C4459:                     auto const size = ranges::size(rng);
1>\range\v3\size.hpp(112,13): message :  see declaration of 'ranges::v3::CPOs::size'
1>\range\v3\size.hpp(112,13): message :             RANGES_INLINE_VARIABLE(_size_::fn, size)

### DIFF
--- a/include/range/v3/to_container.hpp
+++ b/include/range/v3/to_container.hpp
@@ -73,13 +73,13 @@ namespace ranges
                 Cont impl(Rng &&rng, std::true_type) const
                 {
                     Cont c;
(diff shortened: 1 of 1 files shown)


robert-andrzejuk:
@138

PR #1190, opened 2019-05-13
Sentinal-based Ranges should also take end Sentinels

ThePhD:
This pull request engages in a usability question: why does an unbounded range not take a sentinel?

I think that unbounded ranges should take 2 iterators like the rest of the ranges and containers and views, even if the second iterator is unreachable. It helps users not have to `if constexpr` internally in their algorithms to check for double-iterator constructibility to use these kinds of ranges.

I don't know if this should be extended to other range and view types, but I think it makes sense for, at least, `unbounded`.  Thoughts?

### DIFF
--- a/include/range/v3/view/unbounded.hpp
+++ b/include/range/v3/view/unbounded.hpp
@@ -34,6 +34,9 @@ namespace ranges
             constexpr explicit unbounded_view(I it)
               : it_(detail::move(it))
             {}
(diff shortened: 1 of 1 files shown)


CaseyCarter:
> why does an unbounded range not take a sentinel?

Could you please post this as an issue, with an example of problematic generic code that the proposed change fixes? The issue list is a useful archive of design discussion; we're likely to lose track of valuable history if we discuss design questions in pull requests.

PR #1197, opened 2019-05-16
Describe view::trim more conventionally

johelegp:


johelegp:
Would it have been okay to `[skip ci]`?

ericniebler:
What is `[skip ci]`?

CaseyCarter:
> What is `[skip ci]`?

I'll speculate it's a tag prepended to the beginning of the commit message suggesting that CI be skipped for that commit / PR. In which case, yes, it would have been fine to do so since we don't build the docs in CI (although we really should build, and possibly deploy, the docs).

PR #1202, opened 2019-05-21
Implement detail::is_trivial

### LINKED ISSUE #1196: Cannot compile tests with GCC 7.3 and C++17
The following `cmake` command configures range-v3 in such a way that building the tests is unsuccessful on Scientific Linux 7.6 with GCC 7.3.0 (not the native compiler) and C++17:
```
cmake3 -DCMAKE_BUILD_TYPE=Debug -DRANGES_CXX_STD=17 \
-DRANGES_VERBOSE_BUILD=ON -DRANGE_V3_DOCS=OFF \
-DRANGE_V3_EXAMPLES=OFF -DRANGE_V3_HEADER_CHECKS=OFF \
-DRANGE_V3_TESTS=ON \
-DCMAKE_INSTALL_PREFIX=~/tmp/range-v3 ../../sources-ref/range-v3
```
There are several failures, please see the attached
[range-v3-cmake.log](https://github.com/ericniebler/range-v3/files/3187312/range-v3-cmake.log) and [range-v3-build.log](https://github.com/ericniebler/range-v3/files/3187198/range-v3-build.log).
These failures occur with master, but can be reproduced with `0.5.0` with patch 0c20dbf05973368339aeae0c4e106560e2dcf76b.
The build is successful if I configure with the only change from above being `-DRANGES_CXX_STD=14`.
Is there something I'm missing here, or is this a real issue?

CaseyCarter:
Compilers' `is_trivial` intrinsics are not to be trusted (https://godbolt.org/z/8n_4OA).

Fixes #1196.

### DIFF
--- a/include/meta/meta_fwd.hpp
+++ b/include/meta/meta_fwd.hpp
@@ -158,8 +158,7 @@
 #endif
 
 #if (defined(__cpp_lib_type_trait_variable_templates) && \
-    __cpp_lib_type_trait_variable_templates > 0) || \
-    META_CXX_VER >= META_CXX_STD_17
+    __cpp_lib_type_trait_variable_templates > 0)
 #define META_CXX_TRAIT_VARIABLE_TEMPLATES 1
 #else
 #define META_CXX_TRAIT_VARIABLE_TEMPLATES 0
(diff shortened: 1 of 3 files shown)


chissg:
@CaseyCarter Is there a version of this that would apply as a patch directly to version `0.5.0`? This appears to rely on code added since `0.5.0`, and I'm uncertain as to the implications of including that code in the patch.

PR #1204, opened 2019-05-24
clang-format all the things

ericniebler:


ericniebler:
Maybe I should change `PointerAlignment` to be `"Middle"` just to fix the `requires` clauses, which are currently rendered as `requires Concept<T> &&Concept<U>` because clang-format gets confused.

CaseyCarter:
> Maybe I should change `PointerAlignment` to be `"Middle"` just to fix the `requires` clauses, which are currently rendered as `requires Concept<T> &&Concept<U>` because clang-format gets confused.

Use `and` instead of `&&`, maybe? I hate how middle alignment makes `Type & foo` and `Type * ptr` look like binary expressions. (Then again, I have no idea if clang-format treats `and` and `&&` differently.)

CaseyCarter:
Also, this would be a good time to switch to tabs. 👿

ericniebler:
> Also, this would be a good time to switch to tabs.

Never, you filthy heathen!

ericniebler:
> Use `and` instead of `&&`, maybe? 

I considered that, but just discontinued the use of the `not` keyword because of some non-conforming code that was having trouble with it. 

FWIW, I also dislike the middle-alignment of pointers and references, but I consider it an improvement just for the sake of the `requires` clauses.

CaseyCarter:
> I considered that, but just discontinued the use of the `not` keyword because of some non-conforming code that was having trouble with it.

I noticed, and wondered why, but not enough to ask. Now you've piqued my interest.

> FWIW, I also dislike the middle-alignment of pointers and references, but I consider it an improvement just for the sake of the `requires` clauses.

No need to explain; I'm well-versed in clang-format's hatred of *requires-clause*s.

CaseyCarter:
> > Also, this would be a good time to switch to tabs.
> 
> Never, you filthy heathen!

It could be worse: @cjdb could be in here suggesting 3-space indents.

cjdb:
> > > Also, this would be a good time to switch to tabs.
> >
> > Never, you filthy heathen!
>
> It could be worse: @cjdb could be in here suggesting 3-space indents.

This will surely solve the debate about 2-tabs-vs-4-tabs!

CaseyCarter:
> > > > Also, this would be a good time to switch to tabs.
> > > 
> > > 
> > > Never, you filthy heathen!
> > 
> > 
> > It could be worse: @cjdb could be in here suggesting 3-space indents.
> 
> This will surely solve the debate about 2-tabs-vs-4-tabs!

The nice thing about tabs is that everyone can use whatever indent they like. Tabs are inclusive. Thanks for coming to my TED talk.

include/range/v3/action/drop.hpp
```diff
@@ -34,20 +36,22 @@ namespace ranges
         private:
             friend action_access;
             template<typename Int>
-            static auto CPP_fun(bind)(drop_fn drop, Int n)(
+            static auto CPP_fun(bind)(drop_fn drop, Int n)( //
```
FloopCZ: I really have no idea what is the clang-format's reason for a trailing `//` before every `requires`. :slightly_smiling_face:

ericniebler: clang-format didn't add that; I did, so that clang-format wouldn't make a hash of the formatting. It has no idea what a _requires-clause_ is.

ericniebler:
> > I considered that, but just discontinued the use of the not keyword because of some non-conforming code that was having trouble with it.
> 
> I noticed, and wondered why, but not enough to ask. Now you've piqued my interest.

https://github.com/facebook/hhvm/blob/master/hphp/runtime/vm/jit/vasm-instr.h#L1059

:-(

ericniebler:
> The nice thing about tabs is that everyone can use whatever indent they like. Tabs are inclusive. Thanks for coming to my TED talk.

Presented without comment:
https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/

PR #1205, opened 2019-05-27
Adding a constexpr addressof to fix #1167.

### LINKED ISSUE #1167: Unconditionally constexpr ref_view uses non-constexpr std::addressof in C++14

BRevzin:
I'm going on the principle here that it's more important for `constexpr` to work than to handle overloaded unary `operator&()`. This fixes #1167 

If you all disagree, I can instead do something like:
```cpp
#ifdef __cpp_lib_address_constexpr
#define RANGES_ADDRESSOF_CONSTEXPR constexpr
#else
#define RANGES_ADDRESSOF_CONSTEXPR
#endif
```
and changes all the appropriate functions from `constexpr` to `RANGES_ADDRESSOF_CONSTEXPR`.

### DIFF
--- a/include/range/v3/utility/addressof.hpp
+++ b/include/range/v3/utility/addressof.hpp
@@ -14,9 +14,12 @@
 
 #include <memory>
 #include <type_traits>
+
 #include <concepts/concepts.hpp>
-#include <range/v3/detail/config.hpp>
+
 #include <range/v3/range_fwd.hpp>
+
+#include <range/v3/detail/config.hpp>
 
 namespace ranges
 {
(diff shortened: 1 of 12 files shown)


include/range/v3/utility/addressof.hpp
```diff
@@ -0,0 +1,40 @@
...
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
+#ifndef RANGES_V3_UTLITY_ADDRESSOF_HPP
+#define RANGES_V3_UTLITY_ADDRESSOF_HPP
+
+#ifdef __cpp_lib_addressof_constexpr
```
ericniebler: I think you need to `#include` some stdlib header before you can test stdlib feature macros.

include/range/v3/utility/addressof.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <memory>
+#endif
+
+namespace ranges
+{
+    /// \cond
+    namespace detail
+    {
+        template <typename T>
+        constexpr T* addressof(T& arg) noexcept
```
ericniebler: It should be possible to detect at compile-time whether `T` has an overloaded `operator&` and fall back to a non-`constexpr` implementation that does the `reinterpret_cast` dance; I would prefer that than using `&arg` and silently getting some weird result.

BRevzin: I can't seem to come up with a way that works on gcc C++11. I can't just `auto` deduce the return type of `my::test::addressof`, and if I add either a trailing or leading `decltype` expression that does `&t` or whatever, it doesn't find the `ignore operator&(ignore)` -- it just unconditionally gives me `T*`.

It looks like any unevaluated `&x` just short-circuits to doing addressof.

I can get this something like this working on gcc c++14, or clang.

BRevzin: Never mind, I think I got something working.

BRevzin: ... and lol, range-v3 doesn't support C+11 anymore! 🎉 Well, I feel like I accomplished something anyway.

include/range/v3/utility/addressof.hpp
```diff
@@ -0,0 +1,69 @@
...
+    {
+#ifdef __cpp_lib_addressof_constexpr
+        using std::addressof;
+#else
+        namespace test {
+            struct ignore {
+                template <typename T> ignore(T&&) { }
+            };
+
+            ignore operator&(ignore);
```
ericniebler: I suspect if you defined this function, the travis errors will go away.

```suggestion
            inline ignore operator&(ignore) { return {}; }
```

Also, use `detail::ignore_t` which range-v3 already provides, and give the `test` namespace a more descriptive name.

PR #1214, opened 2019-06-04
Make conan package always compatible

### LINKED ISSUE #1212: Conan package could always be compatible, as it's header-only
Being a header-only library, it should be possible to make the range-v3 conan package always compatible with any compiler/architecture/build type/etc. There's no need to re-package range-v3.

Care has to be taken with `write_basic_package_version_file`, though, as it requires matching architecture, which a header-only library does not need. Either we could upgrade to CMake 3.14 and use the `ARCH_INDEPENDENT` option, we can use a custom versioning system, or we can use a hack and set `CMAKE_SIZEOF_VOID_P` to nothing before calling `write_basic_package_version_file`.

Quincunx271:
Fixes #1212 

Changes:

- conanfile.py: Fix typo: `package_info` -> `package_id`. `package_info(self)` is for settings things like compiler flags and options, whereas `package_id(self)` is for determining package compatibility.
- CMakeLists.txt: `write_basic_package_version_file` has a hidden architecture check. Disable it via the workaround of unsetting `CMAKE_SIZEOF_VOID_P`, at least until an upgrade to CMake 3.14 is reasonable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,11 +145,18 @@ if(NOT RANGE_V3_NEW_VERSION_HPP STREQUAL RANGE_V3_OLD_VERSION_HPP)
 endif()
 
 include(CMakePackageConfigHelpers)
+
+# write_basic_package_version_file(...) gained ARCH_INDEPENDENT in CMake 3.14.
+# For CMake 3.6, this workaround makes the version file ARCH_INDEPENDENT
+# by making CMAKE_SIZEOF_VOID_P empty.
+set(OLD_CMAKE_SIZEOF_VOID_P ${CMAKE_SIZEOF_VOID_P})
+set(CMAKE_SIZEOF_VOID_P "")
 write_basic_package_version_file(
   ${CMAKE_CURRENT_BINARY_DIR}/range-v3-config-version.cmake
   VERSION ${RANGE_V3_VERSION}
   COMPATIBILITY ExactVersion
 )
+set(CMAKE_SIZEOF_VOID_P ${OLD_CMAKE_SIZEOF_VOID_P})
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -145,11 +145,18 @@ if(NOT RANGE_V3_NEW_VERSION_HPP STREQUAL RANGE_V3_OLD_VERSION_HPP)
 endif()
 
 include(CMakePackageConfigHelpers)
+
+# write_basic_package_version_file(...) gained ARCH_INDEPENDENT in CMake 3.14.
+# For CMake 3.6, this workaround makes the version file ARCH_INDEPENDENT
+# by unsetting CMAKE_SIZEOF_VOID_P.
+set(OLD_CMAKE_SIZEOF_VOID_P ${CMAKE_SIZEOF_VOID_P})
+unset(CMAKE_SIZEOF_VOID_P)
```
Quincunx271: Should use `set(CMAKE_SIZEOF_VOID_P)` rather than `unset(...)`: https://cmake.org/cmake/help/latest/command/unset.html#unset-normal-variable-or-cache-entry states:

> [U]nsetting a normal variable can expose a cache variable that was previously hidden. To force a variable reference of the form `${VAR}` to return an empty string, use `set(<variable> "")`, which clears the normal variable but leaves it defined.

Quincunx271: Should use `set(CMAKE_SIZEOF_VOID_P)` rather than `unset(...)`: https://cmake.org/cmake/help/latest/command/unset.html#unset-normal-variable-or-cache-entry states:

> [U]nsetting a normal variable can expose a cache variable that was previously hidden. To force a variable reference of the form `${VAR}` to return an empty string, use `set(<variable> "")`, which clears the normal variable but leaves it defined.

ericniebler:
Thanks.

Issue #1162, opened 2019-04-26
Add view::take_last


johelegp:
# `view::take_last`

The name `view::take_last` denotes
a range adaptor object (http://wg21.link/range.adaptors#def:range_adaptor_object).
For some subexpressions `E` and `N`,
the expression `view::take_last(E, N)`
is expression-equivalent to `view::drop_exactly(E, ranges::size(E) - N)`,
except that `E` is only evaluated once.

johelegp:
Thanks @BRezzin.

CaseyCarter:
To be perfectly clear: This issue is closed by #1218.

PR #1218, opened 2019-06-19
Adding take_last.

BRevzin:
Implements #1162, `take_last` simply calls `drop_exactly` and is only supported for `SizedRange`s. 

If you pass in a size larger than the input size, I just round down.


include/range/v3/view/take_last.hpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    namespace view
+    {
+        struct take_last_fn
+        {
+        private:
+            friend view_access;
+            template<typename Int>
+            static auto CPP_fun(bind)(take_last_fn take_last, Int n)( //
+                requires Integral<int>)
```
timsong-cpp: Int, not int.

doc/index.md
```diff
@@ -763,6 +763,8 @@ provides, and a blurb about how each is intended to be used.
   <DD>Given a source range and an integral count, return a range consisting of the first *count* elements from the source range, or the complete range if it has fewer elements. (The result of `view::take` is not a `SizedRange`.)</DD>
 <DT>\link ranges::view::take_exactly_fn `view::take_exactly`\endlink</DT>
   <DD>Given a source range and an integral count, return a range consisting of the first *count* elements from the source range. The source range must have at least that many elements. (The result of `view::take_exactly` is a `SizedRange`.)</DD>
+<DT>\link ranges::view::drop_exactly_fn `view::take_last`\endlink</DT>
```
johelegp: There's a `drop` here that should be changed to `take`.

BRevzin: This one's right - there's not actually a `take_last_fn` in this implementation, `take_last` just calls `drop_exactly`. Unless it should like somewhere else?

johelegp: The `_fn` are the range adaptors, which must exist, unlike the `_view`.

BRevzin: Oh, I misunderstood everything.

johelegp: I must correct myself. The range adaptors are the `_fn` wrapped in `view::view<>`.

BRevzin:
retest this please

CaseyCarter:
LGTM. @ericniebler, do you want to perform design sign-off on this feature addition? Most notably, do we want to take this as-is or require support for forward non-sized ranges?

include/range/v3/view/take_last.hpp
```diff
@@ -0,0 +1,51 @@
...
+            {
+                return make_pipeable(std::bind(take_last, std::placeholders::_1, n));
+            }
+
+        public:
+            template<typename Rng>
+            auto CPP_fun(operator())(Rng && rng, range_difference_t<Rng> n)(
+                const requires ViewableRange<Rng> && SizedRange<Rng>)
+            {
+                auto sz = ranges::distance(rng);
```
ericniebler: Why is this `distance` and not `size`?

BRevzin: signed vs unsigned.

Issue #1171, opened 2019-04-30
Add remove_when_view


johelegp:
`view::remove_when(E, F)` does what `view::split_when(E, F) | view::join` does.

PR #1219, opened 2019-06-19
Adding view::remove_when.

BRevzin:
Implements #1171, based on what my understanding is of what that's supposed to do.


test/view/remove_when.cpp
```diff
@@ -0,0 +1,40 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2014-present
```
johelegp: Same here.

include/range/v3/view/remove_when.hpp
```diff
@@ -0,0 +1,181 @@
...
+    {
+        struct remove_when_fn
+        {
+        private:
+            friend view_access;
+            template<typename Fun>
+            static auto bind(remove_when_fn remove_when, Fun && fun)
+            {
+                return make_pipeable(std::bind(
+                    remove_when, std::placeholders::_1, static_cast<Fun &&>(fun)));
```
ericniebler: Please rebase on master and replace `std::bind` with `bind_back`, and make this function `constexpr`. (You should also `#include <range/v3/functional/bind_back.hpp>`.)

ericniebler:
When you rebase on master, you also need to accommodate the Great Concept Rename (i.e., the great_concept_rename), and the fact that the `view` namespace is now the `views` namespace.

PR #1199, opened 2019-05-18
constexpr friendly view::view

tower120:
This PR partially fix/implement this https://github.com/ericniebler/range-v3/issues/1169.
`view::view` are now constexpr-friendly.

---

Now what we lack is constexpr "bind". Our options are:
  * use lambdas instead of `std::bind`. Lambdas are constexpr in C++17.
  * use third-party constexp bind. Like https://github.com/bolero-MURAKAMI/Sprout/blob/master/sprout/functional/bind/bind.hpp from https://github.com/bolero-MURAKAMI/Sprout/ . What about copyright clashes?
  * write our own constexpr friendly bind. (I don't think I can/want do this)

Issue #1169, opened 2019-04-30
Unmet standard requirements for range adaptors


johelegp:
http://wg21.link/range.adaptor.object and the description of the various range adaptor objects talk about equivalent expressions which Ranges v3 cannot meet.

`view::X(r, args)` can't be [expression-equivalent](http://wg21.link/defns.expression-equivalent) to `X_view{r, args}` when the latter is a constant subexpression as the former can't never be because `view::X_fn::operator()`s aren't marked `constexpr`.

Even if `view::X(r, args)` and `X_view{r, args}` were both constant subexpressions, `view::X(args)` can never be because they use `std::bind` to make a pipeable. `view::view` also seems to be missing some `constexpr` qualifiers.

johelegp:
Isn't this solved?

Issue #1136, opened 2019-04-15
Allow actions to strore references/pointers


tower120:
Actions which accept values as parameters, currently highly inefficient by design. All passed values to action, required to be stored as value. That's **greatly** reduces their efficiency. **And** forbids use of move-only objects.
```cpp
std::list<HeavyObject> list;       // What if HeavyObject move-only?

void foo(const HeavyObject& obj){
     list |= action::remove(obj, pred);   
     // I don't want obj to be copied !!
     // Maybe there is even no obj in list !   
}
```
----
I propose following:
  * store passed rvalues/rvalue references as values.
  * store passed references / const references as pointers.

Distinguish between const T& and T&& with this:

http://coliru.stacked-crooked.com/a/a54daa2b7933e5c4
```cpp
    template<class T>
    void g(const T&){
        std::cout << "T&" << std::endl;
    }
    
    // or std::is_rvalue_reference_v<T> || !std::is_reference_v<T>
    template<class T, typename = std::enable_if_t< !std::is_lvalue_reference_v<T> >>  
    void g(T&&){
        std::cout << "T&&" << std::endl;
    }
```

We'll make concept trait so it don't look so ugly.

----

Dangling.

Storing pointers/references introduce situations were action may outlive stored value pointer. That's true. But I think, considering the fact, that most actions applied here and now, benefits greatly overweight potential "safety". 

Besides that's not something unique to C++. Range as pair of iterators may become invalid by the same reason.

ericniebler:
I am opposed to changing something like lifetime based on something as ephemeral as an expression's value category. Consider:

```
template <Semiregular Pred = ranges::less>
auto sort_and_unique(Pred pred = {}) {
  return action::sort(pred) | action::unique; // Oops!
}
```

Possibly there should be a protocol like how `std::bind` uses `std::ref` through which users can communicate what things should be stored by reference. But I feel that storing by value is the correct default, and that users should have to explicitly opt in to storing by reference.

tower120:
OFFTOPIC ON

> ...  something like lifetime based on ...

What is, "lifetime based", in terms of C++? Ownership based?

>  Consider:
> ```
> template <Semiregular Pred = ranges::less>
> auto sort_and_unique(Pred pred = {}) {
>   return action::sort(pred) | action::unique; // Oops!
> }
> ```

Purely matter of habit:
```cpp
template <Semiregular Pred = ranges::less>
auto sort_and_unique(Pred pred = {}) {
[...]

PR #1233, opened 2019-07-01
Making views binds constexpr friendly

tower120:
This PR is continuation of https://github.com/ericniebler/range-v3/issues/1169 bug fix.

Previously, we agreed that all view bind's must use "function object" instead of `std::bind` (due to being non-constexpr), and instead of lambdas (they're not constexpr in pre C++17 era).
 https://github.com/ericniebler/range-v3/pull/1199#issuecomment-497196332

I tried that. That was HUGE amount of code! `bind` was MUCH shorter.
Instead of using "function object" I made `bind_back`. Please consider this approach...

If using `bind_back` is OK, I'll proceed updating other views.

UPDATE.
`bind_front` renamed to `bind_back`


johelegp:
`bind_front` is actually `bind_tail`.

tower120:
@johelegp Thanks you're right! That would be `bind_tail<n=-1>(fn, args...)`, where if n>0, constrain call_args count.
Nevertheless, I need approve from @ericniebler , that this is acceptable.

tower120:
This PR should probably also implement this https://github.com/ericniebler/range-v3/issues/1136#issuecomment-507680419 . `bind_back` should not unwrap `std::reference_wrapper`s.

include/range/v3/functional/bind_back.hpp
```diff
@@ -0,0 +1,100 @@
...
+        constexpr bind_back_t(Fn && fn, Args &&... args)
+          : fn(std::move(fn))
+          , args_tuple(std::move(args)...)
+        {}
+
+        template<typename... CallArgs>
+        constexpr decltype(auto) operator()(CallArgs &&... call_args) noexcept(
+            noexcept(run(std::declval<Self &>(), std::forward<CallArgs>(call_args)...)))
+        {
+            return run(*this, std::forward<CallArgs>(call_args)...);
```
ericniebler: I always forget if the presence of a member of the same name disables ADL, or if this is hijack-able. So I tend to be conservative and always qualify with the class name, which is clearer to readers who (like me) are confused about this. Could you please qualify all the calls to "`run`"? Thanks.

ericniebler:
I like the direction! I think I would prefer `bind_back` to be an implementation detail, though. It doesn't really behave like `std::bind` with regards to nested binds, and that could be potentially confusing.

tower120:
> I like the direction! I think I would prefer `bind_back` to be an implementation detail, though. It doesn't really behave like `std::bind` with regards to nested binds, and that could be potentially confusing.

FYI, I mimicked `std::bind_front` interface, not `std::bind`. `std::bind_front` does not unwind nested binds too! https://en.cppreference.com/w/cpp/utility/functional/bind_front

But that's not a problem, I put that into `details` namespace.

----

Also, since I putting that into `details`, I want to make additional change. I'll throw out `invoke`, and will directly call function:
 
```cpp
        template<typename Self, std::size_t... I, typename... CallArgs>
        constexpr auto static CPP_auto_fun(apply)
            (Self & self, std::index_sequence<I...>, CallArgs &&... call_args)
        (
            return self.fn(    /* was  invoke(self.fn */
                          std::forward<CallArgs>(call_args)...,
                          std::get<I>(self.args_tuple)...)
        )
```

This will prevent unwrapping `std::ref`s. And `std::ref`s will go into `view::view::operator()` as is. This will allow to implement https://github.com/ericniebler/range-v3/issues/1234 latter.  Plus, that just will compile faster... It should not break anything now, since `range-v3` does not claim that it support `std::ref`s at all.

include/range/v3/functional/bind_back.hpp
tower120: np

tower120:
Views `sample`, `join`,  `split_when` are omitted (left as is). They capture optional argument by automatically wrap it into `std::ref`. I want to make that explicit. Please consider https://github.com/ericniebler/range-v3/issues/1234.

ericniebler:
Does the `bind_back` implementation take into account [p1651](http://wg21.link/p1651)?

include/range/v3/detail/bind_back.hpp
```diff
@@ -0,0 +1,113 @@
...
+            // clang-format on
+
+        public:
+            constexpr bind_back_t(Fn && fn, Args &&... args)
+              : fn(std::move(fn))
+              , args_tuple(std::move(args)...)
+            {}
+
+            template<typename... CallArgs>
+            constexpr decltype(auto) operator()(CallArgs &&... call_args) noexcept(
```
ericniebler: The use of `decltype(auto)` here is problematic since invalid calls to `operator()` will not SFINAE away; instead, they will cause hard errors.

include/range/v3/detail/bind_back.hpp
```diff
@@ -0,0 +1,113 @@
...
+        class bind_back_t
+        {
+            using Self = bind_back_t<args_count, Fn, Args...>;
+
+            // clang-format off
+            template<typename Self, std::size_t... I, typename... CallArgs>
+            constexpr auto static CPP_auto_fun(apply)
+                (Self & self, std::index_sequence<I...>, CallArgs &&... call_args)
+            (
+                return self.fn(std::forward<CallArgs>(call_args)...,
```
ericniebler: I'm unconvinced that this shouldn't be `invoke` here. I think for parity with `std::bind_front` it must.

include/range/v3/detail/bind_back.hpp
```diff
@@ -0,0 +1,113 @@
...
+                     std::remove_reference_t<T> &&>> //
+        constexpr U bind_back_forward(std::remove_reference_t<T> & t) noexcept
+        {
+            return static_cast<U>(t);
+        }
+    } // namespace detail
+
+    // clang-format off
+    // args_count >= 0 constraint arguments count on call
+    template<int args_count = -1, typename Fn, typename... Args>
```
ericniebler: I am not partial to the `arg_count` template parameter. My weak preference is to drop it as needless complication.

tower120: I disagree. It provides additional layer of compile time safety. Without it, you mistakenly may pass additional argument(s), which may result in call of "wrong" `operator()()` overload.

For example:
`bind_back<1>(drop, n)` will result in `fn(auto&&)`, while `bind_back(drop, n)` => `fn(auto&&...)`

include/range/v3/detail/bind_back.hpp
tower120: Why (when) you want SFINAE machinery to work here?

tower120:
> Does the `bind_back` implementation take into account [p1651](http://wg21.link/p1651)?

I was unaware about that proposal previously.
Now looking at it, it seems that current implementation (direct function call, instead of `std::invoke`) follow "propagation semantic" = "do NOT unwrap". IOW, yes, it seems that it follow/implements [p1651](http://wg21.link/p1651).

tower120:
> I think `bind_back_forward` is no longer necessary after p1651, is that right?

No... `bind_back_forward` needed purely as a way to remove additional move constructor call when you already pass argument as rvalue ref. Without it, that would be:

```cpp
            // this idiom used almost eveywhere in range-v3
            constexpr bind_back_t(Fn fn, Args... args)   // move constructors called here!!!
              : fn(std::move(fn))
              , args_tuple(std::move(args)...)
            {}

    constexpr auto CPP_auto_fun(bind_back)(Fn && fn, Args &&... args)
    (
        return detail::bind_back_t<args_count, std::decay_t<Fn>, std::decay_t<Args>...>(
            std::move(fn), std::move(args)...)
    )

  bind_back(fn, std::move(my_obj));
```

There is probably another solution(s) to that... But that one looked like a shortest one to me.

include/range/v3/detail/bind_back.hpp
tower120: With `invoke` it will follow CURRENT `std::bind_front` contract.
With direct call - p1651.

ericniebler:
I've made some changes and resumbitted this PR as #1249. Closing this one.

PR #1237, opened 2019-07-09
Add include/std/* to the list of headers available through bazel rules

Eipifi:
The commit abf16a6c6a1919cacc25645588922b1cebe9c879 added new headers to include/std, but these were not exposed in Bazel rules and caused the builds to break.


ericniebler:
Thank you!

PR #1239, opened 2019-07-21
includes new C++20 range adaptors in namespace cpp20

cjdb:
The following are added to namespace cpp20:

* `take_while_view`
* `drop_view`
* `drop_while_view`
* `keys`
* `values`

The following are not added:

* `elements_view` (doesn't appear to exist)
* `istream_view` (will file NB comment to have `basic_istream_view` renamed to `istream_view`)

### DIFF
--- a/include/range/v3/view/map.hpp
+++ b/include/range/v3/view/map.hpp
@@ -124,6 +124,16 @@ namespace ranges
         /// \ingroup group-views
         RANGES_INLINE_VARIABLE(view<values_fn>, values)
     } // namespace view
+
+    namespace cpp20
+    {
+        namespace view
+        {
+            using ranges::view::keys;
+            using ranges::view::values;
+        }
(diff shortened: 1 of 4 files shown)


include/range/v3/view/map.hpp
```diff
@@ -124,6 +124,16 @@ namespace ranges
...
     } // namespace view
+
+    namespace cpp20
+    {
+        namespace view
+        {
+            using ranges::view::keys;
+            using ranges::view::values;
+        }
+        // TODO(@cjdb): provide implementation for elements_view
```
ericniebler: Shall I wait for this?

cjdb: I think it'd be better for me to make that a separate PR. I don't have the time to make that right now ☹️. Would you like me to remove the TODO?

PR #1244, opened 2019-07-30
Workaround deprecation warnings

CaseyCarter:
...in tests triggered by #1238. Conversions to nested containers use `view_interface`'s now-deprecated conversion operator templates, triggering deprecation warnings. CI didn't catch this problem because the test cases all use standard containers, which are defined in system headers, for which clang and GCC don't emit warnings.

### DIFF
--- a/test/range/conversion.cpp
+++ b/test/range/conversion.cpp
@@ -25,6 +25,8 @@
 #include "../simple_test.hpp"
 #include "../test_utils.hpp"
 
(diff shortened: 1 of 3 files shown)


ericniebler:
I don't want to hide the deprecation warnings. I want to avoid all use of the deprecated conversion operators.

CaseyCarter:
> I don't want to hide the deprecation warnings. I want to avoid all use of the deprecated conversion operators.

Agreed. This PR is a stopgap to keep the tests passing, #1245 is the "oops we need to really fix this" bug.

PR #1248, opened 2019-08-07
MSVC update:

CaseyCarter:
* React to future changes to MSVC's implementation of `iterator_traits` in `std/detail/associated_types.hpp`

* Use correct bug number for the `concepts::swap` workaround

Drive-by: Remove extraneous parens in the definition of concept `detail::ConvertibletoContainerContainerImpl`

### DIFF
--- a/include/std/detail/associated_types.hpp
+++ b/include/std/detail/associated_types.hpp
@@ -243,6 +243,11 @@ namespace ranges
 
         // For testing whether a particular instantiation of std::iterator_traits
         // is user-specified or not.
+#if defined(_MSVC_STL_UPDATE) && defined(__cpp_lib_concepts) && _MSVC_STL_UPDATE >= 201908L
+        template<typename I>
+        inline constexpr bool is_std_iterator_traits_specialized_v =
+            !std::_Is_from_primary<std::iterator_traits<I>>;
+#else
 #if defined(__GLIBCXX__)
         template<typename I>
         char (&is_std_iterator_traits_specialized_impl_(std::__iterator_traits<I> *))[2];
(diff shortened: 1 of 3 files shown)


include/std/detail/associated_types.hpp
```diff
@@ -243,6 +243,11 @@ namespace ranges
 
         // For testing whether a particular instantiation of std::iterator_traits
         // is user-specified or not.
+#if defined(_MSVC_STL_UPDATE) && defined(__cpp_lib_concepts) && _MSVC_STL_UPDATE >= 201908L
+        template<typename I>
+        inline constexpr bool is_std_iterator_traits_specialized_v =
+            !std::_Is_from_primary<std::iterator_traits<I>>;
```
ericniebler: `_Is_from_primary`? What voodoo is this?!

CaseyCarter: Nothing special: just a concept whose definition is `same_as<T, typename T::_From_primary>`. The `iterator_traits` "primary template" has a `using _From_primary = iterator_traits;` member. We've been doing something similar with `allocator` for optimization purposes for a while. (The `same_as` bit is necessary to catch program-defined specializations that inherit from a library-defined specialization.)

ericniebler: Ok yes. I believe I recommend that approach in my Deep Integration paper. 👍

CaseyCarter: Huh - you did recommend a variation thereof (Mea Culpa: I never read anything but the wording.):

> For instance, the primary template `std::iterator_traits<I>` might define a hidden nested typedef `__unspecialized` that is an alias for `I`. Any instantiation `std::iterator_traits<I>` that either lacks that typedef or has one that is not an alias for `I` is necessarily not the primary template.

Aliasing the specialization is a bit more general than aliasing a parameter. Admittedly both use cases in MSVC's STL are single-type-parameter class templates.

PR #1249, opened 2019-08-09
Replacing views binds

ericniebler:


tower120:
Just to note. Your `bind_back` implementation uses `tuple_apply`. `tuple_apply` call `invoke` under the hood. `invoke` unwrap `std::ref`'s. You follow CURRENT `std::bind_front` contract. Without p1651. If I understood it corectly...

I didn't get, which one you wanted... But better safe then sorry.

ericniebler:
`invoke` doesn't treat `reference_wrapper`s specially.

tower120:
> `invoke` doesn't treat `reference_wrapper`s specially.

Well to say the truth, there is one special case https://en.cppreference.com/w/cpp/utility/functional/invoke . But veeery special one. That misguided me. Sorry for false alarm.

PR #1259, opened 2019-08-15
lazy generate is too lazy

### LINKED ISSUE #1258: Unexpected behavior using ranges::view::drop after ranges::view::generate
It appears that using `ranges::view::drop` has no effect when applied to a range produced by `ranges::view::generate`.

```cpp
  auto fib = [p = std::pair{0, 1}]() mutable -> int {
    auto [a, b] = p;
    p = {b, a + b};
    return a;
  };

  auto rng = ranges::view::generate(fib)
    | ranges::view::drop(3)
    | ranges::view::take(5);

  std::cout << rng << std::endl;
  // Expected: [2,3,5,8,13]
  // Got: [0,1,1,2,3]
```

CaseyCarter:
We must call the generate function for elements that are incremented past and never read.

Fixes #1258.


include/range/v3/view/generate_n.hpp
```diff
@@ -67,7 +67,10 @@ namespace ranges
             {
                 RANGES_EXPECT(0 != rng_->n_);
                 --rng_->n_;
-                rng_->val_.reset();
+                if (rng_->val_)
+                    rng_->val_.reset();
+                else
+                    rng_->gen_();
```
CaseyCarter: Hmmm...should we update `n_` after calling `gen_` for exception safety?

CaseyCarter: I don't think it's hugely important, but it does seem more consistent with the behavior of `*i; ++i;` and `*i++` - so why not.

PR #1264, opened 2019-08-19
simplify pipeable, unroll bind_back_fn

ericniebler:


include/meta/meta_fwd.hpp
```diff
@@ -174,6 +174,24 @@
 #define META_IS_SAME(...) std::is_same<__VA_ARGS__>::value
 #endif
 
+#if defined(__GNUC__)
+#define META_IS_BASE_OF(...) __is_base_of(__VA_ARGS__)
```
CaseyCarter: Clang and MSVC also support this intrinsic with the same name.

include/meta/meta_fwd.hpp
```diff
@@ -174,6 +174,24 @@
...
+#elif META_CXX_TRAIT_VARIABLE_TEMPLATES
+#define META_IS_BASE_OF(...) std::is_base_of_v<__VA_ARGS__>
+#else
+#define META_IS_BASE_OF(...) std::is_base_of<__VA_ARGS__>::value
+#endif
+
+#if defined(__clang__)
+#define META_IS_CONSTRUCTIBLE(...) __is_constructible(__VA_ARGS__)
+#elif defined(__GNUC__) && __GNUC__ >= 8
+#define META_IS_CONSTRUCTIBLE(...) __is_constructible(__VA_ARGS__)
```
CaseyCarter: This is identical to 186 - you could disjoin the conditions and eliminate the duplication. Also, MSVC supports this intrinsic.

include/range/v3/range/conversion.hpp
```diff
@@ -206,8 +206,8 @@ namespace ranges
             template(typename Rng, typename Cont)
             concept convertible_to_container_container_impl_,
                 range<Cont> && (!view_<Cont>) && move_constructible<Cont> &&
-                is_true<ranges::defer::range<range_value_t<Cont>> &&
-                    !ranges::defer::view_<range_value_t<Cont>>> &&
+                bool(ranges::defer::range<range_value_t<Cont>> &&
+                    !ranges::defer::view_<range_value_t<Cont>>) &&
```
CaseyCarter: `bool(...)` is not a primary expression - this needs parentheses.

include/meta/meta_fwd.hpp
ericniebler: Clang defines `__GNUC__`. I didn't know about MSVC, thanks.

include/range/v3/range/conversion.hpp
ericniebler: Bah.

include/meta/meta_fwd.hpp
CaseyCarter: > Clang defines `__GNUC__`

Only in "pretending to be GCC" mode. In "pretending to be MSVC" mode it defines `_MSC_VER` instead. But yes, I think `#if defined(__GNUC__) || defined(__MSC_VER__)` covers all three.

ericniebler: I didn't know that!

CaseyCarter:
FYI, I fixed the merge conflict with #1263.

PR #1267, opened 2019-08-20
Workaround fixed MSVC lambda bugs

CaseyCarter:
... which are only fixed in c++latest or with /experimental:newLambdaProcessor

Drive-by:
* Optimize the metaprogramming in `<std/detail/associated_types.hpp>` for MSVC by using the `std` type traits directly (since they're mostly implemented in the compiler)
* Don't wrap lambdas in `bind_back` when we can use them directly.
* In `<action/sort.hpp>`, don't repeat typenames in predicate and projection default arguments.

### DIFF
--- a/include/range/v3/action/push_back.hpp
+++ b/include/range/v3/action/push_back.hpp
@@ -78,17 +78,33 @@ namespace ranges
             {
                 return bind_back(push_back, static_cast<T &&>(val));
             }
+#ifdef RANGES_WORKAROUND_MSVC_OLD_LAMBDA
             template<typename T, std::size_t N>
-            static auto bind(push_back_fn push_back, T (&val)[N])
+            struct lamduh
             {
-                return bind_back(
-                    [push_back](auto && rng, T(*val)[N])
-                        -> invoke_result_t<push_back_fn, decltype(rng), T(&)[N]> {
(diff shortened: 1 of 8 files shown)


include/range/v3/action/push_back.hpp
```diff
@@ -78,17 +78,33 @@ namespace ranges
...
+            {
+                return {val};
+            }
+#else // ^^^ workaround / no workaround vvv
+            template<typename T, std::size_t N>
+            static auto bind(push_back_fn, T (&val)[N])
+            {
+                return [&val](auto && rng)
+                    -> invoke_result_t<push_back_fn, decltype(rng), T(&)[N]> {
+                    return push_back_fn{}(static_cast<decltype(rng)>(rng), val);
```
CaseyCarter: @ericniebler You may want to look at this "no workaround" bit. I'm flattening "generic lambda that captures some stuff wrapped in `bind_back` that captures some other stuff" into "generic lambda that captures all the stuff" for throughput. I don't *think* I've lost any semantic value here - did you have an ulterior motive for the multiple layers of function wrappers that I'm missing?

ericniebler: I think this is fine. `bind_back::operator()` has overloads for `&&`, `&`, and `const &`; none of which matters when the only captures are reference captures.

PR #1273, opened 2019-08-22
Rename meta::detail::bool_ to avoid inadvertent use

CaseyCarter:
...such as on meta.hpp line 3103.


CaseyCarter:
> ...such as on meta.hpp line 3103.

That's right - I'm naming and shaming!

ericniebler:
> That's right - I'm naming and shaming!

Shame

PR #1274, opened 2019-08-22
view and action examples; workaround for gcc-7/8 ICE, fixes #1105

### LINKED ISSUE #1105: Streaming a istream_range/istream_view with ranges-v3 v1.0-beta causes an ICE
Hi, 

I am working with the v1.0-beta branch of ranges-v3 and I ran into an ICE when trying to print an istream_range/istream_view.

The following code fails for 
GCC-7.3, GCC-7.4 (Ubuntu 18.04.1) and 
GCC8 (g++-mp-8 (MacPorts gcc8 8.3.0_0) 8.3.0) 
with the v1.0-beta branch
```cpp
#include <string>
#include <sstream>
#include <iostream>
#include <range/v3/istream_range.hpp>

int main ()
{
    std::string vec{"foo\nbar"};
    std::istringstream ss{vec};
    auto v = ranges::istream_range<char>(ss);

    std::cout << v;
}
``` 
Using the `istream_view` in `<range/v3/view/istream.hpp>` also causes an ICE.
The code compiles fine on versions 0.4.0 and earlier.

ericniebler:


CaseyCarter:
```
..\example\filter_transform.cpp
..\example\filter_transform.cpp(29,57): error: no member named 'to_string' in namespace 'std'
               views::transform([](int i) { return std::to_string(i); });
                                                   ~~~~~^
In file included from ..\example\filter_transform.cpp:20:
In file included from ..\include\range/v3/view/filter.hpp:21:
```
*cough* `#include<string>` *cough*

PR #1275, opened 2019-08-22
Try to reduce issue reports from MSVC users

CaseyCarter:
* export `/permissive- /experimental:preprocessor` as required flags in the cmake config

* Yell in `config.hpp` if we see old MSVC, C++14 mode, or the old preprocessor.


CaseyCarter:
CI succeeded on my Travis for this commit (https://travis-ci.org/CaseyCarter/range-v3/builds/575578118) so I'm going to merge without waiting for CI to complete here to conserve resources.

PR #1277, opened 2019-08-23
Fix inductive proof of distance between cartesian_product iterators

### LINKED ISSUE #1269: cartesian_product + transform + to gives incorrect size
I have the following test code:

```c++
#include <vector>
#include <cassert>
#include <iostream>
#include <range/v3/all.hpp>

template<typename ...T>
auto cp(const std::vector<T>& ...a)
{
    return ranges::view::cartesian_product(a...);
}

std::vector<double> distribute(double len, int n)
{
    return ranges::view::linear_distribute(0.0, len, n) | ranges::to<std::vector>();
}

int main()
{
    unsigned size_x = 5, size_y = 4, size_z = 6;

    auto x = distribute(1.0, size_x);
    assert(x.size() == size_x);
    assert(x.back() == 1.0);

    auto y = distribute(2.0, size_y);
    assert(y.size() == size_y);
    assert(y.back() == 2.0);

    auto z = distribute(3.0, size_z);
    assert(z.size() == size_z);
    assert(z.back() == 3.0);

    auto xyz = cp(x, y, z);
    auto sz = x.size() * y.size() * z.size();

    assert(xyz.size() == sz);

    auto t = ranges::view::transform([&](const auto& location) {
					 return std::get<0>(location) * std::get<1>(location) * std::get<2>(location);
				     });
    auto loc = xyz | t | ranges::to<std::vector>();

    std::cout << "loc.size(): " << loc.size() << "\n";
    assert(loc.size() == sz);
    assert(loc.back() == 6.0);
}
```

The problem is that the `loc` variable never reports a size greater than 100 which triggers the assert:

```bash 
loc.size(): 100
a.out: cartesian_product_test.cpp:44: int main(): Assertion `loc.size() == sz' failed.
Aborted (core dumped)
```
Changing `size_z=5` gives the correct results.  I'm compiling on linux using gcc 8.3.1 and the range v3 master branch at commit  25962cf64.

Am I using ranges incorrectly or is this a bug?

CaseyCarter:
Fixes #1269.

### DIFF
--- a/include/range/v3/view/cartesian_product.hpp
+++ b/include/range/v3/view/cartesian_product.hpp
@@ -118,12 +118,11 @@ namespace ranges
     // clang-format on
 
     template<typename... Views>
-    class cartesian_product_view
-      : public view_facade<cartesian_product_view<Views...>,
-                           detail::cartesian_product_cardinality<Views...>::value>
+    struct cartesian_product_view
+      : view_facade<cartesian_product_view<Views...>,
+                    detail::cartesian_product_cardinality<Views...>::value>
     {
-        /// \cond
-    public: // BUGBUG
+    private:
         friend range_access;
         CPP_assert(and_v<(forward_range<Views> && view_<Views>)...>);
         CPP_assert(sizeof...(Views) != 0);
@@ -134,10 +133,11 @@ namespace ranges
         std::tuple<Views...> views_;
 
         template<bool IsConst_>
-        class cursor
+        struct cursor
         {
+        private:
             using IsConst = meta::bool_<IsConst_>;
-            friend class cursor<!IsConst_>;
+            friend cursor<true>;
             template<typename T>
             using constify_if = meta::const_if_c<IsConst_, T>;
             constify_if<cartesian_product_view> * view_;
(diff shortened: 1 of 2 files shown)


include/range/v3/view/cartesian_product.hpp
```diff
@@ -193,15 +193,15 @@ namespace ranges
...
+                return std::get<0>(that.its_) - std::get<0>(its_);
             }
             template<std::size_t N>
-            auto distance_(cursor const & that, meta::size_t<N>) const
+            std::intmax_t distance_(cursor const & that, meta::size_t<N>) const
             {
-                auto d = distance_(that, meta::size_t<N - 1>{});
-                d *= ranges::distance(std::get<N - 2>(view_->views_));
+                std::intmax_t d = distance_(that, meta::size_t<N - 1>{});
+                d *= ranges::distance(std::get<N - 1>(view_->views_));
```
CaseyCarter: This line is the actual bug fix, hidden among the drive-by cleanups.

CaseyCarter:
There's another bug to fix here: `distance_` has a premise that none of the constituent ranges has size `0`. (Recall that the cross product of arbitrary sets is empty if any of the sets is empty.) Fix it, audit the rest of the view for misbehavior in this case, and add test coverage for "first range empty", "last range empty", and "some middle range empty".

ericniebler:
The constructor for cursor tests for empty ranges and sets itself risk to the end cursor. Doesn't that cover the case you're concerned about? At any rate, I'm all for adding these test cases.

CaseyCarter:
My panic is premature: I haven't broken the laws of mathematics. The proof doesn't need to be correct for two non-equal iterators when one of the constituent ranges is empty, because `cartesian_product_view` has no non-equal iterators when one of the constituent ranges is empty.

I may add some more test coverage in any case, and check-in the integer-overflow debugging paranoia that triggered this panic in the first place.

ericniebler:
Thank you, @CaseyCarter!

Issue #1287, opened 2019-08-26
Calendar example with Howard Hinnate's (C++20) date and also no boost.


davidsummers:
Would it be possible to modify the calendar example to:

1. Use Howard Hinnant's (C++20) date library.
2. Remove Boost dependencies and just use standard C++.

I would like to try to get the calendar example working with standard C++ only, if possible.

I have tried to substitute Howard Hinnnant's (C++20) date::sys_days for the calendar example's 'date' class which depends on Boost.

I've not been able to get the compiler to think that the date::sys_days is an "incrementable" class.

I'm a beginner on C++ templates so it is probably easy when you know the tricks (unlike me).

My thought was to use the date::sys_days as the calendar "date" class (which ultimately is just an integer as the representation behind the scenes ) and then convert to and from date::year_month_day when needed to display or get years, months, days, etc.

This seemed to me the easiest way to accomplish the equivalent of a "range" using begin/end iterators (which behind the scenes is just an integer).

If this could be accomplished, would it be acceptable to merge those changes into this projects?

What "magic" incantation could make date::sys_days "incrementable" according to the range library?

davidsummers:
Thanks for your time and attention. I think both you and Howard Hinnant (indrectly) may have helped me get past the hump.

I found the following template which I used to make an "int-like wrapper" class which now succeeds in passing the static-assert.

template<class T>
class intWrapperImpl {
    T value;
public:
    typedef T value_type;
    intWrapperImpl() :value() {}
    intWrapperImpl(T v) :value(v) {}
    operator T() const {return value;}
    //modifiers
    intWrapperImpl& operator=(T v) {value=v; return *this;}
    intWrapperImpl& operator+=(T v) {value+=v; return *this;}
    intWrapperImpl& operator-=(T v) {value-=v; return *this;}
    intWrapperImpl& operator*=(T v) {value*=value; return *this;}
    intWrapperImpl& operator/=(T v) {value/=value; return *this;}
    intWrapperImpl& operator%=(T v) {value%=value; return *this;}
    intWrapperImpl& operator++() {++value; return *this;}
    intWrapperImpl& operator--() {--value; return *this;}
    intWrapperImpl operator++(int) {return intWrapperImpl(value++);}
    intWrapperImpl operator--(int) {return intWrapperImpl(value--);}
    intWrapperImpl& operator&=(T v) {value&=v; return *this;}
    intWrapperImpl& operator|=(T v) {value|=v; return *this;}
[...]

PR #1288, opened 2019-08-26
No boost

davidsummers:
This addresses the first part of issue #1287 by removing the boost formatting and program_options dependencies.

This is just a quick hack.  I'm sure there are better ways to do it but this works.

Please review and let me know if other ways to do it are required to have this accepted into the range-v3 project.

### DIFF
--- a/example/calendar.cpp
+++ b/example/calendar.cpp
@@ -95,6 +95,8 @@ const int chars_per_week = 22;
 // Spaces per day
 const int spaces_per_day = 3;
 
+// Max number of weeks per month
+const int max_weeks = 6;
 
 // This template wrapper class provides the ability to act both as an
 // "incrementable" class which can be used with ranges and also a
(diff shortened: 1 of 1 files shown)


davidsummers:
With the last commit, I now have a working calendar example program without the requirement for BOOST.

Please consider merging this into the master branch and what changes to this code would be required to accomplish that.

This was just a quick hack and I'm sure there are probably better ways to do it but this works.

davidsummers:
How do I figure out how to fix the failed checks?

ericniebler:
I don't want to drop a full date/time library into the examples/ directory. I would love to see this example use [@HowardHinnant's date library](https://github.com/HowardHinnant/date), or standard facilities when available. I'm not sure what the options are for downloading it and/or taking a dependency on it.

I would be perfectly OK with dropping the dependency on a library for command line argument parsing.

johelegp:
CMake's `file(DOWNLOAD)` can do the job here.

ericniebler:
I have also had success with `ExternalProject_Add` (`include(ExternalProject)`) before.

johelegp:
[`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html) would be preferable to `ExternalProject`, but it was added in CMake 3.11. Since only a header is needed, `file(DOWNLOAD)` is saner than the dance needed to please `ExternalProject` usage.

PR #1298, opened 2019-09-04
Fix RANGES_CXX_STD handling in CMakeLists.txt

CaseyCarter:
...by treating the value numerically only if the string is a sequence of digits.


CaseyCarter:
Note to self: trim two AppVeyor configs to bring the full build time inline with Travis.

PR #1301, opened 2019-09-06
add view::cache1, clang-formatify

ericniebler:


doc/index.md
```diff
@@ -251,6 +251,8 @@ provides, and a blurb about how each is intended to be used.
   <DD>Type-erased range of elements with value type `T`; can store _any_ range with this value type.</DD>
 <DT>\link ranges::views::c_str_fn `views::c_str`\endlink</DT>
   <DD>View a `\0`-terminated C string (e.g. from a `const char*`) as a range.</DD>
+<DT>\link ranges::views::cache1_fn `views::cache1`\endlink</DT>
+  <DD>Caches the most recent element within the view so that dereferencing the view's iterator multiple types doesn't incur any recomputation. This can be useful in adaptor pipelines that includes combinations of `view::filter` and `view::transform`, for instance. `views::cache1` is always single-pass.</DD>
```
CaseyCarter:
```suggestion
  <DD>Caches the most recent element within the view so that dereferencing the view's iterator multiple types doesn't incur any recomputation. This can be useful in adaptor pipelines that include combinations of `view::filter` and `view::transform`, for instance. `views::cache1` is always single-pass.</DD>
```

include/range/v3/view/cache1.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+namespace ranges
+{
+    /// \addtogroup group-views
+    /// @{
+    template<typename Rng>
+    struct cache1_view : view_facade<cache1_view<Rng>, range_cardinality<Rng>::value>
+    {
+    private:
+        CPP_assert(input_range<Rng>);
```
CaseyCarter: The constraints on `range_value_t<Rng>` (IIRC `copy_constructible` for the `semiregular_box`) are missing here.

include/range/v3/view/cache1.hpp
```diff
@@ -0,0 +1,173 @@
...
+            bool equal(cursor const & that) const
+            {
+                return current_ == that.current_;
+            }
+            bool equal(sentinel const & that) const
+            {
+                return current_ == that.last_;
+            }
+            CPP_member
+            auto distance_to(cursor const & that) const -> CPP_ret(difference_type)( //
```
CaseyCarter: If this operation is used and it ever returns non-zero, it almost certainly indicates a bug in the calling code. I suggest dropping it.

include/range/v3/view/cache1.hpp
```diff
@@ -0,0 +1,173 @@
...
+    template<typename Rng>
+    cache1_view(Rng &&)->cache1_view<views::all_t<Rng>>;
+#endif
+
+    namespace views
+    {
+        struct cache1_fn
+        {
+            /// \brief Caches the most recent element within the view so that
+            /// dereferencing the view's iterator multiple types doesn't incur any
```
CaseyCarter:
```suggestion
            /// dereferencing the view's iterator multiple times doesn't incur any
```

include/range/v3/view/cache1.hpp
```diff
@@ -0,0 +1,173 @@
...
+            /// recomputation. This can be useful in adaptor pipelines that include
+            /// combinations of \c view::filter and \c view::transform, for instance.
+            /// \note \c views::cache1 is always single-pass.
+            template<typename Rng>
+            constexpr auto operator()(Rng && rng) const //
+                -> CPP_ret(cache1_view<all_t<Rng>>)(    //
+                    requires viewable_range<Rng> && input_range<Rng> &&
+                        constructible_from<range_value_t<Rng>, range_reference_t<Rng>>)
```
CaseyCarter: Missing my `copy_constructible` from above here, and this is missing from above.

include/range/v3/view/cache1.hpp
CaseyCarter: Returning a `cursor` from `end_cursor` makes this necessary. (I have vague oogie feelings about `common` single-pass ranges, but I'll drop this for now.)

ericniebler: My thinking here is that _someday_ `std::vector`'s constructor will be hep to `sized_sentinel_for` and use it to reserve memory _even if the iterator is not forward_. Wouldn't that be nice?

include/range/v3/view/cache1.hpp
ericniebler: I should have used `non_propagating_cache` instead, which doesn't have that requirement.

include/range/v3/view/cache1.hpp
CaseyCarter: Oops - I suppose I should actually resolve the comment after saying "I'll drop this."

include/range/v3/view/cache1.hpp
CaseyCarter: Still missing `constructible_from<range_value_t<Rng>, range_reference_t<Rng>>` here.

ericniebler: Fine, have it your way.

Issue #1292, opened 2019-08-29
regular_invocable and global variables


morinmorin:
# Background
I'm considering making function objects in adapted iterators (actually, `fun_` in `adaptor`) of `iter_transform_view<EmptySemiregularFunc>` have **zero size**. Whether this is OK or not depends on the definition of `regular_invocable`, I think.

# transform_view in C++20 and range-v3
- From my reading, function objects in C++20's `transform_view` are **not** copied to adapted iterators.
- Function objects in range-v3's `transform_view<F>` are not copied to adapted iterators in most cases; they are stored as `reference_wrapper`. But, when the function object is semiregular and the adapted iterator is const iterator, it is copied.

# `regular_invocable`
One interpretation of `regular_invocable` is that it can modify global variables if they do not affect the output. This means that `square` below models C++20 `regular_invocable`:
```
int num_calls{}; // global var.
int square(int x)
{
    ++num_calls;
    return x * x;
}
```

Under this interpretation, range-v3's transform_view behaves differently than C++20's transform_view; this program outputs `false` (even if we use `ranges::cpp20::views::transform_view`):
```
#include <iostream>
#include <range/v3/core.hpp>
#include <range/v3/view.hpp>

struct Square;
Square const* ptr; // global var.

struct Square { // **semiregular**
    int operator()(int x) const
    {
        ptr = this;
        return x * x;
    }
};

int main()
{
    int r[] = {1, 2, 3};
    auto const& v = r | ranges::views::transform(Square{}); // **const**
    [[maybe_unused]] auto x1 = *v.begin();
    auto p1 = ptr;
    [[maybe_unused]] auto x2 = *v.begin();
    auto p2 = ptr;
    std::cout << "p1 == p2 -> " << std::boolalpha << (p1 == p2) << "\n";
}
```
However, this should output `true` if we use `std::ranges::views::transform_view`. (In range-v3, `Square` in `transform_view` is copied to adapted iterators. In C++20, it is not copied.) Is this behavior non-conforming or just unspecified?

If the above interpretation of `regular_invocable` is incorrect, then I think empty semiregular function objects can be safely copied to adapted iterators. In this case, I would like to 

- add `[[no_unique_address]]` to `fun_` in `iter_transform_view::adaptor`:
  ```
  using fun_ref_ = semiregular_box_ref_or_val_t<Fun, IsConst>;
  RANGES_NO_UNIQUE_ADDRESS fun_ref_ fun_;
  ```

[...]

PR #1303, opened 2019-09-10
Enable EBO for transform_view's iterator and take_while_view's sentinel

morinmorin:
by making `semiregular_box_ref_or_val_t<T>` be just `T` if `T` is empty (and semiregular).
Drive by: fix typo in deprecation message.

Size changes in transform_view's iterator: 

|  views | before | after  |
|--------|:------:|:------:|
|<code>array &#124; transform</code>|16|8|
|<code>array &#124; transform &#124; transform</code>|24|8|
|<code>array &#124; transform &#124; transform &#124; transform</code>|32|8|

Size changes in take_while_view's sentinel:

|  views | before | after  |
|--------|:------:|:------:|
|<code>arr &#124; take_while</code>|16|8|

### Note
I think storing `semiregular_box_ref_or_val_t` (instead of a pointer) in transform_view's iterator is not conforming to C++20, since the difference is observable. Ditto for take_while_view's sentinel. I hope the Standard will allow such implementations eventually. Refs #1292.

### DIFF
--- a/include/range/v3/utility/semiregular_box.hpp
+++ b/include/range/v3/utility/semiregular_box.hpp
@@ -14,6 +14,7 @@
 #ifndef RANGES_V3_UTILITY_SEMIREGULAR_BOX_HPP
 #define RANGES_V3_UTILITY_SEMIREGULAR_BOX_HPP
 
+#include <type_traits>
 #include <utility>
 
 #include <meta/meta.hpp>
@@ -295,10 +296,11 @@ namespace ranges
     using semiregular_box_t = meta::if_c<(bool)semiregular<T>, T, semiregular_box<T>>;
 
     template<typename T, bool IsConst = false>
-    using semiregular_box_ref_or_val_t =
-        meta::if_c<(bool)semiregular<T>, meta::if_c<IsConst, T, reference_wrapper<T>>,
-                   reference_wrapper<meta::if_c<IsConst, semiregular_box<T> const,
-                                                semiregular_box<T>>>>;
+    using semiregular_box_ref_or_val_t = meta::if_c<
+        (bool)semiregular<T>,
+        meta::if_c<IsConst || std::is_empty<T>::value, T, reference_wrapper<T>>,
+        reference_wrapper<
+            meta::if_c<IsConst, semiregular_box<T> const, semiregular_box<T>>>>;
     /// @}
 
     /// \cond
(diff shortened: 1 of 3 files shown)


morinmorin:
Here is the code that checks iterator size.

It might be interesting that the size optimization works (perfectly) even for iterators of `arr | transform(F{}) | transform(F{}) | transform(F{})`.

```
#include <iostream>
#include <range/v3/core.hpp>
#include <range/v3/view.hpp>

struct F
{
    bool operator()(int x) const { return x % 2 == 0; }
};

template <typename R>
auto iterator_byte_size(R&& r)
{
    return sizeof(r.begin());
}

template <typename R>
auto sentinel_byte_size(R&& r)
{
    return sizeof(r.end());
}

int main()
{
    using namespace ranges::views;

    int arr[3] = {4, 5, 6};
    
    std::cout << "arr | transform" << "\n";
    std::cout << iterator_byte_size(arr | transform(F{})) << "\n";

    std::cout << "arr | transform | transform" << "\n";
    std::cout << iterator_byte_size(arr | transform(F{}) | transform(F{})) << "\n";

    std::cout << "arr | transform | transform | transform" << "\n";
    std::cout << iterator_byte_size(arr | transform(F{}) | transform(F{}) | transform(F{})) << "\n";

    std::cout << "arr | take_while" << "\n";
    std::cout << sentinel_byte_size(arr | take_while(F{})) << "\n";
}
```

morinmorin:
Thanks for merging.

> It might be interesting that the size optimization works (perfectly) even for iterators of `arr | transform(F{}) | transform(F{}) | transform(F{})`.

This is surely not an important use case, but let me explain this interesting result. 

At first sight, the iterator of the above view contains one `int*` and three `F`'s (as subobjects, or as subsubobjects, or ...). In this case, only one of the three `F`'s gets optimized out due to the rule of C++. But the size of this iterator is actually 8..., what happens here?

Ultimately, an adapted iterator is a struct whose data members are its base iterator and its adaptor. An adapted iterator **whose adaptor is empty** (and trivial) looks like
```
template <typename BaseIter, typename Adaptor>
struct AdaptedIterator {
    BaseIterator it;
    Coalesced<Adaptor> adpt;
};
```
where `Coalesced<Adaptor>` stores `Adaptor` as a **static data member**. (`Coalesced` is `ranges::box` in coalesce compression mode.) Note that `Coalesced<Adaptor>` does not contain any non-static data member; it's an empty type. 

For transform_view's iterator, `Adaptor` is a local struct in the view: `transform_view<…>::adaptor`. **The adaptor has distinct type is distinct for each distinct transform_view type.** This (type distinctness) and `Coalesced` do the trick for the size optimization. 

The iterator of `arr | transform(F{}) | transform(F{}) | transform(F{})` contains one `int*`, one `Coalesced<Adaptor1>`, one `Coalesced<Adaptor2>`, one `Coalesced<Adaptor3>`; the three `Coalesced<...>` are distinct empty types and so all of them can be optimized out. (As for `F`, each `Coalesced<…>` stores one `F` as a static data member).

PR #1312, opened 2019-09-12
add missing bind operations to actions::split and actions::split_when; fixes #1085

### LINKED ISSUE #1085: split an rvalue
The last line in 
```C++
#include <range/v3/all.hpp>

using namespace ranges;

extern std::vector<int> read_data();

auto vi = read_data();
auto vii = action::split(vi, 1);
auto vii2 = read_data() | action::sort;
auto vii3 = read_data() | action::split(1);
```

fails to compile with `/opt/compiler-explorer/libs/rangesv3/trunk/include/range/v3/action/action.hpp:123:22: note: candidate template ignored: substitution failure [with T = int, Rest = <>, A = ranges::v3::action::split_fn]: no matching function for call to 'bind'`

See https://gcc.godbolt.org/z/4DM3f8

ericniebler:


ericniebler:
@CaseyCarter When you get a sec, can you tell me why `cl.exe` has a hard time with this diff?

include/range/v3/action/insert.hpp
```diff
@@ -179,7 +179,7 @@ namespace ranges
                                        std::move(i),
                                        std::move(j),
                                        meta::bool_ < random_access_reservable<Cont> &&
-                                           sized_sentinel_for<S, I>> {});
+                                           sized_sentinel_for<S, I> > {});
```
CaseyCarter: lol - great job, clang-format.

PR #1323, opened 2019-09-15
Include iterate_range.hpp from all.hpp

berolinux:
Some applications (e.g. Telegram Desktop) include all.hpp and use
iterator_range -- this worked up until v0.5.0, but in 0.9.x and master,
results in an error about make_iterator_range being undefined.

Signed-off-by: Bernhard Rosenkraenzer <bero@lindev.ch>

### DIFF
--- a/include/range/v3/all.hpp
+++ b/include/range/v3/all.hpp
@@ -19,6 +19,7 @@
 #include <range/v3/core.hpp>
 #include <range/v3/functional.hpp>
 #include <range/v3/iterator.hpp>
+#include <range/v3/iterator_range.hpp>
 #include <range/v3/numeric.hpp>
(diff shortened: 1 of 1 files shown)


CaseyCarter:
It's very much intended that `iterator_range.hpp` is not included by the convenience headers. `ranges::iterator_range` is deprecated and replaced by `ranges::subrange`. The library still includes (pun intended) `iterator_range.hpp` to support old code that hasn't yet transitioned, but we deliberately don't make it easy to get to the deprecated component to make it harder for folks to inadvertently find and use it.

ericniebler:
Note: at some point I will probably also deprecate `<range/v3/all.hpp>`. It's best to include only what you use.

PR #1331, opened 2019-09-25
try to get clang to spit out better error messages without breaking the world

ericniebler:


.vscode/settings.json
```diff
@@ -84,7 +84,12 @@
         "__split_buffer": "cpp",
         "filesystem": "cpp",
         "queue": "cpp",
-        "stack": "cpp"
+        "stack": "cpp",
+        "__errc": "cpp",
+        "__hash_table": "cpp",
+        "__string": "cpp",
+        "algorithm": "cpp",
+        "codecvt": "cpp"
```
CaseyCarter: You really should put a `"**/include/c++/**": "cpp",` line in here so VSCode stops spamming every libc++ header here.

CaseyCarter: Also, ewww `<codecvt>`.

CaseyCarter:
This passes CI after merging my change from master, but it's still broken in non-`/std:c++latest` mode (Appveyor only covers `/std:c++latest` with cl). I suggest that you merge anyway.

ericniebler:
Interestingly, something about this PR improved compile times according to Travis. 🤷‍♂

Issue #1342, opened 2019-10-04
CMake targets are not exported into a namespace


MichaelWJung:
If I understand modern CMake practice correctly, exported targets should be in a namespace. A consumer of the library would then write something like the following in his CMakeLists.txt:
```
find_package(range-v3 REQUIRED)
add_executable(MyApp myapp.cpp)
target_link_libraries(MyApp
  PRIVATE
    range-v3::range-v3
)
```

For range-v3 this is currently not done, the exported targets will not be in a namespace. Is there a specific reason for this? I would love to see this changed.

To put the targets in a namespace, the following section would need to be changed in CMakeLists.txt:
```
install(EXPORT range-v3-targets FILE range-v3-config.cmake DESTINATION lib/cmake/range-v3)
```
to
```
install(EXPORT range-v3-targets NAMESPACE range-v3:: FILE range-v3-config.cmake DESTINATION lib/cmake/range-v3)
```

ericniebler:
Would this break existing users? I know nothing about cmake.

MichaelWJung:
The simple change from my post would no longer provide targets without a namespace, so yes, this would break existing users.

It is, however, easily possible to export the targets both with and without namespaces. I could provide a PR.

References:
Daniel Pfeifer recommends the pattern using namespaces in [his talk at C++Now 2017](https://youtu.be/bsXLMQ6WgIk?t=37m15s) ([slide](https://github.com/boostcon/cppnow_presentations_2017/blob/master/05-19-2017_friday/effective_cmake__daniel_pfeifer__cppnow_05-19-2017.pdf) 24ff.) It is also used in the [CMake reference for creating packages](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages).

ericniebler:
Breaking users is <sad face>. So yes, it sounds like this PR needs a PR.

PR #1343, opened 2019-10-05
Export CMake targets into range-v3:: namespace

MichaelWJung:
The old targets without namespace are kept in an additional targets file
for compatibility reasons.

Addresses issue #1342.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,9 +162,11 @@ write_basic_package_version_file(
 set(CMAKE_SIZEOF_VOID_P ${OLD_CMAKE_SIZEOF_VOID_P})
 
 install(TARGETS concepts meta range-v3 EXPORT range-v3-targets DESTINATION lib)
-install(EXPORT range-v3-targets FILE range-v3-config.cmake DESTINATION lib/cmake/range-v3)
+install(EXPORT range-v3-targets NAMESPACE range-v3:: FILE range-v3-targets.cmake DESTINATION lib/cmake/range-v3)
+install(EXPORT range-v3-targets FILE range-v3-targets-compat.cmake DESTINATION lib/cmake/range-v3)
 install(FILES
   ${CMAKE_CURRENT_BINARY_DIR}/range-v3-config-version.cmake
+  cmake/range-v3-config.cmake
   DESTINATION lib/cmake/range-v3)
 install(DIRECTORY include/ DESTINATION include FILES_MATCHING PATTERN "*")
 
--- a/cmake/range-v3-config.cmake
+++ b/cmake/range-v3-config.cmake
@@ -1,0 +1,2 @@
+include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets-compat.cmake")


MichaelWJung:
Looks like the build failure was not introduced by my changes.

johelegp:
This breaks libraries depending on range-v3 that want to be installable. Because now range-v3 is available from 2 export sets (from the 2 `install(EXPORT)` calls) and CMake can't decide which to use. See https://cmake.org/pipermail/cmake/2014-October/058837.html.

ericniebler:
So, this should be reverted in your opinion?

johelegp:
Not as a whole, just the attempt at compatibility. Appending `range-v3::` can't be much harder than renaming `Regular` to `regular`.

ericniebler:
I have never understood the motivation for the change in the first place. "Should be in a namespace" doesn't tell me why. I'm inclined to simply revert this diff unless someone can give me a good reason why I shouldn't.

MichaelWJung:
From my understanding, putting targets into namespaces is considered best practice in CMake mainly for the following reason:
When you add dependency `foo` to a target with a command like `target_link_libraries(mylib PRIVATE foo)`, CMake does not know the nature of `foo`. It could be another CMake target that we have not seen yet, or it could be an external library that is supposed to be added with `-lfoo` at link time. If no target with name `foo` is found at the end of the CMake run, it will assume the latter. The consequence of this is that, in case of a misspelled target `foo`, the error will only become visible at link time.
If a target added in `target_link_libraries` is within a namespace like in `target_link_libraries(mylib PRIVATE foo::bar)`, CMake will always assume that it is a target and will abort with an error if it does not find a target with name `foo::bar` by the end of the CMake run. Using namespaces thus helps us to find mistakes more easily and at an earlier time.

You can find this rule at the end of the *Overview* section in the [documentation of `target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html)

Another benefit of namespaces is the same as in C++: to avoid name clashes. Range-v3, e.g., has targets named `meta` and `concepts`. I could imagine targets with the same names in other projects as well. If they are going to use range-v3, they will an error like the following:

```
CMake Error at CMakeLists.txt:21 (add_library):
  add_library cannot create target "meta" because an imported target with the
  same name already exists.
```

I think the cleanest solution would be to only keep the namespaced targets. But I understand that you are concerned about changes that break users.

I was unaware of the problems that may arise if a target is in more than one export set. If you would like to proceed in the direction of namespaces plus compatibility, I would be happy to try to reproduce the problem and see if I can come up with another solution that maintains compatibility and does not show this issue.

johelegp:
@MichaelWJung is right.

> Another benefit of namespaces is the same as in C++: to avoid name clashes. Range-v3, e.g., has targets named `meta` and `concepts`. I could imagine targets with the same names in other projects as well. If they are going to use range-v3, they will an error like the following:
> 
> ```
> CMake Error at CMakeLists.txt:21 (add_library):
>   add_library cannot create target "meta" because an imported target with the
>   same name already exists.
> ```
> 
> I think the cleanest solution would be to only keep the namespaced targets. But I understand that you are concerned about changes that break users.

You still get name clashes unless you're using an installed version of ranges-v3, because the local buildsystem has the targets without namespace. I'd do what the CMake version of Boost is doing, which is naming the libraries something like `range-v3_meta` and adding an alias `range-v3::meta` (the `::` is the part that lets CMake know it's a target, because `::` can't be part of the name of a library to be linked in).

So that both installed versions and local versions of `range-v3` can be referred to by the same name, the local build should expose the libraries with `range-v3::` preprended, too, which is currently only done by the installed version.

> I was unaware of the problems that may arise if a target is in more than one export set. If you would like to proceed in the direction of namespaces plus compatibility, I would be happy to try to reproduce the problem and see if I can come up with another solution that maintains compatibility and does not show this issue.

One possible implementation is to move the `add_library(range-v3::meta ALIAS range-v3_meta)` and the like to their own CMake file and use [`include_guard()`](https://cmake.org/cmake/help/latest/command/include_guard.html).

PR #1344, opened 2019-10-07
Use SYSTEM in target_include_directories

smessmer:
I'm using ranges-v3 in one of my projects. It lives in a subdirectory and I use the cmake `add_subdirectory` command to add it. I added a `target_link_libraries` from my project to `range-v3`. This caused the range include directory to be added to the compiler command line with `-I`. Since my project uses quite high warning levels, this showed a lot of warnings in the ranges library. It should be included with `-isystem` instead. This PR shows the changes that were required for this to work for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,18 +21,18 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON) # Export compilation data-base
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 add_library(meta INTERFACE)
-target_include_directories(meta INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/>)
+target_include_directories(meta SYSTEM INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/>)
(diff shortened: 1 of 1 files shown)


ericniebler:
Hrm. I'm not convinced that `-isystem` is right for everybody. @CaseyCarter, do you have thoughts about this?

MichaelWJung:
This change looks wrong to me. Adding `SYSTEM` to something with `$<BUILD_INTERFACE:...>` should not affect consumers of the library at all (the generated `range-v3-targets.cmake` file does not change, I checked). It will, however, affect the builds in the range-v3 project. This means that the e.g. tests would be build with `-isystem`.

SYSTEM is set when adding the include directories with `$<INSTALL_INTERFACE:...>`. This already sets the include directories with `-isystem` for consumers.

smessmer:
@MichaelWJung What would be a better way of adding ranges-v3 to my project? I thought add_subdirectory is the preferred way.

MichaelWJung:
I don’t think there’s anything wrong with adding a subproject with `add_subdirectory`. But you then should not expect to get your include directories with `-isystem` by default as you treat the dependency more like a part of your own project in this case. I like to add my dependencies with a dependency manager like [Conan](https://conan.io/). You could also just configure the range-v3 cmake project in another directory (possibly excluding all the tests, demos, etc.) and run `make install`. If you then point the CMAKE_PREFIX_PATH of your project to the install of range-v3, you will find it via `find_package` and also get include directories with `-isystem`.

PR #1345, opened 2019-10-07
Don't muck with CMAKE_BUILD_TYPE

smessmer:
I use range-v3 and include it in a subdirectory of my build. I usually use `-DCMAKE_BUILD_TYPE=RelWithDebInfo`. The `range-v3` cmake script swoops in and says it doesn't recognize it, changes it to `CMAKE_BUILD_TYPE=Release` and changes my project config. I saw some parts of my project now being built with Release instead of RelWithDebInfo.

It's ok if range-v3 assumes that it's a release build if it doesn't recognize it, but it shouldn't modify the variable.


smessmer:
Closing for now, this didn't solve my issue. Need to investigate further.

PR #1348, opened 2019-10-17
integral compilation on the latest gcc fixed

mpusz:


include/meta/meta_fwd.hpp
```diff
@@ -340,7 +340,7 @@ namespace meta
         typename detail::require_constant<decltype(T{}()), T{}()>;
         requires T{}() == T::value;
 
-        { T{} } -> typename T::value_type;
+        // { T{} } -> typename T::value_type;
```
ericniebler: Or...

```suggestion
        { T{} } -> same_as<typename T::value_type>;
```

mpusz: Such a C++20 change will not work on gcc-9 :-(

I am not sure how this case relates to:
```cpp
// { T{}() } -> same_as<typename T::value_type>;
```
Is it the same constraint or another? I the same then it can be removed. If another than it can be probably expanded in a similar way to the one above:
```cpp
T{};
requires same_as<decltype(T{}), typename T::value_type>;
typename detail::require_constant<decltype(T{}), T{}>;
requires T{} == T::value;
```

ericniebler: This is all a bit messy, but the mess predates you. I'm inclined to just take this change and make a note to come back to this later and clean it up.

PR #1350, opened 2019-10-18
Only apply the GCC-91923 workaround in C++17 mode

### LINKED ISSUE #1349: Cannot build with GCC9 in C++14 mode due to use of `is_integral_v`
PR #1336 made use of `is_intergral_v` on GCC 9 (<9.3) for language standard <= C++17. This fails in C++14 mode because `std::is_integral_v` is in fact C++17 feature. 

```
../../range-v3/include/std/detail/associated_types.hpp:131:35: error: ‘is_integral_v’ is not a member of ‘std’; did you mean ‘is_integral’?
  131 |             std::enable_if_t<std::is_integral_v<difference_result_t<T>>>>
      |                                   ^~~~~~~~~~~~~
      |                                   is_integral
```

```
$ g++ --version
g++ (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008
```

CaseyCarter:
And, just for fun, let's ACTUALLY BUILD THE REGRESSION TEST.

Fixes #1349.


CaseyCarter:
> And, just for fun, let's ACTUALLY BUILD THE REGRESSION TEST.

I shouldn't be allowed near any C++ codebase.

PR #1352, opened 2019-10-19
New concepts syntax

CaseyCarter:
One commit that works around what might be a bug in MSVC (I spent an hour scraping the Standard to find definitive wording for template-ids that refer to overload sets, before deciding to simply avoid them), and another commit that will ruin your day.

[[namespace.udecl]/14](http://eel.is/c++draft/namespace.udecl#14):

> When a *using-declarator* brings declarations from a base class into a derived class, member functions and member function templates in the derived class override and/or hide member functions and member function templates with the same name, parameter-type-list, cv-qualification, and *ref-qualifier* (if any) in a base class (rather than conflicting). Such hidden or overridden declarations are excluded from the set of declarations introduced by the *using-declarator*.

Note that there's no mention here of associated constraints. Consequently this:
```c++
template <class T>
concept C1 = std::same_as<T, int>;

template <class T>
concept C2 = std::same_as<T, float>;

struct Base {
    template <class T>
    void f(T) requires C1<T> {}
};

struct Derived : Base {
    using Base::f;

    template <class T>
    void f(T) requires C2<T> {}
};

int main() {
    Derived{}.f(3.14f);
    Derived{}.f(42);
}
```
is ill-formed: the `f` in `Derived` hides the `f` in `Base`. If we replace `Derived` and `Base` with `join_fn` and `cpp20_join_fn`, we'll see why MSVC was failing to compile `view.join`: the `operator()(T&&)` template in `join_fn` hides the `operator()(T&&)` in `cpp20_join_fn`.

There's more to fix, but I'm stopping for a but and thought I'd push what I have for now.

### DIFF
--- a/include/range/v3/view/take_while.hpp
+++ b/include/range/v3/view/take_while.hpp
@@ -153,12 +153,10 @@ namespace ranges
             }
         };
 
-        struct take_while_fn : take_while_base_fn
+        struct take_while_bind_fn
         {
-            using take_while_base_fn::operator();
-
             template<typename Pred>
-            constexpr auto operator()(Pred pred) const
+            constexpr auto operator()(Pred pred) const // TODO: underconstrained
             {
                 return make_view_closure(
                     bind_back(take_while_base_fn{}, std::move(pred)));
@@ -166,11 +164,18 @@ namespace ranges
             template<typename Pred, typename Proj>
             constexpr auto CPP_fun(operator())(Pred && pred,
                                                Proj proj)(const //
-                                                          requires(!range<Pred>))
+                                                          requires(!range<Pred>)) // TODO: underconstrained
             {
                 return make_view_closure(bind_back(
                     take_while_base_fn{}, static_cast<Pred &&>(pred), std::move(proj)));
             }
+        };
+
+        struct RANGES_EMPTY_BASES take_while_fn
+          : take_while_base_fn, take_while_bind_fn
+        {
+            using take_while_base_fn::operator();
+            using take_while_bind_fn::operator();
         };
 
         /// \relates iter_take_while_fn
--- a/include/range/v3/view/trim.hpp
+++ b/include/range/v3/view/trim.hpp
@@ -120,23 +120,28 @@ namespace ranges
             }
         };
 
-        struct trim_fn : trim_base_fn
+        struct trim_bind_fn
         {
-            using trim_base_fn::operator();
-
             template<typename Pred>
-            constexpr auto operator()(Pred pred) const
+            constexpr auto operator()(Pred pred) const // TODO: underconstrained
             {
                 return make_view_closure(bind_back(trim_base_fn{}, std::move(pred)));
             }
             template<typename Pred, typename Proj>
             constexpr auto CPP_fun(operator())(Pred && pred,
                                                Proj proj)(const //
-                                                          requires(!range<Pred>))
+                                                          requires(!range<Pred>)) // TODO: underconstrained
             {
                 return make_view_closure(bind_back(
                     trim_base_fn{}, static_cast<Pred &&>(pred), std::move(proj)));
             }
+        };
+
+        struct RANGES_EMPTY_BASES trim_fn
+          : trim_base_fn, trim_bind_fn
+        {
+            using trim_base_fn::operator();
+            using trim_bind_fn::operator();
         };
 
         /// \relates trim_fn
--- a/include/concepts/concepts.hpp
+++ b/include/concepts/concepts.hpp
@@ -1098,13 +1098,16 @@ namespace concepts
         CPP_DIAGNOSTIC_POP
     } // namespace detail
 
-#if defined(__clang__) || defined(_MSC_VER)
+#if defined(__clang__)
     template<bool B>
     std::enable_if_t<B> requires_()
     {}
-#else
+#elif defined(__GNUC__)
     template<bool B>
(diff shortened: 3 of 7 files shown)


CaseyCarter:
> I'm stopping for a but

Err, "bit". Although I am hungry, so maybe "bite". I haven't been smoking for a few months, though, so definitely not a "butt."

ericniebler:
> Note that there's no mention here of associated constraints.

That _can't_ be the intent. Can it? Core issue?

ericniebler:
I would like to get @asutton's opinion for whether this is the intention, whether this is a core issue, or whether msvc is simply wrong.

CaseyCarter:
> That can't be the intent. Can it? Core issue?

Note that GCC doesn't implement this rule correctly even in the absence of constraints (https://godbolt.org/z/T8I7Jx). I don't know if that's an "implementor veto" or simply a bug, but it suggests that the behavior with constrained declarations may not be intentional design.

I think we can argue that the intent of `using Base::foo;` is to overload declarations of `foo` in this class with those in `Base`, and that the wording in [namespace.udecl]/14 is trying to exclude only those overloads that are a priori ambiguous. If that's actually the intent then constraints should be significant. But then return types and template-heads should likely be significant for function templates, since they often contain disambiguation material (old-style "constraints").

We can try to throw this mess at Core and see if it sticks, but I wouldn't be surprised if we end up writing a paper for EWG.

include/range/v3/view/take_while.hpp
```diff
@@ -151,25 +151,20 @@ namespace ranges
...
                         compose(std::move(pred), std::move(proj))};
             }
-        };
-
-        struct take_while_fn : take_while_base_fn
-        {
-            using take_while_base_fn::operator();
 
             template<typename Pred>
             constexpr auto operator()(Pred pred) const
```
CaseyCarter: `// TODO: underconstrained` (and 164)

include/range/v3/view/trim.hpp
```diff
@@ -118,24 +118,19 @@ namespace ranges
...
                         compose(std::move(pred), std::move(proj))};
             }
-        };
-
-        struct trim_fn : trim_base_fn
-        {
-            using trim_base_fn::operator();
 
             template<typename Pred>
             constexpr auto operator()(Pred pred) const
```
CaseyCarter: Ditto underconstrained (and 130).

ericniebler:
What about if _both_ the member functions are from different bases? What does msvc say about this?

```c++
template <class T>
concept C1 = std::same_as<T, int>;

template <class T>
concept C2 = std::same_as<T, float>;

struct B1 {
    template <class T>
    void f(T) requires C1<T> {}
};

struct B2 {
    template <class T>
    void f(T) requires C2<T> {}
};

struct D : B1, B2 {
  using B1::f;
  using B2::f;
};

int main() {
    D{}.f(3.14f);
    D{}.f(42);
}
```

include/range/v3/view/trim.hpp
```diff
@@ -118,24 +118,19 @@ namespace ranges
...
+                return make_view_closure(bind_back(trim_fn{}, std::move(pred)));
             }
             template<typename Pred, typename Proj>
             constexpr auto CPP_fun(operator())(Pred && pred,
                                                Proj proj)(const //
                                                           requires(!range<Pred>))
             {
                 return make_view_closure(bind_back(
-                    trim_base_fn{}, static_cast<Pred &&>(pred), std::move(proj)));
+                    trim_fn{}, static_cast<Pred &&>(pred), std::move(proj)));
```
ericniebler: The idea of using `trim_base_fn` here (and similar pattern elsewhere) was to narrow the number of `operator()` overloads that must be considered once partial application has happened. I would like to preserve this pattern.

CaseyCarter: Yes, I got that. This was an expedient fix. For the peanut gallery: I'll be reworking this PR to compose overload sets from multiple bases to avoid the [namespace.udecl] problem as @ericniebler and I discussed offline.

ericniebler:
So many test failures.

CaseyCarter:
> So many test failures.

Sorry: Neither I nor MSVC nor GCC noticed that `join_fn` was used before being completed (cursed forward declarations). The [Travis failures I see now](https://travis-ci.org/CaseyCarter/range-v3/builds/601005260) are [the same as the base branch](https://travis-ci.org/ericniebler/range-v3/builds/600098427).

CaseyCarter:
"ICEs - I'll try building with the debug compiler and see which assertions fail." \*Debug compiler ICEs also\*

I'm starting to really hate `range-v3/include/range/v3/action/concepts.hpp(84)`.

ericniebler:
Don't work about the gcc CONCEPTS=On failures. I'm working through those. I think I'm dropping support for clang < 3.9 and gcc-5 so you can ignore those failures, too.

ericniebler:
> I'm starting to really hate `range-v3/include/range/v3/action/concepts.hpp(84)`.

I will interpret that as, "Thank you for helping us flush out bugs in our compiler!"

By the way, I think I have fixed the gcc concepts build. It was getting tripped up on legit bogus concept definitions. You might want to pull in the new diff.

ericniebler:
Superseded by #1357

CaseyCarter:
For posterity: [p1971r0](https://wg21.link/p1971r0), which will likely be merged into the C++20 working draft on 2019-11-09, fixes [namespace.udecl]/14 by making *trailing-requires-clause*s significant.

PR #1355, opened 2019-10-23
added transform_join

cerww:
similar to flat_map in other languages
mostly same as transform | join but works when transform returns a lvalue container

### DIFF
--- a/include/range/v3/view/transform_join.hpp
+++ b/include/range/v3/view/transform_join.hpp
@@ -1,0 +1,261 @@
+#pragma once
+#include <range/v3/all.hpp>
+#include <optional>
+#include <variant>
+namespace ranges {
+	namespace detail {
+
+		template<typename Rng>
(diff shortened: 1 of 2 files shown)


ericniebler:
Are you aware of `views::for_each` ?

cerww:
I am aware of views::for_each. views::for_each uses join(transform(rng,fn)) and suffers the same as problem transform | join. when fn(rng) returns a rvalue(I think) container it doesn't compile.
The point of transform_join was to fix this problem.
IIUC, After seeing #1354 this can be implented with join(cache1(transform(rng,fn))) or rng | transform(fn) | cache1 | join

ericniebler:
> IIUC, After seeing #1354 this can be implented with join(cache1(transform(rng,fn))) or rng | transform(fn) | cache1 | join

Closing this as unnecessary.

PR #1365, opened 2019-10-29
use missing template argument in concepts

marehr:
gcc10 complained about this.

Introduced in https://github.com/ericniebler/range-v3/blame/121dad313e05aaa88b2839b0244172b0f7065ac3/include/range/v3/functional/concepts.hpp#L75


ericniebler:
Thank you!

PR #1369, opened 2019-11-06
Implementing D1870R1.

BRevzin:
This is an implementation of [D1871R1](http://wiki.edg.com/pub/Wg21belfast/LibraryEvolutionWorkingGroup/d1870r1.html) in range-v3. All tests pass.

This hasn't been run through clang-format, etc, and I didn't rename the forwarding range concept - the intent here was to demonstrate the viability of the patch.


BRevzin:
Sorry was apparently nowhere near master? Dunno what happened here, will redo.

ericniebler:
Trunk has problems on msvc and gcc-trunk, but should be a good option otherwise.

PR #1370, opened 2019-11-06
Implementing D1870R1.

BRevzin:
This is an implementation of [D1871R1](http://wiki.edg.com/pub/Wg21belfast/LibraryEvolutionWorkingGroup/d1870r1.html) in range-v3. All tests pass.

This hasn't been run through clang-format, etc, and I didn't rename the forwarding range concept - the intent here was to demonstrate the viability of the patch.


include/range/v3/range/access.hpp
```diff
@@ -31,11 +31,23 @@
...
+    {
+        CPP_def
+        (
+            template(typename T)
+            concept Safe,
+                std::is_lvalue_reference<T>::value ||
+                enable_safe_range<concepts::detail::remove_cvref_t<T>>
+        );
+    }
+
```
ericniebler: `CPP_def` is no longer a thing.

BRevzin: I'm confused, this isn't what the code looks like on my end.

BRevzin: ... because I didn't push it 🤦‍♂

include/range/v3/range/access.hpp
```diff
@@ -57,11 +66,10 @@ namespace ranges
...
         template<typename T>
         CPP_concept_bool has_member_begin =
-            CPP_requires ((T &) t) //
+            CPP_requires ((T &&) t) //
             (
-                _begin_::is_iterator(t.begin())
-            ) &&
-            std::is_lvalue_reference<T>::value;
+                _begin_::is_iterator(CPP_fwd(t).begin())
+            );
```
CaseyCarter: Shouldn't we be simplifying `has_non_member_begin` to work with lvalues rather than adding perfect-forwarding to `has_member_begin`?

include/range/v3/range/access.hpp
```diff
@@ -197,9 +205,8 @@ namespace ranges
         CPP_concept_bool has_member_end =
             CPP_requires ((T &) t) //
             (
-                _end_::is_sentinel<_begin_::_t<CPP_type(T) &>>(t.end())
-            ) &&
-            std::is_lvalue_reference<T>::value;
+                _end_::is_sentinel<_begin_::_t<CPP_type(T)>>(CPP_fwd(t).end())
```
CaseyCarter: Both `CPP_fwd(t)` and `t` are lvalues of type `t` - this doesn't seem to change anything.

include/range/v3/range/access.hpp
```diff
@@ -515,9 +521,8 @@ namespace ranges
         CPP_concept_bool has_member_rend =
             CPP_requires ((T &) t) //
             (
-                _end_::is_sentinel<_rbegin_::_t<CPP_type(T) &>>(t.rend())
-            ) &&
-            std::is_lvalue_reference<T>::value;
+                _end_::is_sentinel<_rbegin_::_t<CPP_type(T) &>>(CPP_fwd(t).rend())
```
CaseyCarter: Ditto "Why are we perfect-forwarding an lvalue?"

include/range/v3/range/access.hpp
BRevzin: I'm not sure I understand the question. The intent is to forward the range expression through both cases.

include/range/v3/range/access.hpp
BRevzin: Fixed this (and the next one) to actually be forwarding refs.

include/range/v3/range/access.hpp
ericniebler: Nit: we don't use PascalCase for concepts any longer.

include/range/v3/range/access.hpp
ericniebler: I think @CaseyCarter is expecting us to promote rvalues to lvalues. I don't think that's necessary, and I like @BRevzin's formulation. Casey, is there a reason why we want to be promoting rvalues to lvalues?

include/range/v3/range/access.hpp
ericniebler:
```suggestion
                _end_::is_sentinel<_rbegin_::_t<CPP_type(T)>>(CPP_fwd(t).rend())
```

include/range/v3/range/access.hpp
BRevzin: Renamed to lower.

include/range/v3/range/access.hpp
CaseyCarter: I'm envisioning a design more like the pre-`forwarding-range` design in which the CPOs always dispatch lvalue arguments to the "hook" functions so implementors of range types - even `safe_range` types - need not worry about accepting rvalues in `begin` and `end`. I'm not sure *why* I'm envisioning that, given that every model of `safe_range` I've ever seen has member `begin`/`end` that are callable on rvalues. 

I, uh, want to eliminate the runtime cost of all of the xvalue casts. Yeah, that's it - those crazy expensive casts.

ericniebler:
With the Better, Safer redesign of `ranges::begin`, we changed the signature of the poison pill. Do we now need to change it back? Why, or why not?

CaseyCarter:
> With the Better, Safer redesign of `ranges::begin`, we changed the signature of the poison pill. Do we now need to change it back? Why, or why not?

I think `ranges::begin` (`ranges::end`, etc.) now perform ADL only for rvalue `safe_range`s, so we no longer need to poison rvalues. There's not a *need* to change the poison pills back, but we *could* do so.

ericniebler:
> I think ranges::begin (ranges::end, etc.) now perform ADL only for rvalue safe_ranges

I hope not! I would expect ADL to be tried on toes that don't provide member `begin()`, even lvalues.

CaseyCarter:
> > I think ranges::begin (ranges::end, etc.) now perform ADL only for rvalue safe_ranges
> 
> I hope not! I would expect ADL to be tried on types that don't provide member `begin()`, even lvalues.

Attempt#2: "...perform ADL for rvalues only if they model `safe_range`, so ..."

CaseyCarter:
> > > I think ranges::begin (ranges::end, etc.) now perform ADL only for rvalue safe_ranges
> > 
> > 
> > I hope not! I would expect ADL to be tried on types that don't provide member `begin()`, even lvalues.
> 
> Attempt#2: "...perform ADL for rvalues only if they model `safe_range`, so ..."

No, Casey - you're reversing the requirements. The CPOs perform ADL for rvalues only if `enable_safe_range` is specialized to `true` for their decayed type. Such rvalues in fact model `safe_range` only if `ranges::begin` and `ranges::end` are well-formed for rvalues and lvalues of the same `remove_reference_t`-ed type.

The effects on the poison pills are the same: we don't need to block ADL lookup for rvalues, since the CPO don't so much as attempt rvalue lookup except for types that specialize `enable_safe_range`. I don't think we need the `initializer_list` poison pills at all.

CaseyCarter:
> Such rvalues in fact model `safe_range` only if `ranges::begin` and `ranges::end` are well-formed for rvalues and lvalues of the same `remove_reference_t`-ed type.

Hmmm - this may have been my motivation for suggesting that the CPOs should always promote rvalues to lvalues. If we define `safe_range<T>` as `range<T> && (is_lvalue_reference_t<T> || enable_safe_range<remove_cvref_t<T>>)` and promote rvalues to lvalues in the CPOs, we can fold `range-impl` back into `range` and checking `safe_range_t<R>` requires two overload resolutions instead of four.

ericniebler:
> ... and checking safe_range_t<R> requires two overload resolutions instead of four.

This needs a paper.

BRevzin:
> This needs a paper.

We actually already require that the lvalue and rvalue forms of `begin` mean the same thing (http://www.eel.is/c++draft/ranges#range.range-5), so we wouldn't lose any use-cases by not even invoking the rvalue version. This also goes to the argument that @ericniebler made in LEWG that doing something different on rvalues (e.g. I suggested the example of returning move iterators) is highly questionable.

[R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1870r0.html) of this paper had the version of concept that @CaseyCarter is suggesting, and had wording to do the rvalue-to-lvalue promotion (if you squint and ignore that it did this by first rejecting rvalues).

I think this would be a better specification, but not a meaningful change in design, in the sense that it doesn't change what you can and cannot do with Ranges... and would suggest I just change R1 to promote rvalues to lvalues.

Agree/disagree?

ericniebler:
Sounds legit.

BRevzin:
New draft on the Library wiki: [R1](http://wiki.edg.com/pub/Wg21belfast/LibraryWorkingGroup/d1870r1.html#wording).

BRevzin:
Implemented that direction, I think.

include/range/v3/range/access.hpp
```diff
@@ -107,16 +115,16 @@ namespace ranges
...
-                requires(has_member_begin<R> || has_non_member_begin<R>))
+                noexcept(noexcept(impl<R>{}(r))) //
+                requires(_safe_::safe<R> && (has_member_begin<R> || has_non_member_begin<R>)))
 #endif
             {
-                return impl<R>{}((R &&) r);
+                return impl<R>{}(r);
             }
 
 #if defined(__cpp_lib_string_view) && __cpp_lib_string_view > 0
```
ericniebler: Perhaps it would be better now to implement this as a partial specialization of `enable_safe_view` on `basic_string_view` instead of this ugly hack in the `begin` CPO.

include/range/v3/range/access.hpp
```diff
@@ -31,11 +31,20 @@
...
 #include <range/v3/detail/disable_warnings.hpp>
 
 namespace ranges
 {
+    namespace _safe_
+    {
+        template<typename T>
+        CPP_concept_bool safe =
+                std::is_lvalue_reference<T>::value ||
+                enable_safe_range<concepts::detail::remove_cvref_t<T>>;
```
ericniebler: I think I would prefer this:

```c++
template<typename T>
RANGES_INLINE_VAR constexpr bool safe =
    enable_safe_range<concepts::detail::remove_cvref_t<T>>;

template<typename T>
RANGES_INLINE_VAR constexpr bool safe<T &> = true;
```

No need to instantiate `std::is_lvalue_reference`.

ericniebler:
@BRevzin heads up I pushed to your branch.

PR #1374, opened 2019-11-22
Fix warning about shadowing in adjacent_filter

alkino:
Removing 2 warnings with gcc 6.2.0
`include/range/v3/view/adjacent_filter.hpp:97:26: error: declaration of 'prev' shadows a member of 'ranges::adjacent_filter_view<Rng, Pred>::adaptor<Const>' [-Werror=shadow]`

### DIFF
--- a/cmake/ranges_flags.cmake
+++ b/cmake/ranges_flags.cmake
@@ -131,7 +131,6 @@ if (RANGES_DEBUG_BUILD AND RANGES_RELEASE_BUILD)
 endif()
 
 if (RANGES_DEBUG_BUILD)
-  ranges_append_flag(RANGES_HAS_O0 -O0)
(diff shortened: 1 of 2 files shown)


alkino:
Close and reopen to trigger travis again

cmake/ranges_flags.cmake
```diff
@@ -131,7 +131,6 @@ if (RANGES_DEBUG_BUILD AND RANGES_RELEASE_BUILD)
 endif()
 
 if (RANGES_DEBUG_BUILD)
-  ranges_append_flag(RANGES_HAS_O0 -O0)
```
ericniebler: What is the reason for this?

alkino:
This error: https://ci.appveyor.com/project/ericniebler/range-v3/builds/29182581/job/h1hkkvrekwun9lqt#L164

`-O0` is not portable and by the way cmake handle this

PR #1376, opened 2019-11-28
Fix group_by to exclude the first element from the checks

vgasparyan1995:
As the description of the `views::group_by` suggests ([Range-v3: User Manual](https://ericniebler.github.io/range-v3/index.html):

> Given a source range and a binary predicate, return a range of ranges where each range contains contiguous elements from the source range such that the following condition holds: for each element in the range **apart from the first**, when that element and the first element are passed to the binary predicate, the result is true. In essence, views::group_by groups contiguous elements together with a binary predicate.

This '_apart from the first_' part is crucial and enables to `group_by` a range like `{ 0, 1, 2, 0, 1, 2}` with a `std::less` predicate and get `{ { 0, 1, 2}, {0, 1, 2} }`.

Without this fix, the resulting view infinitely returns empty ranges.

### DIFF
--- a/include/range/v3/view/group_by.hpp
+++ b/include/range/v3/view/group_by.hpp
@@ -30,7 +30,9 @@
 #include <range/v3/range/traits.hpp>
 #include <range/v3/utility/semiregular_box.hpp>
 #include <range/v3/utility/static_const.hpp>
+#include <range/v3/view/concat.hpp>
 #include <range/v3/view/facade.hpp>
+#include <range/v3/view/single.hpp>
 #include <range/v3/view/subrange.hpp>
 #include <range/v3/view/take_while.hpp>
 #include <range/v3/view/view.hpp>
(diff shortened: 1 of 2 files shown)


vgasparyan1995:
Closing and reopening the PR to trigger checks. Travis failed with such an error (unrelated to the commit):
`Error: retrieving gpg key timed out.`

ericniebler:
`views::concat` is a big hammer. It comes with both a compile-time and a runtime cost. I wonder if there is another way. I also don't yet really grok why the current implementation is failing here, but I haven't investigated.

vgasparyan1995:
The current implementation takes the first element of the remaining view/range, binds it to the binary predicate to get a unary one, which then is used to do a `take_while`. As a result, the first element is compared with itself, then the rest.
In the classic case of grouping with a `std::equal` predicate this works, as the first comparison is always true. But in case of the `std::less` or something else that returns `false`, `take_while` will return an empty view and the "current" iterator won't move forward.

One way I had in mind was to create a variation of the `take_while` view that takes a binary predicate, takes the first one anyway and evaluates the predicate for adjacent pairs, very much like the `adjacent_filter`. I'll give this a try.

Hope this is helpful.

include/range/v3/view/group_by.hpp
```diff
@@ -73,30 +71,32 @@ namespace ranges
             {
                 iterator_t<CRng> first_;
                 semiregular_box_ref_or_val_t<Fun, IsConst> fun_;
+                mutable bool first_invokation_{true};
```
johelegp: Current practice is to not use mutable data members and use non-`const` member functions instead. But that seems like a breaking change.

vgasparyan1995: How so? Just changed it, the test passes fine. Will commit now.

johelegp: It might be observable behaviour. But this is a bug fix, so it's justified.

vgasparyan1995: Hi @johelegp,

After my last change (to remove the mutable and make the `pred::operator()` non-const) I've faced another issue. In such a code:
```
const auto groups = /*...*/ | views::group_by(/*...*/);
for (const auto group : groups) { // Won't compile! Does only after removing the const.
    // ...
}
```
Would you agree that this is unacceptable and unexpected?
And if so, wouldn't it be an acceptable tradeoff to make that `prad::operator()` const again with a mutable member?

johelegp: I'd say the current state is the right one. Views should be non-`const`.

ericniebler:
Fixed by #1395

PR #1378, opened 2019-12-03
Deal with the case Boost_LIBRARY_DIRS is empty

xuning97:
deal with possible Boost_LIBRARY_DIRS empty case,
which happens when I use a local compiled boost 1.71.0 installed to /usr/locl/lib.


example/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ if(RANGES_BUILD_CALENDAR_EXAMPLE)
         message(STATUS "boost: ${Boost_LIBRARY_DIRS}")
         target_compile_definitions(calendar PRIVATE BOOST_NO_AUTO_PTR)
         target_compile_options(calendar PRIVATE -std=gnu++14)
-        set_target_properties(calendar PROPERTIES LINK_FLAGS "-L${Boost_LIBRARY_DIRS}")
+        if (${Boost_LIBRARY_DIRS})
+          set_target_properties(calendar PROPERTIES LINK_FLAGS "-L${Boost_LIBRARY_DIRS}")
+        endif()
```
ericniebler: How does this work? The `calendar` target needs to link to the Boost.DateTime and Boost.ProgramOptions libraries. Does it even make sense to build the `calendar` target if those libraries aren't there?

johelegp: This checks if `Boost_LIBRARY_DIRS` is set. I guess the local installation of @xuning97 doesn't include a CMake package.

ericniebler: I still don't get how `Boost_FOUND` can be true but `Boost_LIBRARY_DIRS` is empty.

johelegp: Maybe if it uses [Boost CMake](https://cmake.org/cmake/help/latest/module/FindBoost.html#boost-cmake), assuming it doesn't set it.

johelegp: Searching for it in its organization turns no matches: https://github.com/search?q=org%3Aboost-cmake+Boost_LIBRARY_DIRS&unscoped_q=Boost_LIBRARY_DIRS.

ericniebler: It is set by FindBoost.cmake, which is a standard part of every CMake install: https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake

johelegp: As stated in the link above for Boost CMake, the [logic](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L426) does use the installed Boost, which does not set `Boost_LIBRARY_DIRS`.

example/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ if(RANGES_BUILD_CALENDAR_EXAMPLE)
         message(STATUS "boost: ${Boost_LIBRARY_DIRS}")
         target_compile_definitions(calendar PRIVATE BOOST_NO_AUTO_PTR)
         target_compile_options(calendar PRIVATE -std=gnu++14)
-        set_target_properties(calendar PROPERTIES LINK_FLAGS "-L${Boost_LIBRARY_DIRS}")
+        if (${Boost_LIBRARY_DIRS})
+          set_target_properties(calendar PROPERTIES LINK_FLAGS "-L${Boost_LIBRARY_DIRS}")
```
johelegp:
```suggestion
        target_link_libraries(calendar PRIVATE Boost::date_time Boost::program_options)
```
Based on the discussion on the line below, this is how it should look like. Discard the other changes.

example/CMakeLists.txt
xuning97: There is no Boost_LIBRARY_DIRS in the cmake file shipped with boost sources.
https://github.com/boostorg/boost_install/blob/develop/BoostConfig.cmake

example/CMakeLists.txt
xuning97: Not a master in this area, Can this type of change support older version of boost?

johelegp: What actually matter is the CMake support for the Boost targets, which were [added in CMake 3.5](https://cmake.org/cmake/help/latest/release/3.5.html#modules). Range-v3 requires 3.6, so it's alright.

xuning97: Updated this PR.

ericniebler:
Very nice, thank you.

Issue #1337, opened 2019-09-29
ranges::to is significantly slower than direct construction


cjdb:
I expected these two implementations to have equivalent or relatively comparable performance, but Google Benchmark tells a different story when run over 476k words. Is this a perf bug or have I done something horribly wrong in making these two programs equivalent?

```cpp
auto palindrome_range_common(std::vector<std::string> const& words)
{
	auto is_palindrome = [](auto const& word) {
		return not ranges::empty(word)
		   and ranges::equal(word, word | views::reverse);
	};
	
	auto palindrome_excalim = [&is_palindrome](auto const& word) {
		return is_palindrome(word) ? word + '!' : word;
	};
	
	auto result = words
	            | views::transform(palindrome_excalim)
	            | views::common;
	
	return std::vector<std::string>{
		ranges::begin(result),
		ranges::end(result)
	};	
}

	
auto palindrome_range_to(std::vector<std::string> const& words)
{
	auto is_palindrome = [](auto const& word) {
		return not ranges::empty(word)
		   and ranges::equal(word, word | views::reverse);
	};
	
	auto palindrome_excalim = [&is_palindrome](auto const& word) {
		return is_palindrome(word) ? word + '!' : word;
	};
	
	return words
	     | views::transform(palindrome_excalim)
	     | ranges::to<std::vector>; // range-v3 extension	
}
```

```
Running ./benchmark-palindromes-o2
Run on (2 X 2684.42 MHz CPU s)
CPU Caches:
  L1 Data 32K (x1)
  L1 Instruction 32K (x1)
  L2 Unified 256K (x1)
  L3 Unified 4096K (x1)
Load Average: 0.69, 0.63, 0.42
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
benchmark_solutions/handrolled     12443168 ns     12439882 ns         7113    # base
benchmark_solutions/algorithm      25840577 ns     25831195 ns         3255
benchmark_solutions/range_common   10816578 ns     10813486 ns         7474
benchmark_solutions/range_to       23025614 ns     23020478 ns         3303
```
```
Running ./benchmark-palindromes-o3
Run on (2 X 2684.42 MHz CPU s)
CPU Caches:
  L1 Data 32K (x1)
  L1 Instruction 32K (x1)
  L2 Unified 256K (x1)
  L3 Unified 4096K (x1)
Load Average: 0.05, 0.36, 0.47
[...]

PR #1381, opened 2019-12-05
add cmake magic to download google-benchmark, implement benchmark for #1337

ericniebler:


ericniebler:
@cjdb I invite you to comment on this benchmark, in regards to #1337.

perf/range_conversion.cpp
```diff
@@ -0,0 +1,100 @@
...
+        auto magic = 476'000u;
+        words_.reserve(magic);
+        for(auto i = 0u; i < magic; ++i)
+        {
+            words_.push_back("this");
+            words_.push_back("is");
+            words_.push_back("his");
+            words_.push_back("face");
+            words_.push_back("abba");
+            words_.push_back("toot");
```
cjdb: I read in a dictionary of words from file, but this will make things easier, and apparently still highlights the issue.

perf/range_conversion.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+};
+
+BENCHMARK_F(Words, RangeCommon)(benchmark::State & st)
+{
+    for(auto _ : st)
+    {
+        auto words = ::palindrome_range_common(words_);
+        benchmark::DoNotOptimize(words.data());
```
cjdb: What is the difference between `benchmark::DoNotOptimize(words)` and `benchmark::DoNotOptimize(words.data())`? I did the former.

perf/range_conversion.cpp
```diff
@@ -0,0 +1,100 @@
...
+        benchmark::ClobberMemory();
```
cjdb: I also forgot to clobber, but I suspect that would only make my times look better?

perf/range_conversion.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+BENCHMARK_F(Words, RangeTo)(benchmark::State & st)
+{
+    for(auto _ : st)
+    {
+        auto words = ::palindrome_range_to(words_);
+        benchmark::DoNotOptimize(words_.data());
```
cjdb:
```suggestion
        benchmark::DoNotOptimize(words.data());
```

cjdb: Alternatively, `s/\<words\>/result/g`.

cjdb: This is the only comment that triggered "Request changes".

perf/range_conversion.cpp
```diff
@@ -0,0 +1,100 @@
...
+//
+//  Copyright Eric Niebler 2019-present
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+//
```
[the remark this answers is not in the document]
ericniebler: BSL is "free for all uses, and you don't even need to include this license but you do have to include copyright." Speaking of which, this code is also yours so I'll add a copyright notice for you unless you say otherwise.

perf/range_conversion.cpp
ericniebler: I'm not sure, but I think the latter is necessary for `ClobberMemory` to have any meaningful effects (if I'm reading it right).

perf/range_conversion.cpp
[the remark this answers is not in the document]
cjdb: SGTM (I use this code in classrooms).

PR #1382, opened 2019-12-05
coerce the value written through ostream_iterator to the right type

### LINKED ISSUE #1380: ranges::ostream_iterator and type conversion
The following code:

    #include <iostream>

    #include <range/v3/all.hpp>
    using namespace ranges;

    int main () {
        ostream_iterator< double > iter (std::cout, " ");
        *iter = char (-1);
        std::cout << " : ";

        std::ostream_iterator< double > iter2 (std::cout, " ");
        *iter2 = char (-1);
        std::cout << std::endl;

        return 0;
    }

prints `\377 : -1`

I expected the value to be cast to the value_type of the iterator, i.e. double, before the actual insertion in the stream. If my expectation and/or usage of `ranges::ostream_iterator` is wrong where lies the error? Thanks.

ericniebler:
fixes #1380

### DIFF
--- a/include/range/v3/iterator/stream_iterators.hpp
+++ b/include/range/v3/iterator/stream_iterators.hpp
@@ -54,7 +54,7 @@ namespace ranges
             requires convertible_to<U, value_t<U> const &>)
         {
             RANGES_EXPECT(sout_);
-            *sout_ << value;
+            *sout_ << static_cast<value_t<U> const &>(static_cast<U &&>(value));
             if(delim_)
                 *sout_ << delim_;
             return *this;


johelegp:
But why is `operator=` a template? `std::`'s isn't.

ericniebler:
`ranges::ostream_iterator` has a trick/feature that `std::ostream_iterator` doesn't. If you don't specify that value type (or specify it as `void`), you can write *anything* through it. It's handy.

johelegp:
If only the specialization when `std::is_void<T>::value` is  `true` is supposed to have that feature, `operator=` should check that before deciding whether to cast to `T`. The current solution seems to have `U` mistyped instead of `T`, and doesn't check for `void`.

ericniebler:
You're mistaken. Look at the definition of the `value_t` alias.

PR #1385, opened 2019-12-06
Fix ::-targets for installed packages

### LINKED ISSUE #1384: CMake configuration failure
After updating `range-v3` to the current `master` I cannot configure any cmake projects using it. The issue can be reproduced with the following `CMakeLists.txt`
```
cmake_minimum_required(VERSION 3.14)

project(Test)

find_package(range-v3)
```
The error message is
```
CMake Error at /usr/lib64/cmake/range-v3/range-v3-config.cmake:2 (add_library):
  add_library cannot create ALIAS target "range-v3::meta" because target
  "range-v3-meta" is imported but not globally visible.
  CMakeLists.txt:5 (find_package)


CMake Error at /usr/lib64/cmake/range-v3/range-v3-config.cmake:3 (add_library):
  add_library cannot create ALIAS target "range-v3::concepts" because target
  "range-v3-concepts" is imported but not globally visible.
  CMakeLists.txt:5 (find_package)


CMake Error at /usr/lib64/cmake/range-v3/range-v3-config.cmake:4 (add_library):
  add_library cannot create ALIAS target "range-v3::range-v3" because target
  "range-v3" is imported but not globally visible.
  CMakeLists.txt:5 (find_package)
```

johelegp:
Fixes #1384.

Reading https://cmake.org/pipermail/cmake-developers/2017-June/030105.html, this seems to be the way to go (as there shouldn't be another way of making this work). Otherwise, we'd have to look how https://github.com/boost-cmake/bcm does it.

I have tested this with a local build and an installed package. I have been meaning to create a CMake module that makes a package available in all possible ways to catch this kind of issues, but haven't gotten around it yet.


ericniebler:
Thanks. I hope this really (really) fixes it this time.

PR #1391, opened 2019-12-13
fix warnings

cyyever:


cyyever:
> Looks like `view_<Rng>` is asserted in the same expression. Nice catch. What compiler/warning combination told you about it? I'd have kept the first occurrence and deleted the second one as that's the usual order.

I use that[pvs](https://www.viva64.com/en/pvs-studio/)

ericniebler:
Fixed in a different way by #1395.

PR #1392, opened 2019-12-13
Add a step parameter to iota_view

### LINKED ISSUE #723: step size for monotonically increasing values
I'm looking for a view that does the same as [boost::irange][irange]:

```c++
boost::irange(3, 14, 5) // 3 8 13
```

I only found `ints` and `iota`. None of them takes a `step_size`. It is always `1`. Have I overlooked something or is this feature missing?

[irange]: http://www.boost.org/doc/libs/1_60_0/libs/range/doc/html/range/reference/ranges/irange.html

cor3ntin:
During LEWGI discussions about P1894R0, I suggested adding
a step parameter to iota_view so that it behaves as Python's range
function.

This patches implement that feature and fixes #723.

Iota views constructed from totally ordered advancable values can
take an extra step parameter which can be positive or negative.

### DIFF
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -414,12 +414,6 @@ namespace ranges
                 return from_ == that.to_;
             }
             CPP_member
-            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
-                requires detail::iota_stridable_<From, To>)
-            {
-                return detail::iota_distance_(from_, that.to_) >= detail::iota_abs_(step_.value());
-            }
-            CPP_member
             auto equal(cursor const & that) const -> CPP_ret(bool)( //
             requires equality_comparable<From>)
             {
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -421,20 +421,13 @@ namespace ranges
             }
             CPP_member
             auto equal(cursor const & that) const -> CPP_ret(bool)( //
-            requires (equality_comparable<From> &&!detail::iota_stridable_<From, To>))
+            requires equality_comparable<From>)
             {
                 return that.from_ == from_;
             }
             CPP_member
-            auto equal(cursor const & that) const -> CPP_ret(bool)( //
-            requires detail::iota_stridable_<From, To>)
-            {
-                return from_ == that.from_;
-            }
-            CPP_member
             auto prev() -> CPP_ret(void)( //
-                requires(detail::decrementable_<From> &&
-                         (!detail::iota_stridable_<From, To>)))
+            requires(detail::decrementable_<From> && !detail::iota_stridable_<From, To>))
             {
                 --from_;
             }
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -429,21 +429,7 @@ namespace ranges
             auto equal(cursor const & that) const -> CPP_ret(bool)( //
             requires detail::iota_stridable_<From, To>)
             {
-                if(from_ == that.from_)
-                    return true;
-
-                if(detail::iota_abs_(step_.value() == 1))
-                    return false;
-
-                //If the step is not 1, we may overshoot the end cursor
-                //which may not be aligned to a multiple of step
-                if(step_.value() > 0 && from_ > that.from_) {
-                    return detail::iota_distance_(that.from_, from_) < step_.value();
-                }
-                else if(step_.value() < 0 && that.from_ > from_) {
-                    return detail::iota_distance_(from_, that.from_) < -(step_.value());
-                }
-                return false;
+                return from_ == that.from_;
             }
             CPP_member
             auto prev() -> CPP_ret(void)( //
@@ -472,7 +458,7 @@ namespace ranges
                 requires detail::advanceable_<From>)
             {
                 const auto d = detail::iota_distance_(from_, that.from_);
-                return (d / step_.value()) + (d % step_.value() ? 1 : 0);
+                return d / step_.value();
             }
             // Extension: see https://github.com/ericniebler/stl2/issues/613
             CPP_member
@@ -480,7 +466,7 @@ namespace ranges
                 requires sized_sentinel_for<To, From>)
             {
                 const auto d = that.to_ - from_;
-                return (d / step_.value()) + (d % step_.value() ? 1 : 0);
+                return d / step_.value();
             }
 
         public:
@@ -506,10 +492,21 @@ namespace ranges
             return sentinel{to_};
         }
         CPP_member
-        auto CPP_fun(end_cursor)()(const requires same_as<From, To>)
+        auto CPP_fun(end_cursor)()(const requires (same_as<From, To> && !detail::iota_stridable_<From, To>))
         {
             return cursor{to_, step_};
         }
+        CPP_member
+            auto CPP_fun(end_cursor)()(const requires detail::iota_stridable_<From, To>)
+        {
+            const auto d = detail::iota_distance_(from_, to_);
+            const auto m = d % step_.value();
+            if(m == 0) {
+                return cursor{to_, step_};
+            }
+            return cursor{To(to_ - m + step_), step_};
+        }
+
         constexpr void check_bounds_(std::true_type)
         {
 
--- a/include/range/v3/view/iota.hpp
+++ b/include/range/v3/view/iota.hpp
@@ -163,6 +163,19 @@ namespace ranges
                                    -static_cast<iota_difference_t<Int>>(i0 - i1))
                              : static_cast<iota_difference_t<Int>>(i1 - i0);
         }
+
+        template<typename Int>
+        auto iota_abs_(Int i) -> CPP_ret(Int)( //
+            requires signed_integral<Int>)
+        {
+            return i < 0 ? -i : i;
+        }
+        template<typename Int>
+        auto iota_abs_(Int i) -> CPP_ret(Int)( //
+            requires unsigned_integral<Int>)
+        {
+            return i;
+        }
     } // namespace detail
     /// \endcond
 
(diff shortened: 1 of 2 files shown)


include/range/v3/view/iota.hpp
```diff
@@ -374,39 +443,54 @@ namespace ranges
...
+        {
+            return sentinel{to_};
+        }
+        CPP_member
+        auto CPP_fun(end_cursor)()(const requires(same_as<From, To>))
         {
-            return detail::if_then_t<same_as<From, To>, cursor, sentinel>{to_};
+            return cursor{to_, step_};
         }
+
```
johelegp:
```suggestion
```

include/range/v3/view/iota.hpp
```diff
@@ -374,39 +443,54 @@ namespace ranges
...
+              , step_(std::move(step))
             {}
         };
         cursor begin_cursor() const
         {
-            return cursor{from_};
+            return cursor{from_, step_};
         }
         CPP_member
         auto CPP_fun(end_cursor)()(const requires(same_as<To, unreachable_sentinel_t>))
```
johelegp:
```suggestion
        auto CPP_fun(end_cursor)()(const requires same_as<To, unreachable_sentinel_t>)
```

include/range/v3/view/iota.hpp
```diff
@@ -374,39 +443,54 @@ namespace ranges
...
         }
         CPP_member
-        auto CPP_fun(end_cursor)()(const requires(!same_as<To, unreachable_sentinel_t>))
+        auto CPP_fun(end_cursor)()(const requires(!same_as<To, unreachable_sentinel_t> &&
+                                                  !same_as<From, To>))
+        {
+            return sentinel{to_};
+        }
+        CPP_member
+        auto CPP_fun(end_cursor)()(const requires(same_as<From, To>))
```
johelegp:
```suggestion
        auto CPP_fun(end_cursor)()(const requires same_as<From, To>)
```

include/range/v3/view/iota.hpp
```diff
@@ -310,10 +315,47 @@ namespace ranges
                               ? finite
                               : unknown>
     {
+
+        struct __step
```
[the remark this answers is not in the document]
cor3ntin: True, this is not cmcstl2 :)

include/range/v3/view/iota.hpp
```diff
@@ -336,16 +378,35 @@ namespace ranges
...
+                requires detail::iota_stridable_<From, To>)
+            {
+                detail::iota_advance_(from_, step_.value());
+            }
+
+            CPP_member
+            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
+                requires detail::iota_stridable_<From, To>)
+            {
+                return step_ > 0 ? from_ + step_ >= that.to_ : from_ + step_ <= that.to_;
```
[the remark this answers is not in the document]
cor3ntin: Good point

test/view/iota.cpp
```diff
@@ -214,6 +214,15 @@ int main()
         CHECK((i - cstr2.begin()) == 4);
     }
 
+
+    {
+       // test step
+       ::check_equal(views::iota(0, 10, 2), {0, 2, 4, 6, 8});
+       ::check_equal(views::iota(5, 0, -1), {5, 4, 3, 2, 1});
+       ::check_equal(size(views::iota(1, 9, 3)), 3U);
```
BRevzin: If you change this to:

```cpp
::check_equal(views::iota(1, 9, 3), {1, 4, 7});
```

The test will fail, because `iota(1, 9, 3)` at the moment is an infinite stream of ints... despite having a correct fixed `size`.

johelegp: Seems like `cursor` is missing an `equal` overload that `requires detail::iota_stridable_<From, To>` equivalent to the `sentinel` overload.

cor3ntin: Hum indeed - I will have to fix that. Might take me a few days to get to it, I need to do something a bit more clever than in the sentinel case

cor3ntin: cursors don't necessary represent the end - I think what i need to do is introduce a `done_` member as is done for `closed_iota_view` - such that cursors compare equals if both are done or have the same `from_`

johelegp: It works in the general case, so you can compare any two cursors from the same `iota_view`. If you use any cursor in its range as the end, and iterate from the begin, you can be sure that it'll eventually reach that end. That makes me think, is there a test for reverse iterating `iota_view`s?

include/range/v3/view/iota.hpp
```diff
@@ -374,39 +457,64 @@ namespace ranges
...
         }
+        CPP_member
+            auto CPP_fun(end_cursor)()(const requires detail::iota_stridable_<From, To>)
+        {
+            const auto d = detail::iota_distance_(from_, to_);
+            const auto m = d % step_.value();
+            if(m == 0) {
+                return cursor{to_, step_};
+            }
+            return cursor{To(to_ - m + step_), step_};
```
johelegp: A `sentinel`-returning `end_cursor` overload that `requires detail::iota_stridable_<From, To>` is missing with this logic included. There also aren't test cases for the case when the iota view doesn't model `common_range`. Seems like `::check_equal(views::iota(1, short{9}, 3) | views::reverse, {7, 4, 1});` would fail.

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,66 @@ namespace ranges
...
             CPP_member
+            auto equal(cursor const & that) const -> CPP_ret(bool)( //
+            requires detail::iota_stridable_<From, To>)
+            {
+                return from_ == that.from_;
+            }
+            CPP_member
             auto prev() -> CPP_ret(void)( //
-                requires detail::decrementable_<From>)
+                requires(detail::decrementable_<From> &&
```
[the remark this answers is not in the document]
cor3ntin: Fixed

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,66 @@ namespace ranges
...
             CPP_member
+            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
+                requires detail::iota_stridable_<From, To>)
+            {
+                return detail::iota_distance_(from_, that.to_) >= detail::iota_abs_(step_.value());
+            }
+            CPP_member
             auto equal(cursor const & that) const -> CPP_ret(bool)( //
-                requires equality_comparable<From>)
+            requires (equality_comparable<From> &&!detail::iota_stridable_<From, To>))
```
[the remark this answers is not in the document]
cor3ntin: Indeed

include/range/v3/view/iota.hpp
cor3ntin: I have not thought of having a step but no upper bound - which is the reason I suppose you would need a sentinel overload.  I will think about it, might be an interesting use case, although not one I thought about or ever needed.

About your second point, it will not compile, and I am not sure whether it should

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,59 @@ namespace ranges
...
+            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
+                requires(!detail::iota_stridable_<From, To>))
             {
                 return from_ == that.to_;
             }
             CPP_member
+            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
+                requires detail::iota_stridable_<From, To>)
+            {
+                return detail::iota_distance_(from_, that.to_) >= detail::iota_abs_(step_.value());
```
BRevzin: This function looks wrong to me and I suspect it's never getting called from the tests. 

The case where `step == 1` should collapse into the case where there is no step, since that's the default. The no step sentinel is `from == to`. But this case is `to - from >= 1` or `from <= to - 1`, which is a different condition.

include/range/v3/view/iota.hpp
johelegp: This is not about `unreachable`, see https://godbolt.org/z/AwjtpF. The bounds aren't of the same type, so the view doesn't model `common_range`. If it doesn't compile in this branch, it must be due to the missing overload.

include/range/v3/view/iota.hpp
johelegp: Yes, the `sentinel` case is not being tested.

cor3ntin: It can never happen, I removed the function entirely

johelegp:
Why is it that both bounds have to be the same type? Why can't `views::iota(1, short{9}, 3)` or `views::iota(1, unreachable, 3)` work?

cor3ntin:
No particular technical reasons beside that I don't have a use case for it
and haven't done the work.
Might be worth doing.

johelegp:
Consider changing `iota_stridable_` to require `same_as<From, To>` and simplify to `totally_ordered<From>`. The current constraints are lax compared to what is actually allowed, so error messages might suffer.

johelegp:
A deduction guide's missing: https://godbolt.org/z/kit5xy.

include/range/v3/view/iota.hpp
```diff
@@ -310,10 +328,45 @@ namespace ranges
                               ? finite
                               : unknown>
     {
+
```
ericniebler, line 331:
```suggestion
```

include/range/v3/view/iota.hpp
```diff
@@ -310,10 +328,45 @@ namespace ranges
...
+                {
+                    return 1;
+                }
+            };
+            RANGES_NO_UNIQUE_ADDRESS
+            std::conditional_t<detail::iota_stridable_<From, To>,
+                               detail::iota_difference_t<From>, empty>
+                value_{1};
+        };
+
```
ericniebler, line 365: This is what `constant` and `mutable_` were for in `<range/v3/utility/box.hpp>`, but they are currently deprecated. Here's what I would suggest: Copy `mutable_` into a new (undeprecated) class `variable` that nukes the `mutable` on the data member and is `const`-correct. Then undeprecate `constant` and use those instead of inventing this here. (Please leave `mutable_` deprecated.)

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,53 @@ namespace ranges
...
             {
                 ++from_;
             }
-            bool equal(sentinel const & that) const
+            CPP_member
+            auto next() -> CPP_ret(void)( //
+                requires detail::iota_stridable_<From, To>)
+            {
+                detail::iota_advance_(from_, step_.value());
+            }
```
ericniebler, line 409: Might be cleaner as:

```c++
void next()
{
  if(RANGES_CONSTEXPR_IF(detail::iota_stridable_<From, To>))
    detail::iota_advance_(from_, step_.value());
  else
    ++from;
}
```

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,53 @@ namespace ranges
...
+            CPP_member
+            auto equal(sentinel const & that) const -> CPP_ret(bool)( //
+                requires(!detail::iota_stridable_<From, To>))
             {
                 return from_ == that.to_;
             }
             CPP_member
             auto equal(cursor const & that) const -> CPP_ret(bool)( //
-                requires equality_comparable<From>)
+            requires equality_comparable<From>)
```
ericniebler, line 418: Indenting.

```suggestion
                requires equality_comparable<From>)
```

include/range/v3/view/iota.hpp
```diff
@@ -336,36 +389,53 @@ namespace ranges
...
+            requires(detail::decrementable_<From> && !detail::iota_stridable_<From, To>))
             {
                 --from_;
             }
             CPP_member
+            auto prev() -> CPP_ret(void)( //
+                requires detail::iota_stridable_<From, To>)
+            {
+                detail::iota_advance_(from_, -step_.value());
+            }
```
ericniebler, line 433: Might be cleaner as:

```c++
CPP_member
auto prev() -> CPP_ret(void)(
    requires detail::decrementable_<From>)
{
    if(RANGES_CONSTEXPR_IF(detail::iota_stridable_<From, To>))
        detail::iota_advance_(from, -step_.value());
    else
        --from;
}
```

ericniebler:
I'm a little concerned in general that this is turning an increment of an `iota_view`'s iterator from a simple integer increment into a loop. I suspect there will be cases where this negatively effects runtime performance. `iota_view` is used heavily. I think I would prefer a version of this that leaves `iota_view` alone and adds a `strided_iota_view`, or even just bundles an `iota_view` with a `stride_view`.

ericniebler:
> @ericniebler Thanks for the review. I've not made progress on this because I'm questioning whether the feature is sufficiently justified too and whether it's worse pursuing.

I'm going to close this PR. I'm not opposed to it in principle, but the motivation is a little weak and the implementation has potential perf problems. I would welcome a different PR to address the same issue.

PR #1395, opened 2019-12-17
group_by() improvements (fixes #1393)

### LINKED ISSUE #1393: group_by becomes infinite loop if initial predicate yields false
From Conor Hoekstra on slack:

```cpp
std::vector v = {1, 2, 3};
auto g = v | group_by([](int, int){ return false; });
```

This should provide a range of ranges of size 1: `[[1], [2], [3]]`. But what it does instead is provide an infinite range of empty ranges. If the _first_ invocation of the predicate in a given subrange yields `false`, we never advance.

One possible fix in https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/group_by.hpp#L96-L105 is to do something like this:

```cpp
auto read() const
{
  auto second = ranges::next(cur_);
  return views::concat(
		  subrange{cur_, second},
		  subrange{second, last_} | views::take_while(pred{cur_, fun_}));
}
void next()
{
  cur_ = find_if_not(ranges::next(cur_), last_, pred{cur_, fun_});
}
```

That is, unconditionally include `cur_` in the "current range" without a predicate check. This gives the correct result, but I'm not sure if it's actually the "correct" fix or if there's a better way to implement something like this. 

`group_by` already requires `forward_range` so at least this wouldn't add any extra requirements.

BRevzin:
- Can handle initially-false predicate
- Only N-1 predicate invocations

### DIFF
--- a/include/range/v3/view/group_by.hpp
+++ b/include/range/v3/view/group_by.hpp
@@ -28,6 +28,7 @@
 #include <range/v3/range/access.hpp>
 #include <range/v3/range/concepts.hpp>
 #include <range/v3/range/traits.hpp>
+#include <range/v3/utility/optional.hpp>
 #include <range/v3/utility/semiregular_box.hpp>
 #include <range/v3/utility/static_const.hpp>
 #include <range/v3/view/facade.hpp>
@@ -53,26 +54,25 @@ namespace ranges
     private:
         friend range_access;
         Rng rng_;
+        // cached version of the end of the first subrange
+        detail::non_propagating_cache<iterator_t<Rng>> next_cur_;
         semiregular_box_t<Fun> fun_;
 
-        template<bool IsConst>
         struct cursor
         {
         private:
-            friend struct cursor<!IsConst>;
             friend range_access;
             friend group_by_view;
-            using CRng = meta::const_if_c<IsConst, Rng>;
-            iterator_t<CRng> cur_;
-            sentinel_t<CRng> last_;
-            semiregular_box_ref_or_val_t<Fun, IsConst> fun_;
-            iterator_t<CRng> local_last_;
+            iterator_t<Rng> cur_;
+            iterator_t<Rng> next_cur_;
+            sentinel_t<Rng> last_;
+            semiregular_box_ref_or_val_t<Fun, false> fun_;
 
             struct mixin : basic_mixin<cursor>
             {
                 mixin() = default;
                 using basic_mixin<cursor>::basic_mixin;
-                iterator_t<CRng> base() const
+                iterator_t<Rng> base() const
                 {
                     return this->get().cur_;
                 }
@@ -80,32 +80,23 @@ namespace ranges
 
             struct pred
             {
-                iterator_t<CRng> first_;
-                semiregular_box_ref_or_val_t<Fun, IsConst> fun_;
-                bool operator()(range_reference_t<CRng> r) const
+                iterator_t<Rng> first_;
+                semiregular_box_ref_or_val_t<Fun, false> fun_;
+                bool operator()(range_reference_t<Rng> r) const
                 {
                     return invoke(fun_, *first_, r);
                 }
             };
-#ifdef RANGES_WORKAROUND_MSVC_787074
-            template<bool Const = IsConst>
-            auto read() const -> subrange<iterator_t<meta::const_if_c<Const, Rng>>>
-#else  // ^^^ workaround / no workaround vvv
-            auto read() const -> subrange<iterator_t<CRng>>
-#endif // RANGES_WORKAROUND_MSVC_787074
+            auto read() const -> subrange<iterator_t<Rng>>
             {
-                return {cur_, local_last_};
+                return {cur_, next_cur_};
             }
             void next()
             {
-                cur_ = local_last_;
-                local_last_ = next_local();
-            }
-            iterator_t<CRng> next_local()
-            {
-                return cur_ != last_
-                           ? find_if_not(ranges::next(cur_), last_, pred{cur_, fun_})
-                           : cur_;
+                cur_ = next_cur_;
+                next_cur_ = cur_ != last_
+                                ? find_if_not(ranges::next(cur_), last_, pred{cur_, fun_})
+                                : cur_;
             }
 
             bool equal(default_sentinel_t) const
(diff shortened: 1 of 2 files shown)


BRevzin:
This is my fix for #1393 / #1376 . It actually does work in the constructor, which is probably not allowed - but at least it makes it easy for me to understand the invariants (the next thing is always `[cur, local)`)

johelegp:
The standard places no explicit requirements on constructors that are not special member functions. I don't remember, if I ever knew, why the views (in `std` and `::range`) don't copy cached iterators (other than it not being a salient property), but if your solution were to do as a Roman in Rome, then it'd wrap `local_last_` in [`non_propagation_cache`](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/utility/optional.hpp#L908).

BRevzin:
@johelegp The only ways I can come up with for this are:

1. Have a `mutable non_propagation_cache<iterator_t<Rng>>` member for the first `find_not_if` result. This gets set on `begin()` if it's not set yet and passed through to the cursor, which then has to flip how it uses this.
2. Same as (1) except have it non-`mutable`, and then disallow you from `const`-iterating (much like `filter`)
3. Flip the invariant in the cursor so that `read()` and `next()` both potentially do the `find_not_if` search - but try to ensure that it's only done one time per iteration, to avoid having to do it in `begin()`. 

(3) would basically have the `cursor` look like:

```cpp
iterator_t<CRng> cur_;
non_propagating_cache<iterator_t<CRng>> next_cur_;
sentinel_t<CRng> last_;

next() {
    cur_ = next_cur_ ? *next_cur_ : next_local();
    next_cur_ = next_local();
}

read() {
    if (!next_cur_) next_cur_ = next_local();
    return {cur_, *next_cur_};
}
```

I haven't fully thought through if (3) works, but assuming it does, I think it's better than (1) or (2). But is it better than (0) (i.e. the current PR)?

johelegp:
As I mentioned in https://github.com/ericniebler/range-v3/pull/1376#discussion_r356905264, (2) should be done instead of (1). As for (3), I'd have to think about it. If some other views do it, it might work.

BRevzin:
Implemented (2).

include/range/v3/view/group_by.hpp
```diff
@@ -53,59 +54,49 @@ namespace ranges
     private:
         friend range_access;
         Rng rng_;
+        // cached version of the end of the first subrange
```
johelegp: You probably wouldn't need the comment if `next_cur_` was named like the equivalent of the other views that cache `begin()`.

BRevzin: ... which would be how? Some of them name it `begin_` (which is wrong for this case) and some of them name it `cache_` (which is meaningless).

johelegp: `begin_` looks right. It _is_ the view's begin.

johelegp: Since `group_by` always includes the first element, `second_` seems like a good name.

These PRs were brought up due to the lack of test coverage. I hope to not have made you choose a solution riddled with more bugs. Fingers crossed that there's an empty range test to make sure `second_`'s not misused.

BRevzin: It could be if I made it like:

```cpp
non_propagating_cache<subrange<iterator_t<Rng>>> begin_;
```

That's larger (if we care) and a bit redundant to store `rng_.begin()` twice. But if the cursor has a `subrange<iterator_t<Rng>>` member, that might be clearer (regardless of any other changes).

BRevzin: > Fingers crossed that there's an empty range test to make sure second_'s not misused.

Uh... of... uh... Why yes of course there will be an empty range test.

johelegp: Make that tests, for each possible return value of the predicate. You never know.

johelegp: Not to make it sound as it being your fault. I was hoping that they were there since the beginning, and not just one test with an obvious use case. Your answer suggests they weren't ;(

vgasparyan1995: Regarding the tests, you could also make use of my [test cases](https://github.com/ericniebler/range-v3/pull/1376/files#diff-d4e7db63ae69a359b671e7c5919644fe), which ensures [the implementation meets the definition](https://github.com/ericniebler/range-v3/pull/1376#issue-346601466). I've tested your changes with them, they pass.

BRevzin: Adding your tests, and an empty and singleton test.

include/range/v3/view/group_by.hpp
```diff
@@ -53,56 +54,52 @@ namespace ranges
...
         private:
-            friend struct cursor<!IsConst>;
             friend range_access;
             friend group_by_view;
-            using CRng = meta::const_if_c<IsConst, Rng>;
-            iterator_t<CRng> cur_;
-            sentinel_t<CRng> last_;
-            semiregular_box_ref_or_val_t<Fun, IsConst> fun_;
+            iterator_t<Rng> cur_;
+            iterator_t<Rng> second_;
```
johelegp: This one really is `next_cur_`, while the view's one is `second_`. For a second, while looking at `next()`, I thought it was doing exactly the same as `views::cache1`.

include/range/v3/view/group_by.hpp
```diff
@@ -53,56 +54,52 @@ namespace ranges
     private:
         friend range_access;
         Rng rng_;
+        // cached version of the end of the first subrange / start of the second subrange
```
johelegp: This comment is wrong. `second_` may be any of those, given the right inputs, but not necessarily.

BRevzin: I do not think the comment is wrong.

johelegp: In the other cases, it is the second element of the first subrange, isn't it?

include/range/v3/view/group_by.hpp
```diff
@@ -109,9 +109,9 @@ namespace ranges
                 return cur_ == that.cur_;
             }
             cursor(semiregular_box_ref_or_val_t<Fun, false> fun, iterator_t<Rng> first,
-                   iterator_t<Rng> next_first, sentinel_t<Rng> last)
+                   iterator_t<Rng> next_cur, sentinel_t<Rng> last)
```
johelegp: Here it is `second`.

include/range/v3/view/group_by.hpp
BRevzin: No. It's always the sentinel of the first subrange. Which would be the beginning of the second subrange, if there is one.

johelegp: How can that be? I'm seeing that the logic in `next()` and `begin_cursor()` is the same.

BRevzin: Consider the range `[1, 2, 2, 3, 3, 4, 5]` and the predicate `==`. `begin_cursor()`
will look like:

```
[1, 2, 2, 3, 3, 4, 5]
 ↑  ↑               ↑
 |  next_cur_       last_
 cur_
```

and on the next iteration will look like:

```
[1, 2, 2, 3, 3, 4, 5]
    ↑     ↑         ↑
    |   next_cur_   last_
   cur_                 
```

The `view`'s `second_` member is the initial value of the `cursor`'s `next_cur_` member.

BRevzin: Yeah, the logic is the same - we're advancing by a whole subrange (not a single element in it).

PR #1397, opened 2019-12-19
Make safe_iterator_t standard conformant

johelegp:
Missed by #1370. A more complete PR would excise `forwarding_range_`.

### DIFF
--- a/include/range/v3/range/dangling.hpp
+++ b/include/range/v3/range/dangling.hpp
@@ -20,6 +20,7 @@
 
 #include <range/v3/range_fwd.hpp>
 
+#include <range/v3/range/access.hpp>
 #include <range/v3/range/concepts.hpp>
 #include <range/v3/utility/static_const.hpp>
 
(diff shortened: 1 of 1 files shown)


johelegp:
After this PR, given `std::vector v;`, `ranges::find(v | views::drop(0), 42)` should return `ranges::dangling`.

CaseyCarter:
> After this PR, given `std::vector v;`, `ranges::find(v | views::drop(0), 42)` should return `ranges::dangling`.

`std::vector<int>&` models `safe_range`; `std::vector<int>` and `std::vector<int>&&` do not.

johelegp:
But prvalues of specializations of `drop_view` never do.

CaseyCarter:
> But prvalues of specializations of `drop_view` never do.

🤦‍♂ Nevermind: I missed the point of the PR, which is that `forwarding_range_` is broken after implementing the P1870 changes to the range access CPOs.

PR #1398, opened 2019-12-19
Finish implementation of P1870R1

CaseyCarter:
* fold `range_impl_` concept back into `range`
* Replace `forwarding_range_` concept with `safe_range`

Drive-by:
* Remove unused `#define CPP_noop`
* Use `view<uncvref_t<T>>` instead of `view<decay_t<T>>` in the definition of `viewable_range` since pointers cannot model `view`. (Per the editorialish change I submitted in https://github.com/cplusplus/draft/pull/3563).

### DIFF
--- a/include/range/v3/range/concepts.hpp
+++ b/include/range/v3/range/concepts.hpp
@@ -76,20 +76,16 @@ namespace ranges
 
     // clang-format off
     template<typename T>
-    CPP_concept_bool range_impl_ =
-        CPP_requires ((T &&) t) //
+    CPP_concept_bool range =
+        CPP_requires ((T &) t) //
         (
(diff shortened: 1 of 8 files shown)


CaseyCarter:
Note that there's no deprecated `forwarding_range_` here, I simply ripped it out.

BRevzin:
@CaseyCarter  Were you planning on writing a paper to rename the `safe_*` things to `borrowed_*`? Or are we just going to stick with `safe_*` for 20?

CaseyCarter:
> Were you planning on writing a paper to rename the `safe_*` things to `borrowed_*`?

I'm not planning such a paper. Maybe @cjdb is? If anything, I think such a change would be easier to handle via the issue resolution process than with a paper.

ericniebler:
Thanks

PR #1405, opened 2020-01-03
Making more range adapters safe (Fixes ericniebler/stl2#640).

BRevzin:
This will also be the content of [P2017R0](https://brevzin.github.io/cpp_proposals/2017_safe_range/p2017r0.html).

I may have missed some, but I'm pretty sure these are all safe.

### DIFF
--- a/include/range/v3/utility/semiregular_box.hpp
+++ b/include/range/v3/utility/semiregular_box.hpp
@@ -293,13 +293,16 @@ namespace ranges
         reference_wrapper<
             meta::if_c<IsConst, semiregular_box<T> const, semiregular_box<T>>>>;
 
-    // for those range adapters that store a function by way of semiregular_box_ref_or_val_t
-    // if that alias is actually a T (that is, it owns the callable), then the range can
-    // be a safe_range since the adapter will not depend on the outer range surviving
-    // this is a necessary, but not sufficient, condition
-    template <typename T>
-    RANGES_INLINE_VAR constexpr bool is_safe_fun_ =
-        defer::same_as<T, semiregular_box_ref_or_val_t<T>>;
+    namespace detail {
+        // For those range adapters that store a function by way of
+        // semiregular_box_ref_or_val_t if that alias is actually a T (that is,
+        // it owns the callable), then the range can be a safe_range since the
+        // adapter will not depend on the outer range surviving.
+        // This is a necessary, but not sufficient, condition
+        template <typename T>
+        RANGES_INLINE_VAR constexpr bool is_safe_fun =
+            defer::same_as<T, semiregular_box_ref_or_val_t<T>>;
+    }
     /// @}
 
     /// \cond
--- a/include/range/v3/utility/semiregular_box.hpp
+++ b/include/range/v3/utility/semiregular_box.hpp
@@ -292,17 +292,6 @@ namespace ranges
         meta::if_c<IsConst || std::is_empty<T>::value, T, reference_wrapper<T>>,
         reference_wrapper<
             meta::if_c<IsConst, semiregular_box<T> const, semiregular_box<T>>>>;
-
-    namespace detail {
-        // For those range adapters that store a function by way of
-        // semiregular_box_ref_or_val_t if that alias is actually a T (that is,
-        // it owns the callable), then the range can be a safe_range since the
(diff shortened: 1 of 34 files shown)


johelegp:
I'm looking forward to reviewing this tomorrow.

johelegp:
On the paper: https://www.diffchecker.com/DsiXrtKr.

BRevzin:
I agree with all of your assessments (some of them took me a while to understand). I made all of the above enhancements, excepting `linear_distribute_view`.

include/range/v3/utility/semiregular_box.hpp
```diff
@@ -292,6 +292,14 @@ namespace ranges
         meta::if_c<IsConst || std::is_empty<T>::value, T, reference_wrapper<T>>,
         reference_wrapper<
             meta::if_c<IsConst, semiregular_box<T> const, semiregular_box<T>>>>;
+
+    // for those range adapters that store a function by way of semiregular_box_ref_or_val_t
+    // if that alias is actually a T (that is, it owns the callable), then the range can
+    // be a safe_range since the adapter will not depend on the outer range surviving
+    // this is a necessary, but not sufficient, condition
+    template <typename T>
+    RANGES_INLINE_VAR constexpr bool is_safe_fun_ =
```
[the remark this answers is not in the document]
BRevzin: Done.

BRevzin:
One of the questions this surfaces (and I noted this [in the paper](https://brevzin.github.io/cpp_proposals/2017_safe_range/d2017r0.html#implementation-strategy)) is what do we actually want to put in the working draft. We can (1) mandate a particular implementation strategy (like range-v3's), (2) allow implementation divergence on what makes a safe range, or (3) just only mark ranges as safe if all reasonable implementations would be. 

I think the long-term correct choice is option 4: aggressively pursue `-Wlifetime` and get to a point where compilers can very reliably detect dangling on their own, in which case we can just silently deprecate `safe_range`.

marehr:
gcc10 will have a `#include <span>` header in `-std=2a` mode, so it would be awesome if `std::span` would also be defined as a safe range.

h-2:
> One of the questions this surfaces (and I noted this in the paper) is what do we actually want to put in the working draft especially for ranges that might have different implementation strategies.

Hey Barry, yes, I thought of this a while ago, as well. Concepts in general have the property that non satisfying a concept can have very specific effects on API and ABI. If it is implementation-defined for some types whether they satisfy a concept, this can get very complicated for users. Implementers may also not be able to change from one implementation-defined behaviour to another because it would change the behaviour of other parts of STD in an incompatible way for their users.

> I think the long-term correct choice is option 4: aggressively pursue -Wlifetime and get to a point where compilers can very reliably detect dangling on their own, in which case we can just silently deprecate safe_range.

safe_range has other implications than memory safety alone, e.g. for "reconstructible ranges".

include/range/v3/utility/semiregular_box.hpp
```diff
@@ -292,6 +292,17 @@ namespace ranges
...
+    namespace detail {
+        // For those range adapters that store a function by way of
+        // semiregular_box_ref_or_val_t if that alias is actually a T (that is,
+        // it owns the callable), then the range can be a safe_range since the
+        // adapter will not depend on the outer range surviving.
+        // This is a necessary, but not sufficient, condition
+        template <typename T>
+        RANGES_INLINE_VAR constexpr bool is_safe_fun =
+            defer::same_as<T, semiregular_box_ref_or_val_t<T>>;
+    }
```
ericniebler: I would prefer not to consider these safe ranges at all. I want the freedom to move the functions into the view objects when compiler optimizers get good enough to eliminate the abstraction cost.

BRevzin: I rolled all of these back - no more `is_safe_fun` or any conditions that depend on them.

However, I kept three views as conditionally safe (er, I guess borrowed now) implicitly based  n this: `keys`, `views`, and `zip`. On the basis that these are commonly used views, their function objects are empty so I know range-v3 stores them in the iterators, and it's useful to have more ranges be safe in this regard.

ericniebler: Thanks, Barry.

PR #1408, opened 2020-01-03
Remove unused header

johelegp:


ericniebler:
Looks like test/view/cycle.cpp was relying on the transitive include. Could you fix the test, pls?

ericniebler:
Looks like test/view/cycle.cpp was relying on the transitive include. Could you fix the test, pls?

PR #1409, opened 2020-01-07
Explicitly opt drop_while_view out of sized_range

johelegp:
I've requested a DR against the CD saying:
> Because it has to compute its `begin()`, it can never model `sized_range` due to not meeting its non-amortized O(1) requirement.


johelegp:
Note that they'll be back to being sized since `sized_range`'s requirement was relaxed to amortized O(1).

Issue #1411, opened 2020-01-08
`find_package(range-v3)` with CMake 3.10 does not work (range-v3 0.10.0)


Marc--Olivier:
## Description

When I run CMake 3.10 on a project that includes range-v3 0.10.0,
I get the following error message:

```
Cannot specify link libraries for target "range-v3::meta" which is not built by this project.
```

It took me quite some time to realize that CMake 3.10 does not seem able to
process [cmake/range-v3-config.cmake](https://github.com/ericniebler/range-v3/blob/0.10.0/cmake/range-v3-config.cmake). At some point, I tried out with CMake 3.16.2 and it worked. It would be good to update range-v3-config.cmake to fail if the CMake version is not greater than, for example, 3.16 (I did not take the time to figure out the smallest CMake version that works).

Note that we could use a check similar to the check contained in the generated file `lib/cmake/range-v3/range-v3-targets.cmake` when installing the library:

```
if("${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}" LESS 2.5)
   message(FATAL_ERROR "CMake >= 2.6.0 required")
endif()
```

## Steps to reproduce:

1. Compile range-v3 and install it:
```sh
git clone https://github.com/ericniebler/range-v3 --branch=0.10.0 --depth=1
mkdir install
mkdir build && pushd build
cmake ../range-v3 -G Ninja -DCMAKE_BUILD_TYPE=Release -DRANGE_V3_TESTS=OFF -DCMAKE_INSTALL_PREFIX="../install"
ninja
ninja install
popd
```

2. Generate a CMake project:
```sh
mkdir test_project && pushd test_project
cat >CMakeLists.txt <<EOL
cmake_minimum_required(VERSION 3.10)
find_package(range-v3 HINTS "../install")
EOL
cmake .
```

=> I get the following error:

```sh
CMake Error at /tmp/tmp.V7kykEykmj/install/lib/cmake/range-v3/range-v3-config.cmake:5 (target_link_libraries):
  Cannot specify link libraries for target "range-v3::meta" which is not
  built by this project.
Call Stack (most recent call first):
  CMakeLists.txt:2 (find_package)


-- Configuring incomplete, errors occurred!
See also "/tmp/tmp.V7kykEykmj/test_project/CMakeFiles/CMakeOutput.log".
```

marehr:
[...]

PR #1412, opened 2020-01-08
Support cmake version less than 3.11 when installing targets (#1411)

marehr:
> I didn't consider the minimum required version when making the change. The [patch notes for 3.11](https://cmake.org/cmake/help/latest/release/3.11.html#commands) says that `target_link_libraries()` was added then. @marehr, could you open a PR, please?

See #1411


ericniebler:
Thanks.

PR #1449, opened 2020-03-16
Get the range library compiling with msvc 19.24.28319 with /std:c++17

ericniebler:


include/range/v3/view/adaptor.hpp
```diff
@@ -251,7 +251,16 @@ namespace ranges
...
+            #ifndef _MSC_VER
             using basic_mixin<adaptor_cursor>::basic_mixin;
+            #else
+            explicit constexpr basic_adaptor_mixin(adaptor_cursor && cur)
+              : basic_mixin<adaptor_cursor>(static_cast<adaptor_cursor &&>(cur))
+            {}
+            explicit constexpr basic_adaptor_mixin(adaptor_cursor const & cur)
+              : basic_mixin<adaptor_cursor>(cur)
+            {}
+            #endif
```
ericniebler: @CaseyCarter This is an example of where I needed to avoid inheriting constructors or else msvc would ICE in c++17-mode.

include/range/v3/view/cartesian_product.hpp
```diff
@@ -435,11 +439,38 @@ namespace ranges
...
+                    requires forward_range<Rng0> && viewable_range<Rng0> &&            //
+                             forward_range<Rng1> && viewable_range<Rng1> &&
+                             forward_range<Rng2> && viewable_range<Rng2>)
+            {
+                return cartesian_product_view<all_t<Rng0>, all_t<Rng1>, all_t<Rng2>>{
+                    all(static_cast<Rng0 &&>(rng0)), //
+                    all(static_cast<Rng1 &&>(rng1)), //
+                    all(static_cast<Rng2 &&>(rng2))};
+            }
+#endif
```
ericniebler: And this is an example of where I had to avoid variadic arguments in `operator()`.

PR #1452, opened 2020-03-18
Remove the need to use msvc's experimental preprocessor.

ericniebler:


CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ add_library(range-v3-concepts INTERFACE)
 add_library(range-v3::concepts ALIAS range-v3-concepts)
 target_include_directories(range-v3-concepts INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/>)
 target_include_directories(range-v3-concepts SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
-target_compile_options(range-v3-concepts INTERFACE $<$<CXX_COMPILER_ID:MSVC>:/permissive- /experimental:preprocessor /wd5105>)
+target_compile_options(range-v3-concepts INTERFACE $<$<CXX_COMPILER_ID:MSVC>:/permissive- /wd5105>)
```
CaseyCarter: `/wd5105` was needed to suppress a warning from the experimental preprocessor about some nastiness in the platform headers. (Waiting for the last few fixes to the Windows SDK is why the new preprocessor _still_ isn't the default.)
```suggestion
target_compile_options(range-v3-concepts INTERFACE $<$<CXX_COMPILER_ID:MSVC>:/permissive->)
```

PR #1455, opened 2020-03-19
Remove the enable_view heuristic

ericniebler:
This is to more closely harmonize range-v3 with C++20's `std::ranges`. Refers to cplusplus/nbballot#278

### DIFF
--- a/include/range/v3/range/concepts.hpp
+++ b/include/range/v3/range/concepts.hpp
@@ -34,33 +34,23 @@
 #ifndef RANGES_NO_STD_FORWARD_DECLARATIONS
 // Non-portable forward declarations of standard containers
 RANGES_BEGIN_NAMESPACE_STD
-    RANGES_BEGIN_NAMESPACE_CONTAINER
-        template<typename Key,
-                 typename Compare /*= less<Key>*/,
-                 typename Alloc /*= allocator<Key>*/>
-        class set;
+    RANGES_BEGIN_NAMESPACE_VERSION
+        template<typename ElementType, size_t Extent>
+        class span;
 
-        template<typename Key,
-                 typename Compare /*= less<Key>*/,
-                 typename Alloc /*= allocator<Key>*/>
-        class multiset;
-
-        template<typename Key,
-                 typename Hash /*= hash<Key>*/,
-                 typename Pred /*= equal_to<Key>*/,
-                 typename Alloc /*= allocator<Key>*/>
-        class unordered_set;
-
-        template<typename Key,
-                 typename Hash /*= hash<Key>*/,
-                 typename Pred /*= equal_to<Key>*/,
-                 typename Alloc /*= allocator<Key>*/>
-        class unordered_multiset;
-    RANGES_END_NAMESPACE_CONTAINER
+        template<typename CharT, typename Traits>
+        class basic_string_view;
+    RANGES_END_NAMESPACE_VERSION
 RANGES_END_NAMESPACE_STD
 #else
-#include <set>
-#include <unordered_set>
+#ifdef __has_include
+#if __has_include(<span>)
+#include <span>
+#endif
+#if __has_include(<string_view>)
+#include <string_view>
+#endif
+#endif
 #endif
 
 #include <range/v3/detail/disable_warnings.hpp>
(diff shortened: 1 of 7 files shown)


include/range/v3/range/concepts.hpp
```diff
@@ -34,33 +34,23 @@
 #ifndef RANGES_NO_STD_FORWARD_DECLARATIONS
 // Non-portable forward declarations of standard containers
```
CaseyCarter:
```suggestion
// Non-portable forward declarations of standard library components
```

include/range/v3/range/concepts.hpp
```diff
@@ -188,88 +178,28 @@ namespace ranges
...
+
+#if defined(__cpp_lib_string_view) && __cpp_lib_string_view > 0
+    template<typename Char, typename Traits>
+    RANGES_INLINE_VAR constexpr bool enable_view<std::basic_string_view<Char, Traits>> =
+        true;
+#endif
+
+#if defined(__cpp_lib_span) && __cpp_lib_span > 0
+    template<typename T, std::size_t N>
+    RANGES_INLINE_VAR constexpr bool enable_view<std::span<T, N>> = true;
```
johelegp:
```suggestion
    RANGES_INLINE_VAR constexpr bool enable_view<std::span<T, N>> = N == 0 || N == std::dynamic_extent;
```

include/range/v3/range/concepts.hpp
```diff
@@ -188,88 +178,28 @@ namespace ranges
...
+          : std::is_base_of<view_base, T>
+        {};
     } // namespace detail
     /// \endcond
 
     // Specialize this if the default is wrong.
     template<typename T>
     RANGES_INLINE_VAR constexpr bool enable_view =
-        detail::enable_view_impl_(detail::nullptr_<T>(0));
+        detail::enable_view_<T>::value;
```
CaseyCarter: Why the added indirection instead of using `std::is_base_of<view_base, T>::value` directly here?

include/range/v3/range/concepts.hpp
```diff
@@ -188,88 +178,28 @@ namespace ranges
...
-        {
-            return nullptr;
-        }
-        template<typename T>
-        constexpr int nullptr_(long)
-        {
-            return 0;
-        }
+        struct enable_view_
+          : std::is_base_of<view_base, T>
```
CaseyCarter: Why `is_base_of` instead of `derived_from`?

johelegp:
`ranges::span<T, N>` should similarly opt-out of being a `view` when `N` isn't `dynamic_extent`. It opts-in by inheriting from `view_interface`, which has `view_base` as a base.

include/range/v3/range/concepts.hpp
CaseyCarter: I really should have worded this as `N + 1 < 2` and told the unsigned type proponents that they asked for it.

(FWIW: the value of `enable_view` doesn't really matter for the specializations of `span` that don't satisfy `semiregular`, I complicated the wording to respect the letter of the wording that says users may specialize `enable_view` to `true` for "types that model `view`".)

CaseyCarter:
> `ranges::span<T, N>` should similarly opt-out of being a `view` when `N` isn't `dynamic_extent`. It opts-in by inheriting from `view_interface`, which has `view_base` as a base.

FWIW, `ranges::span` is currently (non-conformingly) always default constructible and therefore always a `view`. We should probably have a separate issue to make it conform and opt out of `view` at the same time.

include/range/v3/range/concepts.hpp
ericniebler: So that folks who need to customize this trait without dragging in all of the range library can do so. I have it in `detail::` because I am a coward.

include/range/v3/range/concepts.hpp
ericniebler: I am not opposed to folks privately inheriting from `view_base`. That's good enough for me.

include/range/v3/range/concepts.hpp
ericniebler: > I really should have worded this as `N + 1 < 2`

I'm going with that formulation because it doesn't require a declaration of `std::dynamic_extent`.

ericniebler:
@CaseyCarter, @johelegp thanks for the feedback. Very valuable.

PR #1465, opened 2020-04-11
Fix find_package in parent and subdirectory

Arghnews:
Given a directory structure:
- parent
  - CMakeLists.txt
  - child
    - CMakeLists.txt

Calling `find_package(range-v3)` from within `parent/CMakeLists.txt` and `child/CMakeLists.txt` results in an error.

`parent/CMakeLists.txt` contains:
```cmake
cmake_minimum_required(VERSION 3.6)
project(parent)
find_package(range-v3 REQUIRED)
add_subdirectory(child)
```
`child/CMakeLists.txt` contains:
```cmake
cmake_minimum_required(VERSION 3.6)
project(child)
find_package(range-v3 REQUIRED)
```

From parent, running in bash:
`(rm -rf build && mkdir build && cd build && cmake ..)`

Results in this CMake output:

> (Generic cmake build output omitted)
> CMake Error at /usr/local/lib/cmake/range-v3/range-v3-config.cmake:3 (add_library):
>   add_library cannot create imported target "range-v3::meta" because another
>   target with the same name already exists.
> Call Stack (most recent call first):
>   child/CMakeLists.txt:3 (find_package)
> 
> CMake Error at /usr/local/lib/cmake/range-v3/range-v3-config.cmake:4 (add_library):
>   add_library cannot create imported target "range-v3::concepts" because
>   another target with the same name already exists.
> Call Stack (most recent call first):
>   child/CMakeLists.txt:3 (find_package)
> 
> CMake Error at /usr/local/lib/cmake/range-v3/range-v3-config.cmake:5 (add_library):
>   add_library cannot create imported target "range-v3::range-v3" because
>   another target with the same name already exists.
> Call Stack (most recent call first):
>   child/CMakeLists.txt:3 (find_package)
> 
> CMake Error at /usr/local/lib/cmake/range-v3/range-v3-config.cmake:14 (target_link_libraries):
>   Cannot specify link libraries for target "range-v3::meta" which is not
>   built by this project.
> Call Stack (most recent call first):
>   child/CMakeLists.txt:3 (find_package)
> 

Test system used:
Ubuntu 18.10 x64, cmake 3.16.4, range-v3 master, gcc8 or clang10

This is due to the cmake targets `range-v3::concepts`, `range-v3::meta` and `range-v3::range-v3` being defined in two places. These are all alias targets of the INTERFACE targets `range-v3-concepts`, `range-v3-meta` and `range-v3` respectively.
One is in [cmake/range-v3-config.cmake](https://github.com/ericniebler/range-v3/blob/master/cmake/range-v3-config.cmake).
The other is in the generated file (/usr/local/lib/cmake/range-v3/)`range-v3-targets.cmake`. This file is generated or "installed" by [lines 170/171 in CMakeLists.txt](https://github.com/ericniebler/range-v3/blob/master/CMakeLists.txt#L170-L171)

This change simply avoids defining each target in `cmake/range-v3-config.cmake` if the target is already defined, by wrapping each define in an `if (NOT TARGET ${alias_target})` style if.

I've done a quick test with cmake 3.6, 3.10.3 and 3.16 and they all allow me to build and install range-v3 and build my parent/child example.

If range-v3 only supports cmake >=3.10 then [include_guard()](https://cmake.org/cmake/help/v3.10/command/include_guard.html) can be used instead, placing it after line 1 in `cmake/range-v3-config.cmake`. This is mentioned in #1343.

I've read through the pandora's box of cmake related issues with range-v3: #1343, #1360, #1385, #1412.
However, I am definitely not a cmake wizard and could be missing something.
To the best of my current - and mediocre at best - cmake knowledge, this doesn't break anything and fixes this multiple inclusion problem.

### DIFF
--- a/include/range/v3/range/access.hpp
+++ b/include/range/v3/range/access.hpp
@@ -20,25 +20,6 @@
 #include <limits>
 #include <utility>
 
-#include <range/v3/range_fwd.hpp>
-
-#include <range/v3/iterator/concepts.hpp>
-#include <range/v3/iterator/reverse_iterator.hpp>
-#include <range/v3/iterator/traits.hpp>
-#include <range/v3/utility/static_const.hpp>
-
-#ifndef RANGES_NO_STD_FORWARD_DECLARATIONS
-// Non-portable forward declarations of standard library components
-RANGES_BEGIN_NAMESPACE_STD
-    RANGES_BEGIN_NAMESPACE_VERSION
-        template<typename ElementType, size_t Extent>
-        class span;
-
-        template<typename CharT, typename Traits>
-        class basic_string_view;
-    RANGES_END_NAMESPACE_VERSION
-RANGES_END_NAMESPACE_STD
-#else
 #ifdef __has_include
 #if __has_include(<span>)
 #include <span>
@@ -47,7 +28,13 @@ RANGES_END_NAMESPACE_STD
 #include <string_view>
 #endif
 #endif
-#endif
+
+#include <range/v3/range_fwd.hpp>
+
+#include <range/v3/iterator/concepts.hpp>
+#include <range/v3/iterator/reverse_iterator.hpp>
+#include <range/v3/iterator/traits.hpp>
+#include <range/v3/utility/static_const.hpp>
 
 #include <range/v3/detail/disable_warnings.hpp>
 
--- a/include/range/v3/range/concepts.hpp
+++ b/include/range/v3/range/concepts.hpp
@@ -18,6 +18,15 @@
 #include <type_traits>
 #include <utility>
 
+#ifdef __has_include
+#if __has_include(<span>)
+#include <span>
+#endif
+#if __has_include(<string_view>)
+#include <string_view>
+#endif
+#endif
+
 #include <meta/meta.hpp>
 
 #include <concepts/concepts.hpp>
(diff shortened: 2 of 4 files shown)


cmake/range-v3-config.cmake
```diff
@@ -1,17 +1,18 @@
 include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets.cmake")
```
johelegp, line 2:
```suggestion

if(TARGET range-v3::meta)
  return()
endif()
```
should suffice.

johelegp:
I overlooked this scenario.

Arghnews:
That is simpler and seems to work just as well, cheers.

johelegp:
You'll need to manually remove the other changes, as my suggestion was applied on top of those.

Arghnews:
Ah I see yes it's made a minor mess (shame, that little automagical "apply changes" button promised so much). I apologise for my ignorance with this. I was going to make another commit with the changes you proposed, squash the last 3 into that and then push this, is that a good idea? If not, how else could I achieve this?

johelegp:
No worries. It'd be simpler to `git reset --hard master`, apply my suggestion, and then `git push -f`.

Arghnews:
Oops, did I muck something up? It appear to have added the last commit from range-v3 master in too or something.

johelegp:
It should be alright. I shouldn't have abused the suggestion feature.

ericniebler:
Ooops. I tried to fix the history in this by force-pushing, but for some reason github closed the PR, and I can't reopen it. Please create a new clean PR.

ericniebler:
Nevermind, I created a new PR for you here: #1466

PR #1472, opened 2020-04-21
Range std compat

### LINKED ISSUE #1470: gcc10: range-v3 views are not combinable with std::views
https://godbolt.org/z/6_BQS3

```c++
#include <ranges>
#include <string>
#include <range/v3/view/take.hpp>

int main()
{
    std::string str{"foo"};

    auto v3 = str | std::views::take(5) | ranges::view::take(1);
    return 0;
}
```

with `g++10 --std=c++2a`

will result in 

```c++
<source>: In function 'int main()':
<source>:9:63: error: use of deleted function 'constexpr concepts::return_t<Rng, typename std::enable_if<((range<Rng> && (! viewable_range<Rng>)) && concepts::detail::CPP_true(concepts::detail::Nil{})), void>::type> ranges::views::view_closure_base_ns::operator|(Rng&&, const ranges::views::view_closure<ViewFn>&) [with Rng = std::ranges::take_view<std::ranges::ref_view<std::__cxx11::basic_string<char> > >; ViewFn = ranges::detail::bind_back_fn_<ranges::views::take_base_fn, int>; concepts::return_t<Rng, typename std::enable_if<((range<Rng> && (! viewable_range<Rng>)) && concepts::detail::CPP_true(concepts::detail::Nil{})), void>::type> = std::ranges::take_view<std::ranges::ref_view<std::__cxx11::basic_string<char> > >; typename std::enable_if<((range<Rng> && (! viewable_range<Rng>)) && concepts::detail::CPP_true(concepts::detail::Nil{})), void>::type = void]'
    9 |     auto v3 = str | std::views::take(5) | ranges::view::take(1);
      |                                                               ^
In file included from /opt/compiler-explorer/libs/rangesv3/trunk/include/range/v3/view/all.hpp:28,
                 from /opt/compiler-explorer/libs/rangesv3/trunk/include/range/v3/view/take.hpp:28,
                 from <source>:3:
/opt/compiler-explorer/libs/rangesv3/trunk/include/range/v3/view/view.hpp:127:13: note: declared here
  127 |             operator|(Rng &&, view_closure<ViewFn> const &) // ******* READ THIS ********
      |             ^~~~~~~~
```

looking at 

https://github.com/ericniebler/range-v3/blob/1334a702e680dd407d53edc6230ee489d07a9186/include/range/v3/view/view.hpp#L122-L139

says something that I don't understand :)

I guess gcc10 should also be excluded like you did for clang.

BRevzin:
This fixes #1470 by making the two view concepts cross-compatible. range-v3 will just use std's if available, and we set `std::ranges::enable_view<T>` to true for types which set `ranges::enable_view` to true. 

The downside of this approach is that it has to `#include <ranges>`. Also I don't know how to test this on CI since you need pretty recent gcc trunk to do this.

### DIFF
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -25,6 +25,29 @@
 #include <range/v3/utility/static_const.hpp>
 #include <range/v3/version.hpp>
 
+// if the standard library ships <ranges>, play nice with it
+#if __has_include(<version>)
+#  include <version>
+#  if __cpp_lib_ranges >= 201911
+#    define RANGE_V3_STD_COMPAT 1
+     namespace std {
+         template <typename S, typename I>
+         extern const bool disable_sized_sentinel_for;
+     }
+     namespace std::ranges {
+         template <typename T>
+         extern const bool enable_view;
+
+         template <typename T>
+         extern const bool enable_borrowed_range;
+
+         template <typename T>
+         extern const bool disable_sized_range;
+     }
+#  endif
+#endif
+
+
 /// \defgroup group-iterator Iterator
 /// Iterator functionality
 
@@ -820,6 +843,19 @@ namespace ranges
     }
 } // namespace ranges
 /// \endcond
+
+#ifdef RANGE_V3_STD_COMPAT
+namespace std {
+    template <typename S, typename I>
+        requires ::ranges::disable_sized_sentinel<S, I>
+    constexpr bool disable_sized_sentinel_for<S, I> = true;
+}
+namespace std::ranges {
+    template <typename T>
+        requires ::ranges::enable_safe_range<T>
+    constexpr bool enable_borrowed_range<T> = true;
+}
+#endif
 
 RANGES_DIAGNOSTIC_POP
 
--- a/include/range/v3/range_fwd.hpp
+++ b/include/range/v3/range_fwd.hpp
@@ -28,7 +28,7 @@
 // if the standard library ships <ranges>, play nice with it
 #if __has_include(<version>)
 #  include <version>
-#  if __cpp_lib_ranges >= 201911
+#  if defined(__cpp_lib_ranges) && __cpp_lib_ranges >= 201911
 #    define RANGE_V3_STD_COMPAT 1
      namespace std {
          template <typename S, typename I>
--- a/include/range/v3/iterator/concepts.hpp
+++ b/include/range/v3/iterator/concepts.hpp
@@ -36,6 +36,18 @@
 
 #include <range/v3/detail/disable_warnings.hpp>
 
+#if __has_include(<version>)
+#  include <version>
+#  if __cpp_lib_ranges >= 201911
+#    define RANGE_V3_STD_COMPAT 1
+     namespace std {
+         template <typename S, typename I>
+         extern const bool disable_sized_sentinel_for;
+     }
+#  endif
+#endif
+
+
 namespace ranges
 {
     /// \addtogroup group-iterator-concepts
@@ -268,6 +280,9 @@ namespace ranges
             defer::sentinel_for<S, I>) &&
         #else
         (!disable_sized_sentinel<std::remove_cv_t<S>, std::remove_cv_t<I>>) &&
+        #ifdef RANGE_V3_STD_COMPAT
+        (!::std::disable_sized_sentinel_for<std::remove_cv_t<S>, std::remove_cv_t<I>>) &&
+        #endif
         defer::sentinel_for<S, I> &&
         #endif
         CPP_fragment(ranges::sized_sentinel_for_, S, I);
@@ -819,6 +834,14 @@ namespace ranges
         disable_sized_sentinel<std::reverse_iterator<S>, std::reverse_iterator<I>> =
             !static_cast<bool>(sized_sentinel_for<I, S>);
 } // namespace ranges
+
+#ifdef RANGE_V3_STD_COMPAT
+namespace std {
+    template <typename S, typename I>
+        requires ::ranges::disable_sized_sentinel<S, I>
+    constexpr bool disable_sized_sentinel_for<S, I> = true;
+}
+#endif
 
 #endif // defined(__GLIBCXX__) || (defined(_LIBCPP_VERSION) && _LIBCPP_VERSION <= 3900)
 
--- a/include/range/v3/iterator/concepts.hpp
+++ b/include/range/v3/iterator/concepts.hpp
@@ -36,18 +36,6 @@
 
 #include <range/v3/detail/disable_warnings.hpp>
 
-#if __has_include(<version>)
-#  include <version>
-#  if __cpp_lib_ranges >= 201911
-#    define RANGE_V3_STD_COMPAT 1
-     namespace std {
-         template <typename S, typename I>
-         extern const bool disable_sized_sentinel_for;
-     }
-#  endif
-#endif
-
-
 namespace ranges
 {
     /// \addtogroup group-iterator-concepts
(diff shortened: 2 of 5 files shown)


BRevzin:
This is actually still incomplete since I need to do the same thing for all the type traits (`enable_borrowed_range`, `disable_sized_range`, and `disable_sized_sentinel_for`) but those would basically look the same as this one - so if you hate this approach, you'll hate that one too 🙃

include/range/v3/range_fwd.hpp
```diff
@@ -25,6 +25,13 @@
 #include <range/v3/utility/static_const.hpp>
 #include <range/v3/version.hpp>
 
+#if __has_include(<ranges>)
+#  include <ranges>
+#  if __cpp_lib_ranges >= 201911
+#    define RANGE_V3_USE_STD_ENABLE_VIEW 1
+#  endif
```
ericniebler, line 32: I _really_ don't want to pull in all of `<ranges>`. I'm not sure how else to do this, though. Needs more thought.

BRevzin: gcc 10 isn't out yet, could ask Wakely very nicely to stick the variable template and the base class into its own header (right now they're both in `<ranges>`: https://github.com/gcc-mirror/gcc/blob/master/libstdc++-v3/include/std/ranges#L66-L69).

`<version>` exists now so could check the feature-test macro from there.

marehr: Couldn't we just forward declare `std::ranges::enable_view` and `std::ranges::view_base` and do the manual overload that I did in my issue? So ranges keeps it's base classes, but enables them as views?

I think you also forward declared std::span in ranges without actually pulling it in.

ericniebler: They can, but on MSVC, you need [`/Zc:externConstexpr`](https://docs.microsoft.com/en-us/cpp/build/reference/zc-externconstexpr?view=vs-2019).

```c++
template<class>
extern const bool enable_view;
```
 
Later...

```c++
template <class T>
inline constexpr bool enable_view = ...;
```

ericniebler: ... however, it isn't enough to merely forward-declare these variable templates as @BRevzin is doing here. It will lead to hard failures if the definition is not available at the point it is ODR used, as shown here: https://godbolt.org/z/Ewuq4k.

BRevzin: That's weird, because all the tests compile for me even though only one of them includes `<ranges>`...

CaseyCarter: You can't forward-declare variable templates; these are definitions.

include/range/v3/range_fwd.hpp
```diff
@@ -25,6 +25,29 @@
 #include <range/v3/utility/static_const.hpp>
 #include <range/v3/version.hpp>
 
+// if the standard library ships <ranges>, play nice with it
+#if __has_include(<version>)
```
ericniebler, line 29: `__has_include` is C++17. range-v3 still supports C++14. Please guard this with `#if defined(__has_include)`.

include/range/v3/range_fwd.hpp
ericniebler: > That's weird, because all the tests compile for me even though only one of them includes <ranges>...

I'm going to guess that these are getting pulled in from _somewhere_ besides `<ranges>`, but that it's not something we can or should be relying on.

BRevzin: @ericniebler @CaseyCarter Ok tracked that down. `<algorithm>` includes `<bits/ranges_algo.h>` includes `<bits/ranges_algobase.h>` includes `<ranges>`. So that's probably how it's getting in (although really, every unit test in range-v3 includes `<algorithm>`?)

I guess this direction is a non-starter then.

ericniebler:
I've had another thought for how this interoperability might be achieved, at least for `enable_view`. It is possible to forward-declare the struct `std::ranges::view_base`. Then we can change the definition of range-v3's `enable_view<T>` to default to `convertible_to<T*, view_base*>`. This will catch the majority of the cases we care about without the need to #include <ranges>.

marehr:
Polite ping, did you manage to make any progress here? For me it would be okay to include a new header which enables C++20 compatibility if `#include <ranges>` is to heavy to include in a forward header.

BRevzin:
To be honest, I don't really know how else to do this right without including `<ranges>` _early_. 

@ericniebler's suggestion would allow `std` views to be recognized by `range-v3` but it would not allow `range-v3` views to be recognized by `std`.

ericniebler:
> To be honest, I don't really know how else to do this right without including `<ranges>` _early_.

I also can't think of another way either. :-( I suppose including `<ranges>` will have to do, at least until implementations give us an easier way. Should've thought of proposing a `<ranges_fwd>` header. I made my bed. Gotta lie in it.

BRevzin:
> 
> 
> > To be honest, I don't really know how else to do this right without including `<ranges>` _early_.
> 
> I also can't think of another way either. :-( I suppose including `<ranges>` will have to do, at least until implementations give us an easier way. Should've thought of proposing a `<ranges_fwd>` header. I made my bed. Gotta lie in it.

Do you think this is something that could be viewed, through the right lens, as a library defect or is it something that would need to go LEWG? All of this becomes less important when we modularize the library, but until then... given how much more functionality is in range-v3 than in C++20, it seems important to have that compatibility.

On the other hand, range-v3 compile times are sufficiently large that... how much worse do they get from `#include <ranges>`?

marehr:
Hi, just a polite ping. From what I can see, we were pretty close to merging this.

> On the other hand, range-v3 compile times are sufficiently large that... how much worse do they get from `#include <ranges>`?

```c++
#ifdef USE_STD_RANGE
#include <ranges>
#endif

#include <range/v3/view/all.hpp>

static constexpr char str[]{"Hello World"};

int main()
{
    auto view = ranges::views::all(str);
    return ranges::size(view);
}
```

On gcc-10.2 with `-O3` and `-ftime-report`:

|    | -std=c++17 | -std=c++2a |
| -- | --------------- | --------------- |
| without `#include <ranges>` | 1.89s | 2.35 |
| with `#include <ranges>` | 1.98s | 2.35 |

(times are from godbolt https://godbolt.org/z/shEvrK, so might not be the stablest)

BRevzin:
I'm going to close this PR for now.

gcc 11 is going to have a [`bits/ranges_base.h`](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/bits/ranges_base.h) header (thanks Mr. Wakely) that is a much smaller header that has the pieces necessary here (`enable_view`, `view_base`, `enable_borrowed_range`, `disable_sized_range`) that should be more acceptable to use. I'm hoping MSVC (ahem @CaseyCarter) does something similar.

That way, we can do something here like:

```cpp
#ifdef __has_include
#  if __has_include(<version>)
#    include <version>
#    ifdef __cpp_lib_ranges
#      if __has_include(<bits/ranges_base.h>)
#        include <bits/ranges_base.h>
#        define RANGE_V3_USE_STD_RANGES 1
#      elif __has_include(<ranges>)
#        include <ranges>
#        define RANGE_V3_USE_STD_RANGES 1
#      endif
#    endif
#  endif
#endif
```

And then if `RANGE_V3_USE_STD_RANGES` is defined, then do the other stuff I did in this PR. `ranges::view_interface` and `ranges::view_base` can alias the `std` versions. The variable templates need a bit more finesse (e.g. need to specialize `std::ranges::enable_view` whenever `ranges::enable_view` is `true` not by default). I think that should do it then?

That will allow `r | ranges::some_view | std::ranges::some_view | ranges::some_view` to work, but it won't allow just `ranges::some_view | std::ranges::some_view` to work. But to do that, we'd need to figure out how to standardize "range adapter closure" stuff.

cjdb:
> I'm hoping MSVC (ahem @CaseyCarter) does something similar.

I'll keep this in mind as I restart ranges work in libc++ this year.

PR #1474, opened 2020-04-25
MSVC bug workaround updates:

CaseyCarter:
* Don't build with the clang-cl flag `-fno-ms-compatibility` for now.
* VSO-573728 was fixed for 16.3
* VSO-654601 workaround is actually for VSO-784772 (What I thought was a regression was actually a distinct bug hidden behind the old workaround.)
* VSO-779708 was fixed for 16.5
* Workarounds for VSO-780775 were removed; there's no reason to keep the macro around.
* VSO-895622 was fixed for 16.6


ericniebler:
Thanks Casey!

PR #1476, opened 2020-04-28
add data to view_interface

### LINKED ISSUE #1402: view_interface is missing .data()

dvirtz:
Resolves #1402

### DIFF
--- a/include/range/v3/detail/range_access.hpp
+++ b/include/range/v3/detail/range_access.hpp
@@ -337,15 +337,20 @@ namespace ranges
             bidirectional_cursor<T> && //
             sized_sentinel_for_cursor<T, T> && //
             CPP_fragment(detail::random_access_cursor_, T);
+        template<class T>
+        auto is_lvalue_reference(T&&)
+            -> CPP_ret(void)(requires std::is_lvalue_reference<T>::value);
         template<typename T>
-        CPP_concept_bool contiguous_cursor_ = 
-        std::is_lvalue_reference<
-                    decltype(range_access::read(std::declval<T&>()))>::value;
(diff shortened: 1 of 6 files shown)


include/range/v3/detail/range_access.hpp
```diff
@@ -338,17 +338,14 @@ namespace ranges
...
             CPP_fragment(detail::random_access_cursor_, T);
         template<typename T>
-        CPP_concept_fragment(contiguous_cursor_,
-            requires(T & t) //
-            (
-                concepts::requires_<std::is_lvalue_reference<
-                    decltype(range_access::read(t))>::value>
-            ));
+        CPP_concept_bool contiguous_cursor_ = 
+        std::is_lvalue_reference<
```
ericniebler, line 342: Indent here, please (spaces, not tabs).

include/range/v3/detail/range_access.hpp
```diff
@@ -338,17 +338,14 @@ namespace ranges
...
+                    decltype(range_access::read(std::declval<T&>()))>::value;
```
ericniebler, line 343: I think this is incorrect. You are in a non-SFINAE context here. If no `read` overload is found, this will be a hard error in concepts emulation mode.

What was wrong with the original?

dvirtz: You're right.  MSVC `/std:c++latest` choked on the original version and I can't think of another workaround.

I opened a bug for MSVC: https://developercommunity.visualstudio.com/content/problem/1010462/concepts-requires-miscompilation.html.

ericniebler:
Thanks.

PR #1479, opened 2020-05-01
Add a missing `#include <range/v3/view/remove.hpp>`

keryell:
Using a `#include <range/v3/all.hpp>` was not providing `views::remove`.


keryell:
Perhaps there should be some unit-tests testing that each feature `#include` is actually provided by the parent header too. But it might be also some over-engineering...

PR #1487, opened 2020-05-10
fix #1486 zip and zip_with cardianlity

### LINKED ISSUE #1486: cardinality of zip of known cardinalities is only finite
The last static assert in the following code should succeed but fails, as the actual value is `finite`:

```c++
int ai[] = {1, 2, 3};
int bi[] = {4, 5, 6};
static_assert(range_cardinality<decltype(ai)>::value == 3);
static_assert(range_cardinality<decltype(bi)>::value == 3);
auto zipped = views::zip(ai, bi);
static_assert(range_cardinality<decltype(zipped)>::value == 3);
```

see https://gcc.godbolt.org/z/_DGT6g

dvirtz:


ericniebler:
Thanks.

PR #1493, opened 2020-05-20
fix #1322 by testing boolean via explicit conversion to bool instead of contextually converting

### LINKED ISSUE #1322: some core language concepts without std::boolean broken in 0.9.x
```cpp
#include <concepts/concepts.hpp>

struct S
{};

template <typename T>
void foobar(T &&) {}

template <typename T>
    requires concepts::totally_ordered<T>
void foobar(T &&) {}


int main()
{
    std::pair<S, int> p;
    foobar(p);
}
```
(build with `-fconcepts` and GCC)

Instead of the first overload being chosen there is a hard compile error.

marehr:
Fixes #1322

GCC uses the following definition:

```c++
template<typename _Tp>
concept __boolean_testable_impl = convertible_to<_Tp, bool>;

template<typename _Tp>
concept __boolean_testable
= __boolean_testable_impl<_Tp>
  && requires(_Tp&& __t)
  { { !static_cast<_Tp&&>(__t) } -> __boolean_testable_impl; };
```

AFAIS they test implicit and explicit conversion to bool, as well
as negating a temporary which also needs to be implicitly and
explicitly convertible to bool.

I'm unsure why this is needed and how this relates to the current
implementation of contextually convertible it to bool.


ericniebler:
I mean, it's not worse than what's there currently, but it isn't right either. The GCC implementation is correct.

marehr:
> I mean, it's not worse than what's there currently, but it isn't right either. The GCC implementation is correct.

I can also implement the gcc solution, but I don't know how to do this construct with your abstraction layer:

```c++
requires(_Tp&& __t)
  { { !static_cast<_Tp&&>(__t) } -> __boolean_testable_impl; };
```

some pointers would be helpful :)

ericniebler:
something like (untested):

```c++
    template<typename T>
    CPP_concept_fragment(boolean_testable_,
        requires(T && t) //
        (
            concepts::requires_<boolean_testable_impl_<decltype(!(T&&) t)>>,
        ) &&
        boolean_testable_impl_<T>
    );
```

marehr:
and 

```c++
template<typename _Tp>
concept __boolean_testable_impl = convertible_to<_Tp, bool>;
```

would be

```c++
    template<typename T>
    CPP_concept_bool boolean_testable_impl_ = convertible_to<T, bool>;
```

right?

ericniebler:
Yes, and then:

```c++
template<typename T>
CPP_concept_bool boolean_testable = CPP_fragment(boolean_testable_, T);
```

marehr:
would it be okay to move `concepts::detail::weakly_equality_comparable_with_` into `concepts::defs::detail::weakly_equality_comparable_with_` namespace?

or should I do something along this line

```c++
inline namespace concepts::defs // of course in a C++14 way
{
// concepts ...
// covertible_to
}

namespace concepts::detail
{
// boolean_testable_impl_
}

inline namespace concepts::defs // of course in a C++14 way
{
// remaining concepts
// concept equality_comparable
}
```

or should I define `boolean_testable_impl_` and `weakly_equality_comparable_with_` within `concepts::defs` directly?

ericniebler:
They are details. They can be moved.

marehr:
Okay, so I will change `using ::concepts::detail::weakly_equality_comparable_with_;` accordingly.

marehr:
@ericniebler Where to put the regression test

```c++
#include <concepts/concepts.hpp>

struct S
{};

template <typename T>
void foobar(T &&) {}

template <typename T>
    requires concepts::totally_ordered<T>
void foobar(T &&) {}


int main()
{
    std::pair<S, int> p;
    foobar(p);
}
```

?

marehr:
@ericniebler I think this is now ready for review :)

include/concepts/concepts.hpp
```diff
@@ -1043,9 +1025,43 @@ namespace concepts
...
+            requires(detail::as_cref_t<T> t, detail::as_cref_t<U> u) //
+            (
+                concepts::requires_<boolean_testable_<decltype(t == u)>>,
+                concepts::requires_<boolean_testable_<decltype(t != u)>>,
+                concepts::requires_<boolean_testable_<decltype(u == t)>>,
+                concepts::requires_<boolean_testable_<decltype(u != t)>>
+            ));
+        template<typename T, typename U>
+        CPP_concept_bool weakly_equality_comparable_with_ =
+            CPP_fragment(weakly_equality_comparable_with_frag_, T, U);
```
ericniebler, line 1058: I thought you were going to keep this in a detail namespace. I don't want it in the `ranges::` namespace. I apologize if I was confused about what you asking earlier.

ericniebler:
Great! Thanks. I'll wait for the tests to cycle and then merge.

include/concepts/concepts.hpp
marehr: Okay I kept the namespace `concepts::detail` by closing the `concepts::defs` namespace and reopening it later. I couldn't just add a `concepts::defs::detail` namespace, since it would clash with `concepts::detail.`

marehr:
> Great! Thanks. I'll wait for the tests to cycle and then merge.

That scared me, you are to fast :)

marehr:
@ericniebler Tests passed

ericniebler:
Thank you!

PR #1494, opened 2020-05-22
assignable_from: use the definition from the C++20 standard

marehr:
https://eel.is/c++draft/concept.assignable

I noticed with this code that the current implementation of `concepts::assignable_from` is not the same as the one from the standard.

```c++
#include <concepts/concepts.hpp>

template <class LHS, class RHS>
concept assignable_from =
    // ::concepts::common_reference_with<std::remove_reference_t<LHS> const &, std::remove_reference_t<RHS> const&> &&
    ::concepts::assignable_from<LHS, RHS>;

int main() {
  using int_simd [[gnu::vector_size(sizeof(int) * 1)]] = int;
  using uint_simd [[gnu::vector_size(sizeof(unsigned) * 1)]] = unsigned;

//   int_simd a = uint_simd{};
  static_assert(::concepts::common_reference_with<std::remove_reference_t<int> const &, std::remove_reference_t<unsigned> const&>);
  static_assert(!::concepts::common_reference_with<std::remove_reference_t<int_simd> const &, std::remove_reference_t<uint_simd> const&>);
  static_assert(!assignable_from<int_simd &, uint_simd &>);
}
```
https://godbolt.org/z/ZbSt8z
(I know this is not a "normal" use case)

### DIFF
--- a/include/concepts/concepts.hpp
+++ b/include/concepts/concepts.hpp
@@ -1013,6 +1013,7 @@ namespace concepts
         template<typename T, typename U>
         CPP_concept_bool assignable_from =
             std::is_lvalue_reference<T>::value &&
+            common_reference_with<std::remove_reference_t<T> const &, std::remove_reference_t<U> const &> &&
             CPP_fragment(defs::assignable_from_, T, U);
 
         template<typename T>
--- a/include/concepts/concepts.hpp
+++ b/include/concepts/concepts.hpp
@@ -1013,7 +1013,7 @@ namespace concepts
         template<typename T, typename U>
         CPP_concept_bool assignable_from =
             std::is_lvalue_reference<T>::value &&
-            common_reference_with<std::remove_reference_t<T> const &, std::remove_reference_t<U> const &> &&
+            common_reference_with<detail::as_cref_t<T>, detail::as_cref_t<U>> &&
             CPP_fragment(defs::assignable_from_, T, U);
 
         template<typename T>
--- a/include/range/v3/utility/common_tuple.hpp
+++ b/include/range/v3/utility/common_tuple.hpp
@@ -182,7 +182,7 @@ namespace ranges
         auto operator=(std::tuple<Us...> & that) noexcept(
             meta::and_c<std::is_nothrow_assignable<Ts &, Us &>::value...>::value)
             -> CPP_ret(common_tuple &)( //
-                requires assignable_from<detail::args<Ts...> &, detail::rargs<Us...>>)
+                requires std::is_assignable<detail::args<Ts...> &, detail::rargs<Us...>>::value)
         {
             (void)tuple_transform(base(), that, element_assign_{});
             return *this;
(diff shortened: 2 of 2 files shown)


ericniebler:
This cannot proceed until the test failures are addressed.

marehr:
Yeah, I know. I'm a bit stuck, because I don't know how to resolve this.

The only test that is failing is `test/iterator/basic_iterator.cpp`.

The failures are due to the following lines:

https://github.com/ericniebler/range-v3/blob/21570a735a7b02a150e63ddaef6ace354ccf195f/test/iterator/basic_iterator.cpp#L431-L437

(`*i = std::tuple<MoveOnly>{};` and `*i = std::move(v);`)

Which delegate to

https://github.com/ericniebler/range-v3/blob/21570a735a7b02a150e63ddaef6ace354ccf195f/test/iterator/basic_iterator.cpp#L404-L407

```c++
error: no match for ‘operator=’ (operand types are 
  ‘test_move_only::zip1_cursor<test_move_only::MoveOnly*>::reference’ 
    {aka ‘ranges::common_tuple<test_move_only::MoveOnly&>’} and 
  ‘std::remove_reference<std::tuple<test_move_only::MoveOnly>&>::type’
    {aka ‘std::tuple<test_move_only::MoveOnly>’})
   82 |             reference{ *it_} = std::move(v);
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
```

Before the change it used this assignment operator of `common_tuple`

https://github.com/ericniebler/range-v3/blob/21570a735a7b02a150e63ddaef6ace354ccf195f/include/range/v3/utility/common_tuple.hpp#L200-L208

After the change the requirement `assignable_from<detail::args<Ts...> &, detail::args<Us...>>` is not true anymore.

This means that 
```c++
common_reference_with<detail::as_cref_t<detail::args<Ts...> &>, detail::as_cref_t<detail::args<Us...>>> = 
common_reference_with<detail::args<Ts...> const &, detail::args<Us...> const &> 
```
must not be true.

gcc says

```c++
note: candidate: ‘ranges::common_tuple<Ts>& ranges::common_tuple<Ts>::operator=(std::tuple<_Args1 ...>&&) requires  assignable_from<ranges::detail::args_<sizeof ... (Ts ...), Ts ...>&, ranges::detail::args_<sizeof ... (Us ...), Us ...> > [with Us = {test_move_only::MoveOnly}; Ts = {test_move_only::MoveOnly&}]’
  201 |         auto operator=(std::tuple<Us...> && that) noexcept(
      |              ^~~~~~~~
```

that means
```c++
assignable_from<ranges::detail::args_<1, test_move_only::MoveOnly&>&,
                ranges::detail::args_<1, test_move_only::MoveOnly> > =

common_reference_with<ranges::detail::args_<1, test_move_only::MoveOnly&> const &,
                      ranges::detail::args_<1, test_move_only::MoveOnly> const &> 
```

has no common type..

My guess is that 
* `ranges::detail::args_` now would need a `concepts::common_reference::type` overload in addition to requiring that any inner type of `ranges::detail::args_` is assignable or that
* we only require a weaker version of `std::assignable_from`, like `is_assignable_v`, to decide whether to use the overload or not.

https://github.com/ericniebler/range-v3/blob/21570a735a7b02a150e63ddaef6ace354ccf195f/include/range/v3/utility/common_tuple.hpp#L74-L86

This is how far I got.

marehr:
Okay I tried `std::is_assignable` instead of `std::assignable_from` and it worked.

I think the semantics of `std::assignable_from` are not needed in this case, because `ranges::detail::args_` seems to be a helper structure to test whether the inner types of a (common/std) tuple are assignable to each other.

ericniebler:
Looks good. Thanks.

PR #1498, opened 2020-05-30
adds lexicographical_compare_three_way

cjdb:
Blocked by #1497.
Would appreciate a review from @BRevzin, if possible.


include/concepts/compare.hpp
```diff
@@ -0,0 +1,66 @@
+/// \file
```
cjdb, line 1: This file will be merged by #1497.

test/utility/concepts.cpp
```diff
@@ -1,6 +1,7 @@
 // Range v3 library
```
cjdb, line 1: This file will be merged by #1497.

PR #1501, opened 2020-05-31
Fix1477

### LINKED ISSUE #1477: all_t alias not in cpp20 namespace
Hi, I am just wondering why `all_t` was not yet adopted to the ranges::cpp20 namespace. It seems the standard defines now this transformation type trait in favour of all_view.
Is that on purpose or not yet updated? 
If so I can also make a PR for this.

Thanks a lot

marehr:
Fixes #1477

### DIFF
--- a/include/range/v3/range/access.hpp
+++ b/include/range/v3/range/access.hpp
@@ -43,21 +43,21 @@ namespace ranges
 #if defined(__cpp_lib_string_view) && __cpp_lib_string_view > 0
     template<class CharT, class Traits>
     RANGES_INLINE_VAR constexpr bool
-        enable_safe_range<std::basic_string_view<CharT, Traits>> = true;
+        enable_borrowed_range<std::basic_string_view<CharT, Traits>> = true;
 #endif
 
 #if defined(__cpp_lib_span) && __cpp_lib_span > 0
     template<class T, std::size_t N>
-    RANGES_INLINE_VAR constexpr bool enable_safe_range<std::span<T, N>> = true;
+    RANGES_INLINE_VAR constexpr bool enable_borrowed_range<std::span<T, N>> = true;
 #endif
 
     namespace detail
     {
         template<typename T>
-        RANGES_INLINE_VAR constexpr bool _safe_range = enable_safe_range<uncvref_t<T>>;
+        RANGES_INLINE_VAR constexpr bool _borrowed_range = enable_borrowed_range<uncvref_t<T>>;
 
         template<typename T>
-        RANGES_INLINE_VAR constexpr bool _safe_range<T &> = true;
+        RANGES_INLINE_VAR constexpr bool _borrowed_range<T &> = true;
     } // namespace detail
 
     /// \cond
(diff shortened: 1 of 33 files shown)


include/range/v3/range/access.hpp
```diff
@@ -726,6 +726,9 @@ namespace ranges
 
         using ranges::iterator_t;
         using ranges::sentinel_t;
+
+        template<typename T>
+        RANGES_INLINE_VAR constexpr bool enable_borrowed_range = ranges::enable_safe_range<T>;
```
ericniebler, line 731: I would rather make `enable_borrowed_range` the primary, and mark `enable_safe_range` as deprecated. Ditto for the other entities that got renamed.

include/range/v3/range/access.hpp
```diff
@@ -726,6 +726,8 @@ namespace ranges
 
         using ranges::iterator_t;
         using ranges::sentinel_t;
+
+        using ranges::enable_borrowed_range;
```
marehr, line 730: Should I add a deprecation message for `ranges::cpp20::enable_safe_range` ?

include/range/v3/range/concepts.hpp
```diff
@@ -316,6 +316,7 @@ namespace ranges
 
     namespace cpp20
     {
+        using ranges::borrowed_range;
```
marehr, line 319: Should I add a deprecation message for `ranges::cpp20::borrowed_range` ?

marehr:
Build failures are likely unrelated to my changes, because the current master is also failing.

include/range/v3/range/access.hpp
ericniebler: Yes, please.

include/range/v3/range/concepts.hpp
ericniebler: Yep.

marehr: Can you give me some pointers? I tried to look up the standard and I couldn't find a way to add  attributes to concepts.

gcc complains:

```c++
template<typename T>  concept [[deprecated("HEllo")]] True  = true;

static_assert(True<int>);
```

```
<source>:1:47: error: expected unqualified-id

template<typename T>  [[deprecated("HEllo")]] concept True  = true;
```

I guess that this is not possible. I don't know if a deprecated namespace would work?

marehr: @ericniebler I figured out two styles:

```c++
template <typename R>
RANGES_DEPRECATED("Please use ranges::borrowed_range instead.")
RANGES_INLINE_VAR constexpr bool safe_range = borrowed_range<R>;
```

* + It will work for all major compilers
* - safe_range is not a concept any more and might make problems in overload resolution; on the other hand you get a BIG use this concept instead message
* - might change semantic

```c++
namespace detail
{
template <typename R>
RANGES_DEPRECATED("Please use ranges::borrowed_range instead.")
RANGES_INLINE_VAR constexpr bool safe_range_is_deprecated = true;
}

template <typename R>
CPP_concept_bool safe_range = borrowed_range<R> && detail::safe_range_is_deprecated<R>;
```

* - It will **NOT** work for all major compilers (except for >= clang-10 with `-std=c++2a`); This is also okay, because normally you compile with at least a second compiler.
* + safe_range is a concept like before
* + does not change semantic

https://godbolt.org/z/CMCoNC (Red is good, because it triggered a deprecation message)

Any suggestions?!

include/range/v3/range/access.hpp
marehr: done

include/range/v3/range/concepts.hpp
marehr: (I used the second style)

marehr:
Very polite ping, waiting on instructions :)

include/range/v3/range/concepts.hpp
ericniebler: I think I prefer the first style.

include/range/v3/view/istream.hpp
```diff
@@ -112,7 +112,7 @@ namespace ranges
 
     namespace cpp20
     {
-        template<typename Val>
+        template<typename Val, class CharT = char, class Traits = std::char_traits<CharT>>
```
ericniebler, line 115: I'd rather we leave this alone until we make a conforming `basic_istream_view`. Probably a separate PR.

marehr:
@ericniebler everything is passing ;)

ericniebler:
Thank you!

ericniebler:
I think this PR missed the `safe_iterator_t` -> `borrowed_iterator_t` name change.

marehr:
Oops, sorry I can create a new PR :)

PR #1511, opened 2020-06-19
ranges::sample is in C++20 already

al-mission-2016:
See i.e. http://eel.is/c++draft/alg.random.sample


ericniebler:
Dude, sweet.

PR #1515, opened 2020-06-22
make _is_integer_like_<__int128> = true, fixes #1514

### LINKED ISSUE #1514: gcc -std=c++2a iota's implementation can return __int128 as difference type
`gcc -std=c++2a` iota's implementation can return `__int128` as difference type.

This is a problem, because `std::ranges::iota` might not be considered a range / view in ranges-v3.

```c++
using diff_t = std::ranges::range_difference_t<std::ranges::iota_view<size_t, size_t>>;
static_assert(std::same_as<diff_t, __int128>); // ranges returns ranges::detail::diffmax_t
```
https://godbolt.org/z/RG6Y3e

Such a difference type is bad, because `std::ranges::begin` requires an `std::input_or_output_iterator` which needs to be `std::weakly_incrementable` which itself has this weird requirement:

> The range of representable values of an integer-class type is the continuous set of values over which it is defined. The values 0 and 1 are part of the range of every integer-class type.
If any negative numbers are part of the range, the type is a signed-integer-class type; otherwise, it is an unsigned-integer-class type.
> https://eel.is/c++draft/iterator.concept.winc#3

Which means that this requirement can be some augmented expression `std::is_signed_v || more signed like types`.

See the gcc implementation: 
* iota_view::iterator::difference_type: https://github.com/gcc-mirror/gcc/blob/0164e59835de81d758fd4c56248ad7a46435fbfa/libstdc%2B%2B-v3/include/std/ranges#L582-L602
* weakly_incrementable signed ness check: https://github.com/gcc-mirror/gcc/blob/0164e59835de81d758fd4c56248ad7a46435fbfa/libstdc%2B%2B-v3/include/bits/iterator_concepts.h#L512-L544

I have seen that ranges has `_is_integer_like_` to declare types as integer like.

https://github.com/ericniebler/range-v3/blob/de24d443f93428f859e7f66869ea0fc7da2f047b/include/range/v3/iterator/concepts.hpp#L164-L170

Which is specialised for `ranges::detail::diffmax_t`

https://github.com/ericniebler/range-v3/blob/de24d443f93428f859e7f66869ea0fc7da2f047b/include/range/v3/iterator/diffmax_t.hpp#L343-L349

---

I would propose to add

```c++
        // gcc10 uses for std::ranges::range_difference_t<
        // std::ranges::iota_view<size_t, size_t>> == __int128
#if __SIZEOF_INT128__
#if RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
        template<>
        inline constexpr bool _is_integer_like_<__int128> = true;
#else
        template<typename Enable>
        constexpr bool _is_integer_like_<__int128, Enable> = true;
#endif
#endif // __SIZEOF_INT128__
```

to `range-v3/include/range/v3/iterator/concepts.hpp`.

marehr:
fixes #1514

### DIFF
--- a/include/range/v3/iterator/concepts.hpp
+++ b/include/range/v3/iterator/concepts.hpp
@@ -169,6 +169,19 @@ namespace ranges
         constexpr bool _is_integer_like_ = std::is_integral<D>::value;
 #endif
 
+        // gcc10 uses for std::ranges::range_difference_t<
+        // std::ranges::iota_view<size_t, size_t>> == __int128
+#if __SIZEOF_INT128__
+        __extension__ typedef __int128 int128_t;
+#if RANGES_CXX_INLINE_VARIABLES >= RANGES_CXX_INLINE_VARIABLES_17
+        template<>
+        inline constexpr bool _is_integer_like_<int128_t> = true;
+#else
+        template<typename Enable>
+        constexpr bool _is_integer_like_<int128_t, Enable> = true;
+#endif
+#endif // __SIZEOF_INT128__
+
         // clang-format off
         template<typename D>
         CPP_concept_bool integer_like_ = _is_integer_like_<D>;


ericniebler:
Thanks!

Issue #801, opened 2018-02-23
Iteration of a function


vbeffara:
Hi,

Maybe I looked for the wrong keywords, but it would be useful to have a view iterating application of a function from an initial value, like `iterate` in Haskell:

```c++
template <typename T> struct iterate : public ranges::view_facade<iterate<T>> {
    iterate() = default;
    explicit iterate(T ii, function<T(T)> ff) : i(ii), f(ff) {}

    const T & read() const { return i; }
    bool      equal(ranges::default_sentinel /*unused*/) const { return false; }
    void      next() { i = f(std::move(i)); }

    T              i{};
    function<T(T)> f;
};
```

Easy enough to implement in user code, or to emulate using `iota` or something, but that feels like a common enough pattern to be part of the library. Or is it already and I didn't find it?

ericniebler:
Interestingly, the standard `<numeric>` header doesn't have an analogous algorithm for this. This is basically a generalization of `iota` where the successor function can be user-specified, rather than hard-coded to be increment.

I agree this would be a nice algorithm/view to add. I would probably call it `successor`, which is more Stepanov-y, and doesn't overload the term "iterate", which in C++ is already laden with meaning. _I think_ it is also better than trying to shoehorn this functionality into `ranges:[:view:]:iota` as it would complicate the constraints.

CaseyCarter:
`view::generate([state, fn] -> state const& { return state = fn(std::move(state)); })` is *very* close, if you can live without the 0th term.

vbeffara:
Good point :-) one can even catch the 0th term by hand (and btw one should make the lambda mutable):

```c++
template <typename T, typename F> auto iter(T t, F f) {
    return view::generate([t, f, first = true]() mutable -> T const & {
        if (!first) t = f(std::move(t));
        first = false;
        return t;
    });
}
```

vbeffara:
Pointing out for future reference that #807 makes this way of writing it lead to undefined behavior in some cases.

apmccartney:
I took this as an exercise to familiarize myself with some of the mechanics used in the range-v3 implementation and I am struggling. It seems I've misunderstood what is required to fullfill the cursor requirements.

My initial attempt has as follows (please excuse the lack of concept checking in this attempt):
```cpp
#include <range/v3/all.hpp>

template<typename State, typename Function>
struct iterate_view :
  public ranges::view_facade<iterate_view<State, Function>, ranges::infinite>{
protected:
  State initial;
[...]

PR #1517, opened 2020-06-25
Add views::successor

sv1990:
Implements #801.


include/range/v3/view/successor.hpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        struct successor_fn
+        {
+            template<typename G, typename T>
+            auto operator()(G g, T x) const -> CPP_ret(successor_view<G, T>)( //
+                requires invocable<G &, T &&> && copy_constructible<G> &&
+                    copy_constructible<T> &&
+                        assignable_from<T &, invoke_result_t<G &, T &&>>)
+            {
+                return successor_view<G, T>{std::move(g), std::move(x)};
```
johelegp, line 101: `g` and `t` must be forwarding references so that `successor_view{g, x}` is equivalent to `views::successor(g, x)`.

sv1990: Thanks for the suggestion, but I think this is the way other views are implemented as well. See for example [views::generate](https://github.com/ericniebler/range-v3/blob/de24d443f93428f859e7f66869ea0fc7da2f047b/include/range/v3/view/generate.hpp#L97) or [views::repeat](https://github.com/ericniebler/range-v3/blob/de24d443f93428f859e7f66869ea0fc7da2f047b/include/range/v3/view/repeat.hpp#L105).

johelegp: You're right. Let's go for local consistency.

johelegp: My misunderstanding may be related to http://wg21.link/LWG3509, although I'm still trying to get my head around it.

PR #1520, opened 2020-06-27
Add Conan installation instructions

Talkless:
Update README.md with example how to use Conan package manager to get
Range v3 library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,40 @@ You can download and install range-v3 using the [vcpkg](https://github.com/Micro
 
 The range-v3 port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
+Building range-v3 - Using Conan
+-------------------------------
+
+You can download and install range-v3 using the [Conan](https://github.com/conan-io/conan) dependency manager.
+
+Setup your CMakeLists.txt (see [Conan documentation](https://docs.conan.io/en/latest/integrations/build_system.html) on how to use MSBuild, Meson and others):
+```
+project(myproject CXX)
+
+add_executable(${PROJECT_NAME} main.cpp)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake) # Include Conan-generated file
+conan_basic_setup(TARGETS) # Introduce Conan-generated targets
+
+target_link_libraries(${PROJECT_NAME} CONAN_PKG::range-v3)
+```
+Create `conanfile.txt` in your source dir:
+```
+[requires]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,40 @@ You can download and install range-v3 using the [vcpkg](https://github.com/Micro
...
+
+You can download and install range-v3 using the [Conan](https://github.com/conan-io/conan) dependency manager.
+
+Setup your CMakeLists.txt (see [Conan documentation](https://docs.conan.io/en/latest/integrations/build_system/cmake.html) on how to use `find_package` with CMake instead, and how to use different [buid systems](https://docs.conan.io/en/latest/integrations/build_system.html) too):
+```
+project(myproject CXX)
+
+add_executable(${PROJECT_NAME} main.cpp)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake) # Include Conan-generted file
```
marehr, line 94:
```suggestion
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake) # Include Conan-generated  file
```

README.md
```diff
@@ -80,6 +80,40 @@ You can download and install range-v3 using the [vcpkg](https://github.com/Micro
 
 The range-v3 port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
+Building range-v3 - Using Conan
+-------------------------------
+
+You can download and install range-v3 using the [Conan](https://github.com/conan-io/conan) dependency manager.
+
+Setup your CMakeLists.txt (see [Conan documentation](https://docs.conan.io/en/latest/integrations/build_system/cmake.html) on how to use `find_package` with CMake instead, and how to use different [buid systems](https://docs.conan.io/en/latest/integrations/build_system.html) too):
```
ericniebler, line 88: This parenthetical here seems to confuse rather than clarify. Are users to set up their CMakeLists.txt as you show below? Or are they to do something different using `find_package`? Why should they prefer one over the other? 

The reference to different build systems further confuses me. CMake isn't really a build system, it's a generator. If you're going to mention build systems at all, can you be clearer about how it relates to the discussion of CMake and Conan?

Talkless: >  CMake isn't really a build system, it's a generator.

While technically this is correct, I believe it's too nit-picky to not call CMake a build system. How else do we call all similar software that allows to declare how your project is being built? Conan documentation uses "build systems" tern in the link provided too, bringing CMake, Make, and others together.

Anyhow, I will simplify parentheses. All that info is probably off-topic for a quick start example. I wanted for better but.. :-) .

ericniebler:
Better, thanks.

PR #1524, opened 2020-07-07
rename safe_iterator_t -> borrowed_iterator_t

marehr:


marehr:
(Closed / Re-opened since a build was failing in set-up)

PR #1536, opened 2020-07-28
[Docs] views::transform usage example

Dig-Doug:
Adds a simple example for views::transform and links to it in the docs.

Background: The "List of range views" in the docs is very useful, however when you click on a link in the list, e.g. [views::transform](https://ericniebler.github.io/range-v3/structranges_1_1views_1_1transform__fn.html), you land on a page without more info. This PR fixes that problem by creating a very simple usage example and embedding it in the docs as a snippet.

I've only done 1 view to check this format is acceptable.

Screenshot of the docs:

Screenshot from 2020-07-28 08-44-21


BRevzin:
I think this is an improvement but we need to do more.

There are a few things that are really important to know about the range adapters that need to be documented - namely what the parameters mean, how they are used, and what are the types and value categories involved, and what is the range category of the resulting range. For transform, I think we need to talk about:

1) It takes a range of `T` and a function `T -> U` and yield a range of `U`. Not `decay_t<U>`, but actually `U`.
2) It preserves the category of the inbound range (except contiguous ranges are reduced to random access ranges)

Also, while the example is useful, the other Doxygen stuff here is just... not relevant at all. Ideally we just eschew any kind of auto-generation and just write everything by hand. Part of why the generated stuff is so bad is because C++ function declarations are [miserably illegible](https://brevzin.github.io/c++/2018/10/20/concepts-declarations/). 

Good examples to draw insight from are like how [Haskell](https://hackage.haskell.org/package/groupBy-0.1.0.0/docs/Data-List-GroupBy.html) and [Rust](https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.group_by) document something like `group_by` (which have different meanings in the two languages - one takes a binary predicate the other takes a unary transform - but it's pretty clear what each one means). The implementation details (i.e. what Doxygen generations for `transform_fn` - a type that nobody cares about) aren't in the docs, as they shouldn't be. 

So basically - example is great (though could be more concise - you don't have to turn it into a `vector` to index into it) - but we need a lot more info.

Dig-Doug:
Thanks for your feedback. I've included another revision in this comment instead updating the code for now. I included all of the information you suggested. The format is inspired by [cpp ref](https://en.cppreference.com/w/cpp/algorithm/transform) and [moz js docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

Regarding "eschewing auto-generation", are you suggesting using something like [readthedocs](https://readthedocs.org/)?

Also agreed on the example. I want to make sure the program compiles and does what it says (aka. assert the output), but I can also do that separately (e.g. assert in a bash script) so it's hidden from the reader.

-----

# ranges::views::transform

The `transform` view takes in a function `T -> U` and converts an input range of `T` into an output
range of `U` by calling the function on every element of the input range.

## Example

```cpp
#include <iostream>
#include <vector>
#include <range/v3/view/transform.hpp>
#include <range/v3/range/conversion.hpp>

int main()
{
    std::vector<int> numbers{1, 2, 3};

    auto halved = numbers
        // Divide each integer by 2, converting it into a double
        | ranges::views::transform([](const int& num) {
          return num / 2.0;
        })
        // Convert the new range of doubles into a std::vector
        | ranges::to<std::vector<double>>();

    for (const auto& new_num : halved) {
        std::cout << new_num << std::endl;
    }
}
```

## Syntax

```cpp

auto output_range = input_range | ranges::views::transform(transform_func);

```

## Parameters

- `transform_func`
   - This function maps an input value to an output value.
   - The signature should be equivalent to: `U fun(const T& a);`

- `input_range`
   - The range of input values

- `output_range`
   - The range of output values
   - This range will have the same category as the input range (excluding contiguous ranges). Contiguous ranges are reduced to random access ranges.

example/view/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+set(CMAKE_FOLDER "${CMAKE_FOLDER}/algorithm")
```
ericniebler, line 1: I'm a CMake n00b. What does this do?

johelegp: Appends `/algorithm` to `${CMAKE_FOLDER}`.

Dig-Doug:
Great.

One more question: Should I put the docs with the `transform_fn` (aka. where it is now) or should I create a separate page somewhere else?

ericniebler:
> Should I put the docs with the transform_fn (aka. where it is now) or should I create a separate page somewhere else?

I'm not sure. The auto-generated reference docs are so useless, I'm a bit concerned that if you put it there, nobody will see it. But making the reference docs actually useful is a worthy goal. I would say, put it with `transform_fn` for now.

BRevzin:
I think this is a ton better.

One correction:

> The signature should be equivalent to: `U fun(const T& a);`

This isn't quite right. It doesn't take a `const T&`, what it takes is `range_reference_t<decltype(input_range)>` - which need not be a reference to const. 

I'm not sure how to put all of this together into one piece of documentation, but the relevant pieces of information overall that we need to somehow expose are:

* The input range's reference type (call this `T`) and range category (call this `cat`?)
* The requirements of the function (a type `F` such that `regular_invocable<F&, T>`, and let `U = invoke_result_t<F&, T>`). I've taken to writing this as `T -> U` and I've found that people tend to understand this without problems even if they've never seen Haskell. 
* The output range's characteristics (the category is `min(cat, random_access)`, the reference type is `U`, and the value type is `decay_t<U>`)
* Each element is produced by invoking `f` on the corresponding element of the input range.

Taking something like `filter` in contrast:

* The input range's reference type (call this `T`) and the range category (call this `cat`?)
* The requirements of the function type (which has type `T -> bool`).
* The output range's characteristics (the category is `min(cat, bidirectional)`, and the reference and value type match the input range.
* How we keep/drop the elements based on the unary predicate

I also wonder if we shouldn't have just... code. So like, a coroutine that returns a generator isn't a valid implementation of any of these because that would always be an input_range... but a generator implementation will typically be fairly short and illustrative. Taking again `group_by` as my go-to sufficiently complicated range adapter:

```cpp
template <forward_range R, indirect_binary_predicate<iterator_t<R>, iterator_t<R>> F>
auto group_by(R&& range, F predicate)
    -> generator<subrange<iterator_t<R>>>
{
    iterator_t<R> b = ranges::begin(range);
    sentinel_t<R> e = ranges::end(range);
    if (b == e) {
        co_return;
    }
    for (;;) {
        iterator_t<R> next = ranges::find_if(b, e, [&](auto&& elem){
            return not std::invoke(predicate, *b, elem);
        });
        co_yield subrange(b, next);
        if (next == e) {
            break;
        }
        b = next;
    }
}
```

This isn't how `group_by` is implemented, but I think it does help illustrate what it actually does (notably the predicate is always looking at `*b`, it's not looking at pairwise-consecutive elements). Which should also appear in text form, but both probably help?

Dig-Doug:
Thanks for all of the feedback, I have updated the PR to reflect the changes.

I have also added a test for the example output.

Here's what it looks like:

Screenshot from 2020-08-16 08-37-33

- RE: `U fun(const T& a);`

  You are correct, I've updated the bullet.
  
- RE: Relevant info:
  - Input range type
    - Stated in the top description
  - Range category
    - Stated in the top description
  - Function requirements
    - Added in parameters description
  - Output range characteristics
    - This is stated in the parameters
    - Added the reference and value type to the parameters
  - How elements are produced
    - Stated in the top description
  
- RE: Code snippet
  
  I think that snippet is pretty complicated because it requires a lot of background knowledge to understand (e.g. what is `indirect_binary_predicate<>`/`generator<>`/`sentinel_t<>`?). I do think there's value in having a "possible implementation" section (e.g. [here](https://en.cppreference.com/w/cpp/algorithm/transform)), but I don't think it's required to use the range. I will leave it for someone else to add :)

example/view/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
-set(CMAKE_FOLDER "${CMAKE_FOLDER}/algorithm")
+set(CMAKE_FOLDER "${CMAKE_FOLDER}/view")
 
 rv3_add_test(example.view.transform example.view.transform transform.cpp)
+
+find_program(BASH_PROGRAM bash)
+
+if (BASH_PROGRAM)
+  add_test(example.view.check_example_outputs
+          ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/check_example_outputs.sh ${CMAKE_CURRENT_SOURCE_DIR})
```
dvirtz, line 9: you can use the cross platform `cmake -E compare_files`. See https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-a-command-line-tool

example/view/transform.cpp
```diff
@@ -1,26 +1,21 @@
...
+        // Divide each integer by 2, converting it into a double
+        | ranges::views::transform([](const int& num) {
+          return num / 2.0;
+        });
 
-    for (const auto& new_num : doubled) {
-      std::cout << new_num << std::endl;
+    for (const auto& new_num : halved) {
+        std::cout << new_num << std::endl;
     }
```
dvirtz:
```suggestion
    std::cout << halved << '\n';
```

range-v3 can print views. also [Stop using endl](https://www.youtube.com/watch?v=GMqQOEZYVJQ)

example/view/transform.cpp
```diff
@@ -1,26 +1,21 @@
 //! [transform example]
 #include <iostream>
 #include <vector>
-#undef NDEBUG // Force assertions on
-#include <assert.h>
-
 #include <range/v3/view/transform.hpp>
 #include <range/v3/range/conversion.hpp>
```
dvirtz, line 5: redundant include

include/range/v3/view/transform.hpp
```diff
@@ -517,8 +517,33 @@ namespace ranges
...
+        /// ```cpp
+        /// auto output_range = input_range | ranges::views::transform(transform_func);
+        /// ```
+        ///
+        /// ## Parameters
+        /// <pre><b>transform_func</b></pre>
+        ///   - Maps an input value to an output value (`transform_func(T) -> U`)
+        ///
+        /// <pre><b>input_range</b></pre>
+        ///   - The range of elements to transform
```
dvirtz, line 541: maybe add that Reference type is `T` like you do for `output_range`

Dig-Doug: Done

example/view/transform.cpp
Dig-Doug: Done

example/view/transform.cpp
Dig-Doug: Done

example/view/CMakeLists.txt
Dig-Doug: I updated the code to use compare_files. However, I couldn't figure out how to write the test output file without using `sh`. If `sh` is not found, it should skip the test and the doc build should work fine since it uses the checked in golden.

Let me know if you have any suggestions, I'm not familiar with cmake.

dvirtz: you can use [PASS_REGULAR_EXPRESSION](https://cmake.org/cmake/help/latest/prop_test/PASS_REGULAR_EXPRESSION.html) property and omit the reference output as follows:
```cmake
rv3_add_test(example.view.transform example.view.transform transform.cpp)
set_tests_properties(range.v3.example.view.transform PROPERTIES PASS_REGULAR_EXPRESSION "\\[0.5,1,1.5\\]")
```

Dig-Doug: Done.

example/view/transform_golden.txt
```diff
@@ -0,0 +1 @@
+[0.5,1,1.5]
```
dvirtz, line 1: This is now redundant

Dig-Doug: This file is included as a snippet in the docs: `/// \include example/view/transform_golden.txt`

ericniebler:
This is AWESOME. Is this ready to go, or is more work planned?

Dig-Doug:
This is ready to go, I wasn't planning on adding anything else.

ericniebler:
Thanks for sticking with it. This is a welcome addition.

PR #1547, opened 2020-08-10
fix #1546: Visibility bug in gcc-9 and below using concepts

### LINKED ISSUE #1546: Visibility bug in gcc-9 and below using concepts
The following code has unwanted side effects:

```c++
#include <range/v3/view/split.hpp>

class A {
  int i{5};
};

int main()
{
    return A{}.i; // should be illegal, because it is private
}
```
https://godbolt.org/z/rP1bGK

which makes any private/protected member public.

Expected output:

```
<source>: In function 'int main()':
<source>:9:16: error: 'int A::i' is private within this context
    9 |     return A{}.i; // should be illegal, because it is private
      |                ^
<source>:4:7: note: declared private here
    4 |   int i{5};
      |       ^
```

This is a known gcc-bug which was fixed in gcc-10. We internally have the following concept check to make sure we don't use any concept syntax that triggers this bug:

```c++
template <typename t> concept private_bug = requires(t a) { a.i; };
static_assert(!private_bug<A>, "See https://github.com/seqan/seqan3/issues/1317");
```
(I love this expression, because you can have self-warning code for concepts with concepts xD)

---

I tried to reduce the code and this came out:

```c++
#include <utility>
namespace ranges {
template <typename I1, typename I2 = I1> concept indirectly_swappable = true;
}
namespace ranges {
template <typename Rng>
using iterator_t = decltype(begin(std::declval<Rng &>()));
template <typename V, typename Pattern> struct split_view;
namespace detail {
template <typename JoinView, bool Const> struct split_inner_iterator;
template <typename V, typename Pattern, bool Const>
struct split_inner_iterator<split_view<V, Pattern>, Const> {
  using Base = V;
// <<<<<<<<<<<<<<<<<<<<<<<< misbehaving code
  template <bool = false>
  friend constexpr auto iter_swap(split_inner_iterator const &x,
                                  split_inner_iterator const &y)
      -> std::enable_if_t<indirectly_swappable<iterator_t<Base>> && true> {}
// >>>>>>>>>>>>>>>>>>>>>>>>
};
} // namespace detail
} // namespace ranges
class A {
  int i{5};
};
template <typename t> concept private_bug = requires(t a) { a.i; };
static_assert(!private_bug<A>,  "See https://github.com/seqan/seqan3/issues/1317");
```
https://godbolt.org/z/rP1bGK

Adding parenthesis around the `&&` expression, solves this issue.

```c++
// <<<<<<<<<<<<<<<<<<<<<<<< misbehaving code
  template <bool = false>
  friend constexpr auto iter_swap(split_inner_iterator const &x,
                                  split_inner_iterator const &y)
      -> std::enable_if_t<(indirectly_swappable<iterator_t<Base>> && true)> {}
// >>>>>>>>>>>>>>>>>>>>>>>>
```

marehr:
See #1546 for details.

### DIFF
--- a/include/concepts/concepts.hpp
+++ b/include/concepts/concepts.hpp
@@ -427,7 +427,7 @@
 
 /// INTERNAL ONLY
 #define CPP_BROKEN_FRIEND_RETURN_TYPE_AUX_3_(...)                               \
-    __VA_ARGS__ && CPP_TRUE_FN>>
+    (__VA_ARGS__ && CPP_TRUE_FN)>>
 
 #ifdef CPP_WORKAROUND_MSVC_779763
 #define CPP_broken_friend_member                                                \


ericniebler:
Wow, cool. Thank you.

PR #1549, opened 2020-08-13
Fixed installation on other than Ubuntu distributions

xvitaly:
Fixed installation on other than Ubuntu GNU/Linux distributions.

Fedora and other GNU/Linux distributions use different $libdir prefixes. Now it can be installed on any GNU/Linux distributions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,12 +168,13 @@ write_basic_package_version_file(
 )
 set(CMAKE_SIZEOF_VOID_P ${OLD_CMAKE_SIZEOF_VOID_P})
 
-install(TARGETS range-v3-concepts range-v3-meta range-v3 EXPORT range-v3-targets DESTINATION lib)
-install(EXPORT range-v3-targets FILE range-v3-targets.cmake DESTINATION lib/cmake/range-v3)
+include(GNUInstallDirs)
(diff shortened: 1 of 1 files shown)


johelegp:
I suppose the intention is for the installation to use the system's conventional directories, because it should otherwise work according to the [search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure). Now I know why I should use `GNUInstallDirs` and not just any hard-coded path guaranteed to work.

CMakeLists.txt
```diff
@@ -168,12 +168,13 @@ write_basic_package_version_file(
 )
 set(CMAKE_SIZEOF_VOID_P ${OLD_CMAKE_SIZEOF_VOID_P})
 
-install(TARGETS range-v3-concepts range-v3-meta range-v3 EXPORT range-v3-targets DESTINATION lib)
-install(EXPORT range-v3-targets FILE range-v3-targets.cmake DESTINATION lib/cmake/range-v3)
+include(GNUInstallDirs)
```
ericniebler, line 171: What does this do on a Windows system? Or MacOS?

xvitaly: Now it will work correctly on all *nixes. On Windows it will be ignored like all install() methods.

PR #1554, opened 2020-08-18
Add build2 support

Klaim:
These changes adds [`build2`](https://build2.org) support for:
 - building, testing, working with the project as-is once cloned locally (with one or more configurations at a time);
 - other `build2` projects can use this repository to automatically acquire and depend on the `range-v3` library;
 - packaging, publishing a package by sending it to `cppget.org` (the central `build2` package repository);
 - sending a specific revision to `build2`'s provided CI;
 - installing a specific revision using `build2`;

See modified `README.md` and [./NOTES-build2.md](./NOTES-build2.md) for details.

This is a first step in supporting `build2` though it is not absolutely complete compared to features provided by the current CMakeFiles.txt.

What's in:
 - Build files (`build2file`, equivalent to makefiles) describing how to build the project's targets.
 - Tests and examples [builds and run on `build2`'s CI](https://ci.cppget.org/@b9c91655-6e39-46de-987a-a9777e85a239). They are setup as "sub-project" so they can be tested as seprate packages using `range-v3` by the CI.
 - Project/package files (the most interesting to look at is `./manifest`).
 - The only target exported at the moment is the `lib{range-v3}` library (tell me if you want users to be able to refer to other targets of the project).
 - `./README.md` was updated with minimal info about using this project with `build2`.
 - `./NOTES-build2.md` was added to provide `build2`-specific documentation for handling the project when working with `build2` for people not yet completely familiar with `build2`. We can remove it if it's too much documentation, though I thought it would help the maintainers.

What is not provided yet (should I create tickets for these later?):
 - None of the project-specific configuration options provided in the CMake files are supported yet (because of lack of time). Except for potential bugs, the project should build with `build2` with the default project-specific configuration (no coroutines, etc.).
 - The performance tests (in `./perf`) have build files but are not built without an explicit command: they require `google-benchmark` which I didn't have time to properly try.
 - Tests which depends on coroutines are currently not built, I intend to add them once I add the relative configuration option.
 - `build2`'s CI builds for other compilers than `msvc` on Windows are deactivated/excluded because of either [compiler ICEs](https://bugs.llvm.org/show_bug.cgi?id=47044) or [other issues](https://ci.cppget.org/@67363d29-7540-447d-bb77-bc0ba86f8184?builds).

Because this PR targets the `master` branch, I set the version to ~`0.12.0-a.0.z`~ `0.11.1-a.0.z` which indicates to `build2` that this is a work in a progress, not a fixed version ([see this document for details](https://build2.org/bpkg/doc/build2-package-manager-manual.xhtml#package-version)).
The information for what to do when a release is ready to have a fixed version are in [./NOTES-build2.md](./NOTES-build2.md).

Publishing a `build2` package have to be done from a repository owned by the publisher. So for `range-v3 v0.11.0` we can either:
 - let the maintainers do the publishing or;
 - let me handling it, in which case I'll do it from my fork.

In all cases it will have to be done via a separate (release?) branch based on this PR changes because `v0.11.0` was already published before I added `build2` support, so we would need to do it from a branch starting at tag `0.11.0`.

Note: as an alternative setup, these changes could be moved in a separate `build2`-package repository that would then use `git` submodules to work. It would be less intrusive to do so and I have experience doing it. In any case, my fork (or my package repository) will be passed to [the `build2` community so that we can help each other update this project](https://github.com/build2-packaging/).

Current `build2` CI results: https://ci.cppget.org/@b9c91655-6e39-46de-987a-a9777e85a239


ericniebler:
> Because this PR targets the master branch, I set the version to 0.12.0-a.0.z which indicates to build2 that this is a work in a progress, not a fixed version (see this document for details).

I will probably release a v0.11.1 before v0.12.0. Not sure if you want to change anything in this PR accordingly.

Klaim:
OK then I'll change this PR (ASAP) to `v0.11.1-a.0.z` (meaning "in-progress v0.11.1") and it will have to be changed to just `v0.11.1` later when it's ready for release (ideally the tag match the revision where the version is fixed).

Klaim:
I think it's the same issue when you add multiple new build systems to a project, not just package managers manifests. This is why I mentioned the alternative that we often do: have the build2 project/package in a separate repository (maintained by me and some others in build2 community). I should probably have started this way, anyway.

(Note that I did add a lot so that it is maintainable by you in case I can't be around, it's why I took the time to redact ./NOTES-build2.md - it's not lost though, I can reuse that file I think).

Maintenance of packages/versions is easy for people used to build2 (because there are commands handling them for us) but it would cost you some time to learn it indeed, and I agree it doesnt help for example that CMake have a different source of version number.
Going with a separate package seems the the best way to go to make it happen fast. In that case, I will maintain that package and report to you if I find issues that are not visible on your current CI (like the Windows+clang++ issues that cannot be reproduced with CMake because it does not provide that configuration).

(see https://github.com/build2-packaging for multiple examples of such separate pacakge repository)

If we agree, I'll change ASAP this PR or make another one that document to build2 users to use that other repository, for users it will be almost indistinct from having the code here - the PR would then be only a short section in the ./README.md file.
Then I'll publish a package for at least v0.11.0 on https://cppget.org, and future versions.
Then if you need help with the issues I pointed only visible on Windows+clang++/gcc I can probably help in some ways.

Klaim:
I'll make another PR with the separate repository for build2 packaging. I already started it, it's almost finished.

PR #1561, opened 2020-08-25
Avoid including empty <span> header when using MSVC and C++17

facutuesca:
While compiling the latest `master` with MSVC and C++17 the same warning shows over and over:
```
...
[build] [14/483   0% :: 4.966] Building CXX object test\algorithm\CMakeFiles\range.v3.alg.for_each_n.dir\for_each_n.cpp.obj
[build] The contents of <span> are available only with C++20 or later.
[build] [15/483   0% :: 5.069] Building CXX object test\algorithm\CMakeFiles\range.v3.alg.generate.dir\generate.cpp.obj
[build] The contents of <span> are available only with C++20 or later.
[build] [16/483   0% :: 5.826] Building CXX object test\algorithm\CMakeFiles\range.v3.alg.make_heap.dir\make_heap.cpp.obj
[build] The contents of <span> are available only with C++20 or later.
[build] [17/483   0% :: 6.051] Building CXX object test\algorithm\CMakeFiles\range.v3.alg.max_element.dir\max_element.cpp.obj
[build] The contents of <span> are available only with C++20 or later.
...
```

This is due to the `<span>` header being available to include (but not to use) with C++17:
```c++
// MSVC's <span> header
// (...)
#if !_HAS_CXX20
#pragma message("The contents of <span> are available only with C++20 or later.")
#else // ^^^ !_HAS_CXX20 / _HAS_CXX20 vvv
#include <cstddef>
// (...)
```

The current implementation guards against the use of `std::span` by checking the `__cpp_lib_span` feature-test macro, but `#include <span>` is only guarded by `__has_include()`

This PR adds a workaround macro so that we don't include `<span>` when compiling with MSVC and C++17 or lower

### DIFF
--- a/include/range/v3/detail/config.hpp
+++ b/include/range/v3/detail/config.hpp
@@ -280,6 +280,11 @@ namespace ranges
 #define RANGES_WORKAROUND_MSVC_OLD_LAMBDA
 #endif
 
+#if _MSVC_LANG <= 201703L
+#define RANGES_WORKAROUND_MSVC_UNUSABLE_SPAN // MSVC provides a <span> header that is
+                                             // guarded against use with std <= 17
+#endif
+
 #elif defined(__GNUC__) || defined(__clang__)
 #define RANGES_PRAGMA(X) _Pragma(#X)
 #define RANGES_DIAGNOSTIC_PUSH RANGES_PRAGMA(GCC diagnostic push)
--- a/include/range/v3/range/access.hpp
+++ b/include/range/v3/range/access.hpp
@@ -14,6 +14,8 @@
 #ifndef RANGES_V3_RANGE_ACCESS_HPP
 #define RANGES_V3_RANGE_ACCESS_HPP
 
+#include <range/v3/detail/config.hpp>
+
 #include <functional> // for reference_wrapper (whose use with begin/end is deprecated)
 #include <initializer_list>
 #include <iterator>
@@ -21,7 +23,7 @@
 #include <utility>
 
 #ifdef __has_include
-#if __has_include(<span>)
+#if __has_include(<span>) && !defined(RANGES_WORKAROUND_MSVC_UNUSABLE_SPAN)
 #include <span>
 #endif
 #if __has_include(<string_view>)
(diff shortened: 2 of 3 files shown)


include/range/v3/detail/config.hpp
```diff
@@ -757,4 +757,10 @@ namespace ranges
 #endif
 #endif
 
+#if __has_include(<span>) && defined(__cpp_lib_span) && __cpp_lib_span > 0
```
ericniebler, line 760: `__has_include` is a C++17 feature. Range-v3 supports C++14. Also, where does `__cpp_lib_span` get defined?

facutuesca: Ah my bad. I see it's being used in other parts of the file, but conditional on its existence. Would this work?
```c++
#if defined(__has_include) && __has_include(<span>) && defined(__cpp_lib_span) && __cpp_lib_span > 0
#define RANGES_CXX_SPAN 1
#else
#define RANGES_CXX_SPAN 0
#endif
```

To the other question, `__cpp_lib_span` [should](https://en.cppreference.com/w/cpp/utility/feature_test) be defined when `<version>` is included: 
https://github.com/ericniebler/range-v3/blob/ffa477cb30e964d8578cb77ac8bc45f9743a94b0/include/range/v3/detail/config.hpp#L17-L24

include/range/v3/detail/config.hpp
```diff
@@ -757,4 +757,11 @@ namespace ranges
 #endif
 #endif
 
+#if defined(__has_include) && __has_include(<span>) && \
+    defined(__cpp_lib_span) && __cpp_lib_span > 0
```
ericniebler, line 761: Not all stdlib implementations have a `<version>` header. Are there any stdlib implementations have a usable `<span>` header but not a `<version>` header (plausable), or that have a `<version>` header but fail to define `__cpp_lib_span` in it (unlikely)? If so, this risks concluding that there is no usable `<span>` header when in fact there might be.

facutuesca: This is a good point. The problematic case is, as you said, a compiler that has a usable `<span>` header (which defines `__cpp_lib_span`) but no `<version>` header. Which means the only way to know it's usable would be to include it.

I think a good approach would be to blacklist the known problematic case. In this case: a workaround macro that is defined when compiling with MSVC and C++17 (or lower), and then avoid including `<span>` if the macro it's defined.

I have force-pushed a new implementation based on that, and edited the PR description to match

PR #1564, opened 2020-08-31
Change span to use std::size_t

### LINKED ISSUE #1563: std::span and views::span use different index type
Hello,

Thanks a lot for the work done on this project!

I noticed that [`std::span`](https://en.cppreference.com/w/cpp/container/span) uses `std::size_t` as its extent/`size_type` while ranges uses `std::ptrdiff_t`
https://github.com/ericniebler/range-v3/blob/167d94b2f3de8a7d279a47da55b0cad536665a15/include/range/v3/view/span.hpp#L42
https://github.com/ericniebler/range-v3/blob/167d94b2f3de8a7d279a47da55b0cad536665a15/include/range/v3/view/span.hpp#L151
Additionally, the dependent names are not the same `size_type` vs `index_type`.

Should range-v3 aligns itself on the standard?

I'm happy to send a PR if that makes sense to you.

AntoinePrv:
These changes are breaking `span` API

Close #1563

### DIFF
--- a/include/range/v3/view/span.hpp
+++ b/include/range/v3/view/span.hpp
@@ -39,11 +39,11 @@ namespace ranges
     /// \cond
     namespace detail
     {
-        using span_index_t = std::size_t;
+        using span_size_t = std::size_t;
     } // namespace detail
     /// \endcond
 
-    constexpr detail::span_index_t dynamic_extent = SIZE_MAX;
+    constexpr detail::span_size_t dynamic_extent = SIZE_MAX;
 
     /// \cond
     namespace detail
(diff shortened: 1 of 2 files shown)


AntoinePrv:
@ericniebler I'm not sure about what is the expected behaviour in the failing test.

What is expected when converting a `span<int>` to a `span<int, 1>`? Should it not be forbidden? What if the `span<int>` had zero element?
I'm under the impression that this may have not been intended, relying on the fact that `dynamic_extend` was `-1` (now `SIZE_MAX`) and so the cardinality was detected smaller.

PR #1566, opened 2020-09-08
[Docs] views::filter and views::ints usage examples

Dig-Doug:
Similar to #1536.


include/range/v3/view/filter.hpp
```diff
@@ -97,6 +97,35 @@ namespace ranges
...
+        /// <pre><b>filter_func</b></pre>
+        ///   - Called once for each element of the input range
+        ///   - Returns true for elements that should present in the output range
+        ///
+        /// <pre><b>input_range</b></pre>
+        ///   - The range of elements to filter
+        ///   - Reference type: `T`
+        ///
+        /// <pre><b>output_range</b></pre>
+        ///   - The range of filtered values
```
ericniebler, line 126: Might be nice to document the the concept satisfied by the resulting range. It is the stronger of either `forward_range` or the concept satisfied by the input. Also, the result doesn't satisfy `sized_range`, and the result is a `common_range` if and only if the input is a `common_range`. The result ranges also does not satisfy `borrowed_range`. I think I got all that right, but it's best to double-check.

include/range/v3/view/iota.hpp
```diff
@@ -523,6 +523,34 @@ namespace ranges
...
+        ///
+        /// <pre><b>upper_bound</b></pre>
+        ///   - Exclusive upper bound
+        ///   - Required when `lower_bound` is specified
+        ///     - To create an infinite range with a `lower_bound`, use
+        ///     `ranges::unreachable` as the `upper_bound`
+        ///
+        /// <pre><b>output_range</b></pre>
+        ///   - Range of monotonically increasing ints
+        ///   - When an `upper_bound` is not specified, the range is quasi-infinite
```
ericniebler, line 552: Is "quasi-infinite" defined anywhere? What does it mean? What concepts does the resulting range satisfy?

Dig-Doug: I based this off of the existing docs "Generate a range of monotonically increasing ints. When used without arguments, it generates the quasi-infinite range". If I'm reading the code correctly, it looks like the iterator will eventually overflow.

Not sure what concepts it satisfies or how to find out. Can you point me to that info?

include/range/v3/view/filter.hpp
Dig-Doug: Added that info.

include/range/v3/view/iota.hpp
ericniebler: All "infinite" ranges have difference types that will eventually overflow. It's a problem with the ranges design I was never able to properly solve.

PR #1573, opened 2020-10-23
zip_with on no ranges should derive its value_type from the callable.

BRevzin:
Currently, `zip_with(f)` is hardcoded to yield `empty_view<tuple<>>` but it should really be based on the callable itself. This fixes `zip_with([]{ return 42; })` to provide `empty_view<int>` instead.

### DIFF
--- a/include/range/v3/view/zip_with.hpp
+++ b/include/range/v3/view/zip_with.hpp
@@ -410,8 +410,9 @@ namespace ranges
 
             template(typename Fun)(
                 /// \pre
-                requires zippable_with<Fun>)
-            constexpr empty_view<std::tuple<>> operator()(Fun) const noexcept
+                requires zippable_with<
+                    Fun>) constexpr empty_view<detail::decay_t<invoke_result_t<Fun &>>>
+            operator()(Fun) const noexcept
             {
                 return {};
             }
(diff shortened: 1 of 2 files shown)


BRevzin:
The CI failures seem unrelated to this change?

ericniebler:
> The CI failures seem unrelated to this change?

Uh, that's weird. Something changed on the CI machines? The tests were passing before.

include/range/v3/view/zip_with.hpp
```diff
@@ -410,8 +410,9 @@ namespace ranges
 
             template(typename Fun)(
                 /// \pre
-                requires zippable_with<Fun>)
-            constexpr empty_view<std::tuple<>> operator()(Fun) const noexcept
+                requires zippable_with<
+                    Fun>) constexpr empty_view<detail::decay_t<invoke_result_t<Fun &>>>
```
ericniebler, line 414: Formatting here, pls. If you're running this through `clang-format` you might need something like this:

```suggestion
                requires zippable_with<Fun>) //
            constexpr empty_view<detail::decay_t<invoke_result_t<Fun &>>>
```

include/range/v3/view/zip_with.hpp
```diff
@@ -437,8 +438,9 @@ namespace ranges
 
             template(typename Fun)(
                 /// \pre
-                requires copy_constructible<Fun> AND invocable<Fun &>)
-            constexpr empty_view<std::tuple<>> operator()(Fun) const noexcept
+                requires copy_constructible<Fun> AND invocable<
+                    Fun &>) constexpr empty_view<detail::decay_t<invoke_result_t<Fun &>>>
```
ericniebler, line 442: Here as well.

ericniebler (changes requested):

include/range/v3/view/zip_with.hpp
BRevzin: I don't understand why people use clang-format, exhibit N.

ericniebler: To be fair, I'm really making its job unreasonably difficult with all these macros.

PR #1577, opened 2020-11-04
Remove references to Valgrind as it is not used anymore

furkanusta:
Workarounds for the Valgrind seems to be introduced in eacc71a as a workaround for Travis bugs, but
Travis CI is not being used anymore (See Pull #1492)

### DIFF
--- a/cmake/ranges_env.cmake
+++ b/cmake/ranges_env.cmake
@@ -79,18 +79,5 @@ else()
   set(RANGES_RELEASE_BUILD TRUE)
 endif()
 
-# Find Valgrind
-find_program(MEMORYCHECK_COMMAND valgrind)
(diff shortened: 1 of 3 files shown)


ericniebler:
It's true that I don't use valgrind in CI anymore, but I expect others who use range-v3 may also wish to use valgrind. I don't want to remove the workarounds from the code and cause spurious failures for them. Could you leave in the workarounds? I am OK with removing the cmake stuff.

furkanusta:
Yeah, that makes sense.
I've reverted the code changes but they are appearing as two separate commits as now. Hope that's okay

ericniebler:
Thanks. I've squashed the commits and merged.

PR #1586, opened 2020-12-04
adds `ranges::unformatted_ostream_iterator`

cjdb:
`std::ostream_iterator` is only useful if the output needs to be
formatted and `std::ostreambuf_iterator` only works with types that
have a `std::char_traits` specialisation.

The purpose of `unformatted_ostream_iterator` is to write to an ostream
object without formatting (so integers will be written as binary rather
than as text).


cjdb:
GH Actions failure related to #1587.

ericniebler:
Cool, thanks!

Issue #1587, opened 2020-12-04
Coroutines tests break CI


cjdb:
Coroutines tests are failing when building with Clang because libstdc++ headers want a `-fcoroutines` option enabled, which Clang doesn't have (it's got `-fcoroutines-ts`). There are two solutions, and I think they should happen in parallel.

1. Make all Clang-based coro-aware CI use libc++, which doesn't have this issue. Simple and can be done internally, but adds to the CI load.
2. File a bug with libstdc++ to check for Clang flag as well. Ultimately transparent to range-v3 and improves the dev ecosystem, but it could take time (will it be a GCC 10 hotfix? Even if it is, will GitHub's GCC 10 get this hotfix?).

PR #1588, opened 2020-12-04
adds `noexcept` to coro detection `final_suspend`

cjdb:
The CMake coroutine-detection test fails for recent versions of Clang
because `present::promis_type::final_suspend` wasn't `noexcept`.


cjdb:
Actions failures related to #1587.

PR #1593, opened 2020-12-29
adds `ranges::basic_unformatted_istream_view`

cjdb:
`istream_view` is only useful if the input to be formatted on read
and `std::istreambuf_iterator` only works with types that have a
`std::char_traits` specialisation.

The purpose of `unformatted_istream_view` is to read from an istream
object without formatting (so integers will be read as binary rather
than as text).


cjdb:
A brief look at the failed builds suggests this is related to #1587.

Issue #1598, opened 2021-01-09
drop_last - source range with fewer elements than to be dropped


szilardszaloki:
Shouldn't `drop_last` produce an empty range if the source range has fewer elements than to be dropped?
(_"Given a source range and an integral count, return a range consisting of all but the last count elements from the source range, or an empty range if it has fewer elements."_ - https://ericniebler.github.io/range-v3/index.html#autotoc_md8)
```
for (std::vector one{ 1 }; auto i : one | ranges::views::drop_last(2)) {
    std::cout << i;
}
```
drop_last.hpp(53): `RANGES_EXPECT(initial_size >= n);`
https://godbolt.org/z/81Kvs8

JohelEGP:
@ericniebler This can be closed.

PR #1599, opened 2021-01-10
Removing bogus assert in drop_last.

BRevzin:
The documentation states:

> Given a source range and an integral count, return a range consisting of all but the last count elements from the source range, or an empty range if it has fewer elements. 

But there's currently an assert that prevents the empty range case from working.


johelegp:
Resolves #1598.

ericniebler:
Thanks!

Issue #1603, opened 2021-01-27
Undefined behavior caught by GCC 9 on Ubuntu 20.04 when using `views::cache1`


Marc--Olivier:
Hi there,

I recently upgraded my OS from Ubuntu 18.04 to Ubuntu 20.04 and since, my unit tests that were using `views::cache1` started to fail. The code repeatedly returns the error `../deps/range-v3/include/range/v3/view/cache1.hpp:35:12: runtime error: load of value 255, which is not a valid value for type 'bool'`, where the value `255` varies.

I would be very grateful if someone could have a look at this issue...

Below is a shell script that should help to reproduce the error:

```sh
export CC=gcc-9
export CXX=g++-9
RANGE_V3_BRANCH=0.10.0

tmp_dir="$(mktemp -d)"
test_project_dir="$tmp_dir/test_project"
mkdir -p "$test_project_dir"/build

## Compile range-v3 library
pushd "$tmp_dir"
git clone https://github.com/ericniebler/range-v3.git --branch "$RANGE_V3_BRANCH" --depth 1
mkdir range-v3/build
pushd range-v3/build
cmake .. -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DRANGE_V3_TESTS=OFF -DRANGE_V3_HEADER_CHECKS=OFF \
  -DRANGE_V3_DOCS=OFF -DRANGE_V3_EXAMPLES=OFF \
  -DCMAKE_INSTALL_PREFIX="$test_project_dir/deps/range-v3"
ninja && ninja install
popd

popd
pushd "$test_project_dir"/build

## Create CMake file and test C++ file.
cat <<EOF >"$test_project_dir"/CMakeLists.txt
cmake_minimum_required(VERSION 3.19)
project(test_cache1)

set(CMAKE_CXX_STANDARD 17)

add_compile_options(-fno-sanitize-recover=all -fsanitize=undefined)

find_package(range-v3 REQUIRED HINTS "deps/range-v3")

add_executable(test_cache1 test_cache1.cpp)

target_link_libraries(test_cache1
  PRIVATE
    range-v3::range-v3
    -fsanitize=undefined
)
EOF

cat <<EOF >"$test_project_dir"/test_cache1.cpp
#include <range/v3/view.hpp>
#include <iostream>

using namespace ranges;

int main(int argc, char* const argv[]) {
  auto v2 = views::iota(1,argc+9)
    | views::transform([](int i) { return i+1; })
[...]

PR #1610, opened 2021-03-03
Fix read of uninitialized bool in cache1

jonathanhaigh:
For #1603: Undefined behavior caught by GCC 9 on Ubuntu 20.04 when using `views::cache1`

Ensure that cache1's dirty_ member is initialized to avoid UB found by
UBSan:
range-v3-0.11.0/include/range/v3/view/cache1.hpp:35:12: runtime error: load of value 39, which is not a valid value for type 'bool'


ericniebler:
Thanks!

PR #1622, opened 2021-04-22
CMake: Fix find_package and_add subdirectory. Fixes #1621

### LINKED ISSUE #1621: CMake find_package and add_subdirectory
Hi. It seems impossible to use add_subdirectory(range-v3) and find_package(range-v3) in one project. In my case there is a parent project, that contains
```
add_subdirectory(range-v3) 
add_subdirectory(<lib that uses find_package(range-v3)>) 
```
cmake exits with error:
```
CMake Error at range-v3/cmake/range-v3-config.cmake:1 (include):
  include could not find requested file:

    C:/projects/testCmake/range-v3/cmake/range-v3-targets.cmake
```
The minmal case is project that contains clone of range-v3 and CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.5)
project(testCmake)

add_subdirectory(range-v3)
find_package(range-v3 REQUIRED)
```

I think it should work and there is a commit that says that fixes this behaviour: [8473ac782a3501a399a83c3517505b1bfd2627f9](https://github.com/ericniebler/range-v3/commit/8473ac782a3501a399a83c3517505b1bfd2627f9)
But it does not fix anything. `if (TARGET range-v3::meta)` should be placed before include because range-v3-targets.cmake is generated by cmake install and there is no such file in add_subdirectory mode.
 
The only thing we need to fix is moving one line in cmake/range-v3-config.cmake
```
if (TARGET range-v3::meta)
  return()
endif()

include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets.cmake")
....
```
I can make PR soon if it helps.

SmorkalovG:
Full description in issue #1621. CMakeLists.txt to reproduse bug and check fix is:
```
cmake_minimum_required(VERSION 3.5)
project(testCmake)

add_subdirectory(range-v3)

list(APPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_SOURCE_DIR}/range-v3)
find_package(range-v3 REQUIRED)

add_executable(${PROJECT_NAME} "main.cc")
target_link_libraries(${PROJECT_NAME} range-v3::range-v3)

target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_14)

```

### DIFF
--- a/cmake/range-v3-config.cmake
+++ b/cmake/range-v3-config.cmake
@@ -1,8 +1,8 @@
-include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets.cmake")
-
 if (TARGET range-v3::meta)
   return()
 endif()
+
+include("${CMAKE_CURRENT_LIST_DIR}/range-v3-targets.cmake")
 
 add_library(range-v3::meta INTERFACE IMPORTED)
 add_library(range-v3::concepts INTERFACE IMPORTED)


ericniebler:
Thanks!

PR #1625, opened 2021-05-13
accommodate libgcc-9's unusual value for the __cpp_lib_string_view feature test macro - part 2

marehr:
Follow up of https://github.com/ericniebler/range-v3/pull/1609


ericniebler:
Thanks!

PR #1628, opened 2021-05-23
Fold algos

BRevzin:
This is my current intended direction for what to propose in [P2322](https://wg21.link/p2322) (the latest draft still uses `fold_left` and `fold_right`, but the next one will propose what I'm doing in this PR). There's a few more fold algorithms I intend to add to range-v3 but not std (like a short-circuiting fold and a fold that returns the iterator that it leaves off at), but the combinatorial explosion is a bit much, so I thought I'd start with this.

* `foldl` is like `accumulate` except it doesn't take a default operation and can return a type other than the initial value (so that `foldl([0.25, 0.75], 1, plus()) == 2.0`).
* `foldl1` doesn't take an initial value, and instead returns an `optional<U>`. For instance, `min(r)` can be implemented as `*foldl1(r, min)`.
* `foldr` and `foldr1` are the right-fold counterparts. 

I like going with Haskell's names here because they're short, symmetric, and unambiguous. Some languages went the route of using `fold` (for the algorithm that takes an initial value) and `reduce` (for the algorithm that doesn't), Rust being the latest addition to that unfortunate club, and we can't do that even if we wanted to.


BRevzin:
retest this please

PR #1629, opened 2021-05-25
Ensure unstable_remove_if calls the predicate exactly once per element

mjbeardsley:
There are cases where unstable_remove_if can call the predicate on the converged-upon "partition element" twice:
https://godbolt.org/z/a8W5oWr3T

It should have the same guarantee as remove_if:

https://en.cppreference.com/w/cpp/algorithm/remove
> Complexity
> Exactly std::distance(first, last) applications of the predicate.

https://en.cppreference.com/w/cpp/algorithm/ranges/remove
> Complexity
> Exactly N applications of the corresponding predicate and any projection, where N = (last - first), and N-1 move operations at worst.


ericniebler:
Wow, nice find!

PR #1630, opened 2021-05-27
Fixing OS in ci.yml to find older compilers.

BRevzin:


ericniebler:
Range-v3 badly needs a more active maintainer than I can be at the moment. Thank you for stepping in.

BRevzin:
Alright, so I removed the clang 5,6,8,9 C++20 tests (since clang 5,6 C++20 is... optimistic, clang 8 doesn't have `std::is_constant_evaluated` so can't compile with whatever gcc toolchain, and clang 9 doesn't properly handle `<=>` and has issues there). I compensated for this by adding gcc 10 C++20 tests. 

But finally, everything passes 💪

Issue #1633, opened 2021-06-07
`std/detail/associated_types.hpp` will break with recent libc++


ldionne:
Those lines won't work anymore with recent versions of libc++:

https://github.com/ericniebler/range-v3/blob/e157199166e20f7f4a730292e212239682a89095/include/std/detail/associated_types.hpp#L267-L272

I haven't done a formal bisect, but I think this is caused by

```
commit 9f01ac3b3257ab925a2b1229dba19e3eb86a706b
Author: zoecarver <z.zoelec2@gmail.com>
Date:   Tue Apr 20 08:50:11 2021 -0400

    [libcxx] makes `iterator_traits` C++20-aware

    * adds `iterator_traits` specialisation that supports all expected
      member aliases except for `pointer`
    * adds `iterator_traits` specialisations for iterators that meet the
      legacy iterator requirements but might lack multiple member aliases
    * makes pointer `iterator_traits` specialisation require objects

    Depends on D99854.

    Differential Revision: https://reviews.llvm.org/D99855
```

That change was made as part of our ongoing implementation of Ranges in libc++.

It looks like you are trying to detect whether a specialization of `std::iterator_traits` refers to the base template or to a full/partial specialization. If so, you could use the presence of the `__primary_template` member inside `std::iterator_traits` to detect that in newer versions. Even better would be to not rely on any implementation detail of libc++, however I expect you'll say that you can't live without it.

cc @zoecarver for awareness

ldionne:
Godbolt reproducer: https://godbolt.org/z/3fdqsY7Yj

I'm about to submit a PR with a fix for this.

ldionne:
Fixed by PR #1635.

piotrrak:
`-DRANGES_DEEP_STL_INTEGRATION` option requires similar fix #1666

PR #1635, opened 2021-06-15
Work around broken integration with latest libc++.

ldionne:
In newer versions of libc++, the base template of `std::iterator_traits` provides a member typedef called `__primary_template` which is an alias to the `std::iterator_traits` specialization itself.

The fix works as follows:
1. In older versions of libc++, `__primary_template` does not exist, so that overload is never taken and we always use the test for `std::__iterator_traits<I, B>` that we used to have unconditionally.

2. In any version of libc++ where `Traits::__primary_template` exists, it will be a better match than the `__iterator_traits<T, B>` overload, because one is a direct match and the other one requires a conversion to a base.

3. In any version where `std::__iterator_traits` takes only one template parameter (instead of two), the first overload will be discarded in a SFINAE-friendly manner instead of creating a hard-error (as previously).

Fixes issue #1633.


ldionne:
I tried fixing MSVC's behavior for the `std::iterator_traits<T*>` specialization as a fly-by fix.

ericniebler:
Thanks, Louis!

Issue #1645, opened 2021-08-12
Fail with std::back_inserter


sergegers:
After upgrading Visual Studio to 2022 Preview 3 the following code is failed

```

#include <vector>
#include <range/v3/algorithm/copy.hpp>

int main()
{
    std::vector<int> from_vector(10);
    std::vector<int> to_vector;

    ranges::copy(from_vector, std::back_inserter(to_vector));
}
```
As I can see the compilation is failed because std::back_insert_iterator<std::vector<int>> doesn't satisfy concept weakly_incrementable -> semiregular -> default_constructible. But back_insert_iterator is not default constructible! How did it work before? Anyway it looks like a bug. 
The code with STL range implementation works fine
```

#include <vector>
#include <algorithm>
#include <iterator>

int main()
{
    std::vector<int> from_vector(10);
    std::vector<int> to_vector;

    std::ranges::copy(from_vector, std::back_inserter(to_vector));
}
```
because weakly_incrementable implementation is look like 
```
template <class _Ty>
concept weakly_incrementable = movable<_Ty> && requires(_Ty __i) {
    typename iter_difference_t<_Ty>;
    requires _Signed_integer_like<iter_difference_t<_Ty>>;
    { ++__i } -> same_as<_Ty&>;
    __i++;
};

```
and doesn't require default construction.

sergegers:
@ericniebler pls, leave a comment. Did I make a mistake in my conclusions?

ericniebler:
You're correct. The solution is to use the back_inserter from range-v3, and to wait for Microsoft to update its back_inserter implementation for C++20.

JohelEGP:
Range-v3 hasn't adopted https://github.com/cplusplus/draft/commit/479c6a8f333fa8ea6fac11563d715f0cba376d82#diff-0043c904d3e034a77b44feeddf18719e63e646d55db1c564f2a5defe7330c9eaL1321 nor its previous relaxation, as seen at https://github.com/ericniebler/range-v3/blob/master/include/range/v3/iterator/concepts.hpp#L252. My guess is that some wrongly hard-coded logic made Range-v3 fall back to using its own concepts due to not detecting that your implementation still provides them.

sergegers:
So you are waiting when [P2325R3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2325r3.html) leaks in production compiler. Thank you @ericniebler , @JohelEGP .

huixie90:
[...]

Issue #1585, opened 2020-12-03
Range Output


sergegers:
I'm porting a project from boost.ranges v2 to ranges v3 and I 've got a custom stream output for ranges already.  As I know the range printing isn't a part of the standard. Is it possible to disable default range output by conditional compilation or offer the extension machinery for custom output?
Thanks, Serge.

marehr:
https://github.com/ericniebler/range-v3/blob/fa00307bebb503d58107c6abab3bea4ac4930177/include/range/v3/view/interface.hpp#L508-L535

I guess this code conflicts with yours.

sergegers:
Yes, it is. And, due to the output isn't a part of the standard I think to have a way to disable it is a good thing. The similar problem was in ranges v2 but the library looks abandoned and I patched the code myself.

PR #1652, opened 2021-08-31
Fixing Issues #1585 & #1645

sergegers:
There are two fixes in the pull request:
1. Optionally suppress range stream printing #1585 
2. Relax weakly_incrementable concept #1645


ericniebler:
Thank you!

PR #1655, opened 2021-09-13
Joining all views of ranges.

BRevzin:
This implements [P2328](https://wg21.link/p2328) in range-v3, so that users can just write `transform(f) | join` (or `for_each(f)`) without having to deal with `cache1`.

### DIFF
--- a/include/range/v3/utility/optional.hpp
+++ b/include/range/v3/utility/optional.hpp
@@ -978,6 +978,11 @@ namespace ranges
                 return *this;
             }
             using optional<T>::operator=;
+            template<class I>
+            constexpr T & emplace_deref(const I & i)
+            {
+                return optional<T>::emplace(*i);
+            }
         };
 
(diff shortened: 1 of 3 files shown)


include/range/v3/utility/optional.hpp
```diff
@@ -978,6 +978,11 @@ namespace ranges
                 return *this;
             }
             using optional<T>::operator=;
+            template<class I>
+            constexpr T & emplace_deref(const I & i)
+            {
+                return optional<T>::emplace(*i);
+            }
```
JohelEGP, line 985: This won't work if `T` is not copyable.

BRevzin: Fixed. At least for C++17.

ericniebler:
You know, if you were a range-v3 maintainer, you could just make these edits and merge them yourself. 😀

PR #1660, opened 2021-09-20
common_iterator: expose iterator_traits aliases

theodelrieu:
Boost.Move own `iterator_traits` fails to compile when using `ranges::to<flat_map>`, as it relies on the iterator's member types.


brevzin:
Can you (a) provide a full example that fails with (b) a standalone test case that demonstrates the current failure that would pass with these changes? [Ion said](https://github.com/boostorg/move/pull/43#issuecomment-1001211585) he couldn't reproduce it, and it's not clear to me what the issue is.

PR #1661, opened 2021-09-22
functional: fix ranges::on

### LINKED ISSUE #1656: Surprising behaviour of `ranges::on`
The current behaviour of the `ranges::on` is
```
ranges::on(f, g)(x...) === f(g(x...))
```
This is inconsistent with Haskell `Data.Function.on` and `boost::hana::on`

In Boost.Hana,
```
boost::hana::on(f, g)(x...) === f(g(x)...)
```
https://www.boost.org/doc/libs/1_61_0/libs/hana/doc/html/group__group-functional.html#ga35c4fc3c5677b9f558150b90e74d3ab1

And  Haskell behaves similarly to Boost.Hana
```
((+) `on` f) x y = f x + f y
```
https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Function.html#v:on

I think this is a bug and we should fix it directly instead of making a new one that behaves similarly to Haskell one.

theodelrieu:
Fixes #1656

### DIFF
--- a/test/functional/on.cpp
+++ b/test/functional/on.cpp
@@ -1,0 +1,26 @@
+// Range v3 library
+//
+//  Copyright Eric Niebler 2020
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/ericniebler/range-v3
+
+#include <range/v3/functional/on.hpp>
+#include <range/v3/functional/concepts.hpp>
+#include "../simple_test.hpp"
+
+using namespace ranges;
+
+int square(int i) { return i * i; }
+
+int main()
+{
+    auto fn = on(std::multiplies<>{}, square);
+    CHECK(fn(2, 4) == 64);
+
+    return ::test_result();
+}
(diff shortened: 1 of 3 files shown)


test/functional/on.cpp
```diff
@@ -0,0 +1,26 @@
...
+//
+// Project home: https://github.com/ericniebler/range-v3
+
+#include <range/v3/functional/on.hpp>
+#include <range/v3/functional/concepts.hpp>
+#include "../simple_test.hpp"
+
+using namespace ranges;
+
+int square(int i) { return i * 2; }
```
huixie90, line 18: is it intentional to call "times two" "square"?

BRevzin, line 18: Can you either make this function actually square its argument or rename it to indicate that it's doubling?

BRevzin (changes requested):
I know this library isn't big on comments or documentation, but it would help if we added a Boost.HOF-style comment in the header somewhere. Like:

```cpp
assert(on(f, p)(xs...) == f(p(xs)...));
```

test/functional/on.cpp
BRevzin: Beat me too it! I was staring at this test forever thinking how does `4` times `16` equal `32`...

theodelrieu: Yeah I messed up 😂

PR #1668, opened 2021-10-19
Updated fold to P2322R5 (no projections, rename)

BRevzin:
Per LEWG telecon, renaming `foldl`/`foldr` to `fold_left`/`fold_right` and the `*1` suffix to `_first` for the left fold and `_last` for the right fold. Names not finalized yet, but seem more likely to be the final names than `foldl`/`foldl1`/`foldr`/`foldr1`.

Also removing the projections. First, they were implemented incorrectly anyway. Second, for the variants with no initial value, the projection doesn't really work. This is explained in P2322R5 (see [here](https://isocpp.org/files/papers/P2322R5.html#no-projections) until that one is published).

### DIFF
--- a/include/range/v3/algorithm.hpp
+++ b/include/range/v3/algorithm.hpp
@@ -36,6 +36,7 @@
 #include <range/v3/algorithm/find_first_of.hpp>
 #include <range/v3/algorithm/find_if.hpp>
 #include <range/v3/algorithm/find_if_not.hpp>
+#include <range/v3/algorithm/fold.hpp>
 #include <range/v3/algorithm/for_each.hpp>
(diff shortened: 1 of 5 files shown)


cjdb:
Can you please add `#include <range/v3/algorithm/fold.hpp>` to `<range/v3/algorithm.hpp>`? I can close #1682 after that.

BRevzin:
> Can you please add `#include <range/v3/algorithm/fold.hpp>` to `<range/v3/algorithm.hpp>`? I can close #1682 after that.

Sure.

CaseyCarter (approved):

CaseyCarter:
Merging despite unrelated CI failures.

PR #1675, opened 2021-11-08
adds `ranges::contains_subrange`

cjdb:
P2302 proposes an `find`-based contains (called `contains`) and a
`search`-based contains (called `contains_subrange`). The former is used
to determine if an element exists in a range, while the latter is used
to determine whether or not a needle range is a subrange of a
haystack range.


JeffGarland:
Given that LWG has adopted this paper for C++23 it would be nice if this could be added.

brevzin (approved):

test/algorithm/contains_subrange.cpp
```diff
@@ -0,0 +1,95 @@
...
+    ++comparison_count;
+    return a == b;
+}
+
+int main()
+{
+    using namespace ranges;
+    auto full_range = std::array<int, 10>{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+    auto valid_subrange = std::array<int, 5>{2, 3, 4, 5, 6};
+    auto invalid_subrange = std::array<int, 5>{3, 4, 5, 6, 2};
```
brevzin, line 38: With MSVC, `std::array<int, N>::iterator` isn't `int*`, so this test is failing. Can you change these to an `int[N]` instead? The clang 5 build is failing for reasons that are also basically related to this.

Issue #1672, opened 2021-11-04
concepts.hpp:455:5: error: satisfaction of atomic constraint ‘requires(T& t, T& u) {(concepts::<unnamed>::swap)(t, u);} [with T = T]’ depends on itself


mymedia2:
Fail to build autotests of the latest release [v0.11.0](https://github.com/ericniebler/range-v3/releases/tag/0.11.0) with GCC 11.2.0. Could you please take a look? A thousand thanks in advance.

```
[  0%] Building CXX object CMakeFiles/test.range.v3.headers.dir/headers/range/v3/utility.cpp.o
In file included from /build/range-v3/include/range/v3/utility/any.hpp:23,
                 from /build/range-v3/include/range/v3/utility.hpp:17,
                 from /build/range-v3/obj-x86_64-linux-gnu/headers/range/v3/utility.cpp:1:
/build/range-v3/include/concepts/swap.hpp: In substitution of ‘template<class TRef, class T>  requires (copyable<T>) && !(same_as<T, ranges::any>) ranges::any::any(TRef&&) [with TRef = std::default_delete<ranges::any::interface>&; T = std::default_delete<ranges::any::interface>]’:
/build/range-v3/include/concepts/swap.hpp:179:22:   required by substitution of ‘template<class T, class U> decltype (concepts::adl_swap_detail::swap(declval<T>(), declval<U>())) concepts::adl_swap_detail::try_adl_swap_(int) [with T = std::default_delete<ranges::any::interface>&; U = std::default_delete<ranges::any::interface>&]’
/build/range-v3/include/concepts/swap.hpp:186:14:   required from ‘constexpr const bool concepts::adl_swap_detail::is_adl_swappable_v<std::default_delete<ranges::any::interface>&>’
/build/range-v3/include/concepts/swap.hpp:205:18:   required by substitution of ‘template<class T> constexpr meta::if_c<((! is_adl_swappable_v<T&>) && is_movable_v<T>)> concepts::adl_swap_detail::swap_fn::operator()(T&, T&) const [with T = std::default_delete<ranges::any::interface>]’
/build/range-v3/include/concepts/concepts.hpp:937:9:   required by substitution of ‘template<class TRef, class T>  requires (copyable<T>) && !(same_as<T, ranges::any>) ranges::any::any(TRef&&) [with TRef = std::default_delete<ranges::any::interface>&; T = std::default_delete<ranges::any::interface>]’
[...]

PR #1676, opened 2021-11-12
[FIX] circular swap concept

eseiler:
Fixes https://github.com/ericniebler/range-v3/issues/1672

I'm not too sure about whether the replacement is correct (let's see what CI says).

Type is `ranges::any`, and we check for `swappable`.
In the end, we want to use the `any` constructor with `copyable` restriction.

`copyable` requires `movable`, and `movable` requires `swappable`

### DIFF
--- a/include/range/v3/utility/any.hpp
+++ b/include/range/v3/utility/any.hpp
@@ -124,7 +124,7 @@ namespace ranges
         any() noexcept = default;
         template(typename TRef, typename T = detail::decay_t<TRef>)(
             /// \pre
-            requires copyable<T> AND (!same_as<T, any>)) //
+            requires copy_constructible<T> AND (!same_as<T, any>)) //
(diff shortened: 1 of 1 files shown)


include/range/v3/utility/any.hpp
```diff
@@ -124,7 +124,7 @@ namespace ranges
         any() noexcept = default;
         template(typename TRef, typename T = detail::decay_t<TRef>)(
             /// \pre
-            requires copyable<T> AND (!same_as<T, any>)) //
+            requires copy_constructible<T> AND (!same_as<T, any>)) //
```
ericniebler, line 127: I think if the `!same_as<...>` check came before the `copyable` check, that would fix the problem.

ericniebler (changes requested):

PR #1679, opened 2021-11-25
linear_distribute improvement

MaciejPatro:
`linear_distribute` has an unexpected behavior for integral values when number of values to be calculated is greater than distance between `from` and `to`. My expectation was similar to `numpy` behavior where calculated values are "evenly" distributed with integral precision.

Expectation: 
```
linear_distribute(0, 10, 20) == linear_distribute(0.0, 10.0, 20) | transform(convert_to<int>{})
```

Current implementation:
```
linear_distribute(0, 3, 7)     = [0, 0,   0, 0,   1, 2,   3]
linear_distribute(0.0, 3.0, 7) = [0, 0.5, 1, 1.5, 2, 2.5, 3]
```

Improved implementation:

```
linear_distribute(0, 3, 7)     = [0, 0,   1, 1,   2, 2,   3]
linear_distribute(0.0, 3.0, 7) = [0, 0.5, 1, 1.5, 2, 2.5, 3]
```

### DIFF
--- a/include/range/v3/view/linear_distribute.hpp
+++ b/include/range/v3/view/linear_distribute.hpp
@@ -84,7 +84,7 @@ namespace ranges
             constexpr linear_distribute_view(T from, T to, std::ptrdiff_t n) noexcept
               : from_(from)
               , to_(to)
-              , delta_((to - from) / Calc(n - 1))
+              , delta_(n > 1 ? (to - from) / Calc(n - 1) : 0)
               , n_(n)
             {
                 RANGES_EXPECT(n_ > 0);
--- a/include/range/v3/view/linear_distribute.hpp
+++ b/include/range/v3/view/linear_distribute.hpp
@@ -17,6 +17,8 @@
 
 #include <type_traits>
 
+#include <meta/meta.hpp>
+
 #include <range/v3/range_fwd.hpp>
 
 #include <range/v3/iterator/default_sentinel.hpp>
@@ -39,8 +41,10 @@ namespace ranges
 
         private:
             friend range_access;
+            using Calc = meta::conditional_t<std::is_floating_point<T>::value, T, double>;
 
             T from_, to_;
+            Calc delta_;
             std::ptrdiff_t n_;
 
             constexpr T read() const noexcept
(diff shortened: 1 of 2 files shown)


include/range/v3/view/linear_distribute.hpp
```diff
@@ -71,15 +75,16 @@ namespace ranges
...
             }
 
         public:
             constexpr linear_distribute_view() = default;
-            constexpr linear_distribute_view(T from, T to__, std::ptrdiff_t n) noexcept
+            constexpr linear_distribute_view(T from, T to, std::ptrdiff_t n) noexcept
               : from_(from)
-              , to_(to__)
+              , to_(to)
+              , delta_((to - from) / Calc(n - 1))
```
roman-vorobiov, line 87: division by 0 if `n == 1`

MaciejPatro: You are right, I've missed it. It didn't affect the functionality as for n == 1 the delta was not used in calculations.

ericniebler:
Sorry for the delay. Would you be able to fix the precision warning?

```
../include/range/v3/view/linear_distribute.hpp:78:51: error: implicit conversion from 'long' to 'double' may lose precision [-Werror,-Wimplicit-int-float-conversion]
                    from_ = T(to_ - (delta_ * (n_ - 1)));
                                            ~  ~~~^~~
```

And as a heads up, I have rebased this PR. You'll need to fetch it fresh.

include/range/v3/view/linear_distribute.hpp
```diff
@@ -71,15 +75,16 @@ namespace ranges
                 }
                 else
                 {
-                    from_ += (to_ - from_) / T(n_);
+                    from_ = T(to_ - (delta_ * (n_ - 1)));
```
ericniebler, line 78: Maybe?

```suggestion
                    from_ = T(to_ - (delta_ * Calc(n_ - 1)));
```

MaciejPatro: Makes sense - thank you for pointing that out.

PR #1700, opened 2022-05-02
Fixing ranges::to<std::map>(v)

brevzin:
The current overload set is inconsistent:

```cpp
std::vector<std::pair<int, int>> v = {{1, 2}, {3, 4}};

auto m1 = ranges::to<std::map<int, int>>(v); // ok
auto m2 = v | ranges::to<std::map<int, int>>(); // ok
auto m3 = ranges::to<std::map>(v); // compile error
auto m4 = v | ranges::to<std::map>(); // ok
```

`m3` fails because the logic here tries to instantiate `std::map<std::pair<int, int>>` (specifically, `Cont<range_value_t<Rng>>`) instead of actually using the existing deduction machinery (which is what the `m4` overload correctly uses).

### DIFF
--- a/.github/workflows/range-v3-ci.yml
+++ b/.github/workflows/range-v3-ci.yml
@@ -319,37 +319,37 @@ jobs:
             cxx_standard: 20,
           }
 
-        # MSVC 2019
+        # MSVC 2022
         - {
-            name: "Windows MSVC 2019 Debug (C++17)", artifact: "Windows-MSVC.tar.xz",
+            name: "Windows MSVC 2022 Debug (C++17)", artifact: "Windows-MSVC.tar.xz",
             os: windows-latest,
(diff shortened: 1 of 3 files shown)


ericniebler:
Thanks @brevzin. Can you look into the test regressions?

brevzin:
> Thanks @brevzin. Can you look into the test regressions?

Yeah, looks like this was cause of some implementations not having the `std::map` deduction guide (that we technically have a feature test macro for, that nobody actually defines, oh well). You worked around this one clever way with `test_zip_to_map`, I just muscled through making a `std::map`-equivalent. This should work 🤞.

brevzin:
Okay now everything passes, except the windows ones where it just looks like the CI setup is broken.

PR #1701, opened 2022-05-02
Fix name meaning change in nested class

jicama:
C++20 [basic.scope.class]: A name N used in a class S shall refer to the
same declaration in its context and when re-evaluated in the completed scope
of S. No diagnostic is required for a violation of this rule.

This rule is violated by the unqualified use of meta::invoke followed by the
declaration of partition_<Fn>::invoke.


ericniebler:
Thank you!

PR #1730, opened 2022-08-05
Fixing repo when #include <complex.h>

### LINKED ISSUE #1729: Error when compiling with complex.h
Reproducer: https://godbolt.org/z/b89bYKohM

This is due to a collision with the I macro from https://en.cppreference.com/w/c/numeric/complex/I

Similar fix by fmt: https://github.com/fmtlib/fmt/commit/5f774c0aed4af6494b412cd3883e0f09295bd79d

brevzin:
Fixes #1729. Apparently `<complex.h>` just defines a macro named `I`, which is obviously pretty hostile in our world where `I` is the obvious template parameter name for an iterator and is used throughout.


include/range/v3/detail/prologue.hpp
```diff
@@ -45,3 +45,6 @@ RANGES_DIAGNOSTIC_KEYWORD_MACRO
 #define declval(...) static_cast<__VA_ARGS__(*)()>(nullptr)()
 #endif
 /// \cond
+
+#pragma push_macro("I")
```
ericniebler, line 49: Pragmas are all implemented defined. Do we know any compilers that _don't_ support `push/pop_macro`?

brevzin: I tried this out on all clang, gcc, and msvc and it all seemed to work. That definitely seems like a better bet than... renaming all the `I`s. 

If this doesn't work, then uh...

brevzin: lol it looks like the Windows build is failing because just including `<complex.h>` is deprecated? Am I understanding that correctly?

I have no idea why some clang builds are failing and some not.

CaseyCarter: > I have no idea why some clang builds are failing and some not.

libc++ builds are failing, libstdc++ builds are not.

CaseyCarter: > lol it looks like the Windows build is failing because just including `<complex.h>` is deprecated? Am I understanding that correctly?

C++17 said that "`#include <complex.h>` is equivalent to `#include <ccomplex>`", and that `<ccomplex>` "behaves as if it simply includes the header `<complex>`". C++20 said that "`#include <complex.h>` is equivalent to `#include <complex>`", and deprecated `<ccomplex>`. We've implemented the C++17 redirect, and the deprecation from C++20, but apparently not yet the C++20 redirect. (The UCRT never services old releases for anything short of catastrophe, so the redirect will eventually appear in some new version of Windows.)

TLDR: `#include <complex.h>` on Windows includes `<ccomplex>` which tells you that it's deprecated and then includes `<complex>`. In another ten years or so, `#include <complex.h>` on Windows will simply include `<complex>` directly.

CaseyCarter: TLDR: The "bug1729" test _should_ fail and return `1` on C++17-or-later.

CaseyCarter: The fact that it's failing on libc++ in C++14 mode suggests that libc++ implemented the C++17 change unconditionally (as did MSVC). The fact that it passes on libstdc++ suggests that libstdc++ has not implemented the C++17 change.

CaseyCarter: Maybe @jwakely can comment?

brevzin: Well as long as it literally compiles (which it did not before) is a success, so maybe I'll just change the test to `return 0;` and add defines for MSVC to not give me the depreciation warnings?

CaseyCarter: `#define _SILENCE_CXX17_C_HEADER_DEPRECATION_WARNING` before including anything will silence the MSVC deprecation warning.

brevzin: Looks like everything works now 🎉

PR #1752, opened 2022-11-23
Fixing unconditional noexcept in join of glvalue case

### LINKED ISSUE #1751: views::join terminates program because of noexcept clause
The following program terminates abnormally due to exception throw despite the presence of try-catch block: 

```cpp
static std::vector<char> conv(std::vector<char> inp)
{
    return {}; 
}
static std::vector<char> finish()
{
    return {}; 
}
int main()
{
    try {
        std::vector<char> const vi{'a', 'b', 'c', 'd', 'e', 'f'};
        auto rng = ranges::views::concat(vi
              |    ranges::views::chunk(2)
              | ranges::views::transform([](auto&& chs){
                    throw std::runtime_error("");
                    return conv(chs | ranges::to<std::vector<char>>);
              }) | ranges::views::cache1,
                ranges::views::generate_n([]() {
                    return finish();
                }, 1)) 
            | ranges::views::join;
        std::cout << rng << std::endl;
    }   
    catch(...)
    {   
        std::cout << "exception occurred.";
    }   
}
```

This is because of the following code in view/join.hpp:

             // Intentionally promote xvalues to lvalues here:
             template<typename OuterIt>
             static constexpr auto && update_inner_(OuterIt && it) noexcept
             {
                 return *it;
             }

I think it should be:

             // Intentionally promote xvalues to lvalues here:
             template<typename OuterIt>

            static constexpr auto && update_inner_(OuterIt && it) noexcept(noexcept(*it))
             {
                 return *it;
             }

brevzin:
Fixes #1751

### DIFF
--- a/include/range/v3/view/join.hpp
+++ b/include/range/v3/view/join.hpp
@@ -97,7 +97,7 @@ namespace ranges
         {
             // Intentionally promote xvalues to lvalues here:
             template<typename OuterIt>
-            static constexpr auto && update_inner_(OuterIt && it) noexcept
+            static constexpr auto && update_inner_(OuterIt && it) noexcept(noexcept(*it))
             {
                 return *it;
             }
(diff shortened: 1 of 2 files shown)


brevzin:
Fixes #1753

PR #1761, opened 2022-12-20
Don't install Range-v3 when it is used as a subproject

Attractadore:
IMHO installing a subproject with the main project is undesirable, so this patch addresses this issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,27 +162,29 @@ if(NOT RANGE_V3_NEW_VERSION_HPP STREQUAL RANGE_V3_OLD_VERSION_HPP)
   message(STATUS "  conan upload --all range-v3/${RANGE_V3_VERSION}@ericniebler/stable -r=range-v3")
 endif()
 
-include(CMakePackageConfigHelpers)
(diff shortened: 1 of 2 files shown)


brevzin (approved):

loicmouton:
Hello,
After this change I have the following error when using FetchContent to integrate range-v3 into my project :
`CMake Error: install(EXPORT "myProjectTargets" ...) includes target "myTarget" which requires target "range-v3" that is not in any export set.`.

I believe this is due to the introduction of the RANGE_V3_INSTALL as an option dependent on "is_standalone". I believe that this is a great option to have but I also believe it should be a "simple" option that it always available.
If it were a "simple" option, I could simply override its value in my CMake project like I usually do for most CMake libraries I use :
`set("RANGE_V3_INSTALL" ON CACHE BOOL "Enable install rule for range-v3" FORCE)` (before the call to FetchContent).

Furthermore, the same is true for RANGE_V3_TESTS and RANGE_V3_EXAMPLES and, although I have yet no interest in enabling them, someone might want to enable them even if range-v3 is a sub-project.

PR #1276, opened 2019-08-23
Apply [[no_unique_address]] to predicates

morinmorin:
This PR applies `RANGES_NO_UNIQUE_ADDRESS` to functions/predicates in 

- transform, take_while, drop_while views
- `logical_negate`

The latter allows EBO out empty predicates in `filter_view`.

PR #1770, opened 2023-03-01
Re-enable no_unique_address on GCC

morinmorin:
ICE on GCC 10 prerelease has been fixed in GCC 10 release series.

### DIFF
--- a/include/range/v3/detail/config.hpp
+++ b/include/range/v3/detail/config.hpp
@@ -693,9 +693,7 @@ namespace ranges
 #define RANGES_IS_SAME(...) std::is_same<__VA_ARGS__>::value
 #endif
 
-// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93667
-#if defined(__has_cpp_attribute) && __has_cpp_attribute(no_unique_address) && \
-    !(defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 10)
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(no_unique_address)
 #define RANGES_NO_UNIQUE_ADDRESS [[no_unique_address]]
 #else
 #define RANGES_NO_UNIQUE_ADDRESS


brevzin (approved):

brevzin:
Cool, thanks for the catch! Confirmed that the test case in the issue seems to pass on every gcc I tried.

morinmorin:
Thank you for merging!
I tested to measure the size of some views in libstdc++ and libc++ (and range-v3). The tested code is here (slightly modified version of [this code](https://github.com/ericniebler/range-v3/pull/1276#issuecomment-524415141))
```
#define USE_RANGE_V3

#include <iostream>
#ifdef USE_RANGE_V3
#include <range/v3/core.hpp>
#include <range/v3/view.hpp>
#else
#include <ranges>
#endif

struct F
{
    bool operator()(int x) const { return x % 2 == 0; }
    int operator()(int x, int y) const { return x + y; }
};

int main()
{
#ifdef USE_RANGE_V3
    using namespace ranges::views;
#else
    using namespace std::ranges::views;
#endif

    int arr[3] = {4, 5, 6};
    
    std::cout << "arr | transform" << "\n";
    std::cout << sizeof(arr | transform(F{})) << "\n";

    std::cout << "arr | transform | transform" << "\n";
    std::cout << sizeof(arr | transform(F{}) | transform(F{})) << "\n";

    std::cout << "arr | transform | transform | transform" << "\n";
    std::cout << sizeof(arr | transform(F{}) | transform(F{}) | transform(F{})) << "\n";

    // std::cout << "transform (binary)" << "\n";
    // std::cout << sizeof(transform(arr, arr, F{})) << "\n";

    std::cout << "arr | take_while" << "\n";
    std::cout << sizeof(arr | take_while(F{})) << "\n";

    std::cout << "arr | drop_while" << "\n";
    std::cout << sizeof(arr | drop_while(F{})) << "\n";

    std::cout << "arr | filter" << "\n";
    std::cout << sizeof(arr | filter(F{})) << "\n";

    // std::cout << "arr | remove_if" << "\n";
    // std::cout << sizeof(arr | remove_if(F{})) << "\n";
}
```
Then, I noticed that views in range-v3 on GCC have strangely large size, which turned into this PR.

I also noticed that views' size in range-v3 is a bit larger than those of libstdc++ and libc++.

|  views | libstdc++ | libc++ | range-v3 |
|--------|:---------:|:------:|:--------:|
|<code>arr &#124; transform</code>| 8| 8|16|
|<code>arr &#124; transform &#124; transform</code>|16|16|24|
|<code>arr &#124; transform &#124; transform &#124; transform</code>|24|24|32|
|<code>arr &#124; take_while</code>| 8| 8|16|
|<code>arr &#124; drop_while</code>|16|24|32|
|<code>arr &#124; filter</code>|16|24|32|

PR #1773, opened 2023-03-13
Adapted MSVC std::span<T> (_Span_iterator) for ranges::contiguous_ite…

refvalue:
…rator.


include/range/v3/iterator/concepts.hpp
```diff
@@ -95,6 +95,10 @@ namespace ranges
             requires same_as<I, class I::_String_view_iterator>)
         auto iter_concept_(I, priority_tag<3>)
             -> ranges::contiguous_iterator_tag;
+		template(typename I)(
```
brevzin, line 98: Can you fix the formatting here?

refvalue: It has been resolved. Please review.

brevzin (approved):

PR #1775, opened 2023-04-21
Rename Nil because it conflicts with Apple headers

rnikander:
There is a `#define Nil` buried in Apple's system headers and it breaks range-v3.


brevzin (approved):
I have no idea how any of this stuff works, but I can confirm that these are all 10 `Nil`s - pretty obnoxious to have `#define Nil` though, thanks Apple. 

Hard to disapprove with this name too....

PR #1776, opened 2023-04-23
Fix zip/zip_with distance_to

PkXwmpgN:
The distance between iterators is wrongly calculated when the first zipped range is empty:

```cpp
    std::vector<int> a = {1, 2, 3, 4, 5};
    std::vector<int> b;

    auto z0 = zip(b, a);
    auto z1 = zip(a, b);

    auto d0 = ranges::distance(ranges::begin(z0), ranges::end(z0));
    auto d1 = ranges::distance(ranges::begin(z1), ranges::end(z1));
    
    assert(d0 == 0); // failed d0 == 5
    assert(d1 == 0); // pass
 ```
 With MSVC
 ```cpp
    std::vector<int> a = {1, 2, 3, 4, 5};
    std::vector<int> b; 
    auto v = zip(b, a) | ranges::to_vector; // access violation
```

### DIFF
--- a/include/range/v3/view/zip_with.hpp
+++ b/include/range/v3/view/zip_with.hpp
@@ -282,7 +282,11 @@ namespace ranges
             {
                 // Return the smallest distance (in magnitude) of any of the iterator
                 // pairs. This is to accommodate zippers of sequences of different length.
-                if(0 < std::get<0>(that.its_) - std::get<0>(its_))
+                auto first_size = std::get<0>(that.its_) - std::get<0>(its_);
+                if(!first_size)
+                    return static_cast<difference_type>(0);
+
+                if(0 < first_size)
                     return tuple_foldl(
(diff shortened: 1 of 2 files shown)


include/range/v3/view/zip_with.hpp
```diff
@@ -282,7 +282,11 @@ namespace ranges
             {
                 // Return the smallest distance (in magnitude) of any of the iterator
                 // pairs. This is to accommodate zippers of sequences of different length.
-                if(0 < std::get<0>(that.its_) - std::get<0>(its_))
+                auto first_size = std::get<0>(that.its_) - std::get<0>(its_);
+                if(!first_size)
```
brevzin, line 286:
```suggestion
                if(first_size == 0)
```

brevzin (approved):

brevzin (approved):

brevzin:
Thanks for the fix!

PR #1784, opened 2023-05-25
Minor change to avoid 'unused variable' compiler warning/error

roibarkan:
fixes many of the github actions failures.
2 macOs github actions still remain as failures with segFault in example/range.v3.comprehensions


CaseyCarter:
Looks like this problem was already resolved by #1812.

PR #1799, opened 2023-11-02
Add missing calls to test_result()

serpent7776:
Some tests lack a call to `test_result`. In effect, these tests could fail without setting nonzero exit code.


brevzin (approved):
Something something rolling your own test framework...

Thanks for catching this!

PR #1800, opened 2023-11-03
Enumerate with arbitrary starting index

serpent7776:
Improves `enumerate` so that it's possible to start at any given index. Similar functionality exists in [boost::adaptors::indexed](https://www.boost.org/doc/libs/1_75_0/libs/range/doc/html/range/reference/adaptors/reference/indexed.html)

This is just a draft, it just adds an additional parameter to `enumerate_fn::operator()`. It works well when used with function call syntax, but:
- doesn't work with pipe syntax without breaking existing user code (big issue)
- increases size of `index_view` (not sure if an issue)
- breaks existing ABI of `enumerate` (probably not an issue)

Another way to implement this would be to transform constructed `zip`, but I'm not sure that's any better (also doesn't fix the first issue).

### DIFF
--- a/include/range/v3/view/enumerate.hpp
+++ b/include/range/v3/view/enumerate.hpp
@@ -35,6 +35,7 @@ namespace ranges
         {
         private:
             friend range_access;
+            size_t start_{0};
 
             struct cursor
             {
@@ -72,10 +73,12 @@ namespace ranges
 
             public:
                 cursor() = default;
+                cursor(size_t n) : index_{n}
+                {}
             };
             cursor begin_cursor() const
             {
-                return cursor{};
+                return cursor{start_};
             }
             unreachable_sentinel_t end_cursor() const
             {
@@ -84,6 +87,8 @@ namespace ranges
 
         public:
             index_view() = default;
+            index_view(size_t n) : start_{n}
+            {}
         };
 
     } // namespace detail
@@ -103,11 +108,11 @@ namespace ranges
         {
             template(typename Rng)(
                 requires viewable_range<Rng>)
-            auto operator()(Rng && rng) const
+            auto operator()(Rng && rng, size_t n = 0) const
             {
                 using D = range_difference_t<Rng>;
                 using S = detail::iter_size_t<iterator_t<Rng>>;
-                return zip(detail::index_view<S, D>(), all(static_cast<Rng &&>(rng)));
+                return zip(detail::index_view<S, D>(n), all(static_cast<Rng &&>(rng)));
             }
         };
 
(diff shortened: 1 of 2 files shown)


JohelEGP:
> doesn't work with pipe syntax without breaking existing user code (big issue)

How come?

> increases size of `index_view` (not sure if an issue)
> breaks existing ABI of `enumerate` (probably not an issue)

You can avoid those.

For both, you can have `enumerate(range, size)` return `drop(enumerate(range), size)`.

For only the first one, you can parameterize the size.
Default the size to `std::integral_constant<std::size_t, 0>` and
store it in a base or use `[[no_unique_address]]`.

serpent7776:
> How come?

When I tried modifying the last test to use enumerate with custom index start I got compilation error
```
enumerate.cpp:157:29: error: no match for call to ‘(const ranges::views::view_closure<ranges::views::enumerate_fn>) (int)’                                                   
  157 |           | views::enumerate(99); 
```
I'm assuming this is because `enumerate` is defined to be `view_closure` that doesn't accept parameters on `operator()`:
https://github.com/ericniebler/range-v3/blob/97452bb3eb74a73fc86504421a6a27c92bce6b99/include/range/v3/view/enumerate.hpp#L116
I could change that to
```cpp
RANGES_INLINE_VARIABLE(enumerate_fn, enumerate)
```
and add overload `operator()(size_t n = 0)` to `enumerate_fn`, but that breaks existing code, because `range | enumerate` no longer works and `range | enumerate()` is required.

> For both, you can have `enumerate(range, size)` return `drop(enumerate(range), size)`.

This is interesting, but I'm assuming you meant  `drop(detail::index_view<S, D>(), n)`

> For only the first one, you can parameterize the size.

I don't think I understand.

JohelEGP:
> > For only the first one, you can parameterize the size.
> 
> I don't think I understand.

I meant something similar to how the bound of `std::ranges::iota_view` defaults to `std::unreachable_sentinel_t`.
But in this case, you'd be defaulting the start index to `std::integral_constant<std::ranges::range_difference_t<R>, 0>`.

serpent7776:
I updated the code, so that it works with both function call syntax and piping syntax. I looked at existing code in `split_when.hpp`.

serpent7776:
I'd like to ask if the existing solution is OK or should I go with eliminating the size overhead of `index_view`

serpent7776:
ping @JohelEGP ?

JohelEGP:
Yes?
I'm just a passing contributor.
I have no say in what gets merged,
or what the correct ways are.

JohelEGP:
I think you want feedback from a maintainer now.

PR #1831, opened 2024-09-03
Include string class in primitive.hpp

Mathu-lmn:
Some compilers throw an error when compiling - See : https://github.com/citizenfx/fivem/pull/2763

The easy fix is to add the `#include <string>` as some compilers / environnement are not including it, leading to an `error C2039: 'basic_string' is not a member of 'std'`

### DIFF
--- a/include/range/v3/range/primitives.hpp
+++ b/include/range/v3/range/primitives.hpp
@@ -25,6 +25,8 @@
 
 #include <range/v3/detail/prologue.hpp>
 
+#include <string>
+
 namespace ranges
 {
     /// \addtogroup group-range
--- a/include/range/v3/range/primitives.hpp
+++ b/include/range/v3/range/primitives.hpp
@@ -25,7 +25,9 @@
 
 #include <range/v3/detail/prologue.hpp>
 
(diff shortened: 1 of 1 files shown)


include/range/v3/range/primitives.hpp
```diff
@@ -25,6 +25,8 @@
 
 #include <range/v3/detail/prologue.hpp>
 
+#include <string>
```
CaseyCarter, line 28: The only reference to `basic_string` in this header is guarded by `#if RANGES_CXX_STD <= RANGES_CXX_STD_14`, so presumably this could be guarded similarly:
```suggestion
#if RANGES_CXX_STD <= RANGES_CXX_STD_14
#include <string>
#endif
```

CaseyCarter (changes requested):

include/range/v3/range/primitives.hpp
Mathu-lmn: You're correct ! I commited this change thank you

CaseyCarter (approved):

CaseyCarter:
The CI is hopelessly out-of-date, I'm going to merge this as-is.

PR #1833, opened 2024-09-03
P2770R0 makes stashing `regex_token_iterator` input

CaseyCarter:
... instead of lying about being bidi.

### DIFF
--- a/test/view/tokenize.cpp
+++ b/test/view/tokenize.cpp
@@ -4,6 +4,16 @@
 #include "../simple_test.hpp"
 #include "../test_utils.hpp"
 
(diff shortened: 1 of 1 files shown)


CaseyCarter:
The CI failures appear unrelated. It looks like apple clang has decided to start emitting deprecation warnings for out-of-class definitions of constexpr static data members: "out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated [-Werror,-Wdeprecated]".

ericniebler:
dude sweet

CaseyCarter:
> dude sweet

image

PR #1838, opened 2024-10-01
Fix for C++23 implicit move

jicama:
Under [P2266](https://wg21.link/p2266r3) 't' by itself is move-eligible, so we need to obfuscate it to get an lvalue.


CaseyCarter (approved):
Thanks, Jicama!

PR #1843, opened 2024-12-11
Add support for move-only `view`s

joka921:
In C++20, a `std::ranges::view` either has to be copyable in constant time, or a move-only type. In `range-v3`, the `view` concept currently requires the view to be copyable. This PR lifts this restriction, and makes the `::ranges::view` concept idential to `std::ranges::view`.

This allows the implementation of an `owning_view` that can take ownership of a range that is expensive to copy.
While the `owning_view` can be implemented outside of `range-v3`, the proposed (minimal) changes to the `view` concept are necessary as a first step to bring `range-v3` and the C++20 `std::ranges` closer to each other.

### DIFF
--- a/include/range/v3/range/concepts.hpp
+++ b/include/range/v3/range/concepts.hpp
@@ -251,7 +251,7 @@ namespace ranges
     template<typename T>
     CPP_concept view_ =
         range<T> &&
-        copyable<T> &&
+        movable<T> &&
         enable_view<T>;
 
     /// \concept viewable_range
--- a/test/utility/concepts.cpp
+++ b/test/utility/concepts.cpp
@@ -318,6 +318,14 @@ struct myview : ranges::view_base {
 };
 CPP_assert(ranges::view_<myview>);
 
+struct move_only_view : myview {
+    move_only_view(const move_only_view&) = delete;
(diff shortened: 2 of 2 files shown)


Alex2772:
Hi,

Do I understand correctly, are you trying to address this issue?

```cpp
#include <ranges>
#include <range/v3/all.hpp>

int main() {
    std::vector<int> ints = {1, 3, 4};
    std::move(ints) | std::views::filter([](int i) { return true; });    // ok
    std::move(ints) | ranges::views::filter([](int i) { return true; }); // compile error * READ THIS *
}
```

I've encountered this recently, after discovering that `ranges-v3` does not provide its own `owning_view` implementation. It seems like the lack of `owning_view` is a top of iceberg.

joka921:
@Alex2772 Yes, exactly.
To make your example work we need
* this PR with the movable `view` concept
* An implementation of `owning_view`
* Extend the implementation of `views::all[_t]` to use `owning_view` for rvlaues of `ranges::range`s.

We currently have the `owning_view` in our own library, but I am planning to move it to our fork of `rage-v3` s.t. it becomes closer to a drop-in replacement of `std::ranges`.

Issue #1839, opened 2024-10-20
CMake version 3.6 is getting outdated.


christianbrugger:
When configuring range-v3 with the new `cmake 3.31` I get the following warning:

```cmake
CMake Deprecation Warning at external/range-v3/CMakeLists.txt:8 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- [range-v3]: C++ std=latest
```

As stated in #1724 the actual needed feature set of the `CMakeLists.txt` seems to be at least 3.15. In this case I suggest to change this line
```cmake
cmake_minimum_required(VERSION 3.6)
```
simply to
```cmake
cmake_minimum_required(VERSION 3.15)
```

This would also close #1724

Issue #1724, opened 2022-07-13
cmake change requires version 3.15 but did not bump cmake_minimum_required from 3.6


KarateSnowMachine:
[This commit](https://github.com/ericniebler/range-v3/commit/3a0881685803c659eba945788f9fe81acfe51c8a) introduces `COMPILE_LANG_AND_ID` which was added cmake 3.15. However, this change did not bump `cmake_minimum_required` (currently at 3.6) and thus is causing build issues for me. While I have an easy workaround for my own use-case, it would be nice to fix this so it doesn't break downstream projects.

alexreinking:
I wrote a blog post about this class of bug. This is #1 on the list:

> You must also test with the actual CMake 3.X release on your local development machine and on CI.
>
> Why? Simply because the policy mechanism ensures backwards compatibility, not forwards compatibility. If you use a more recent CMake version, nothing will stop you from using a feature that is too new for the declared minimum version. This is very, very common, too. Here are three examples off the top of my head that have bitten me:
>
> * You might use a generator expression that was not in the old CMake version. CMake will not even try to warn you about this, and many common and useful generator expressions were introduced later than you think.

https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-1.html

brevzin:
@alexreinking So is this fix just changing the minimum version to 3.15 or what? I don't have a ton of cmake versions on my box, and the CI job here is more concerned about testing a lot of compilers.

alexreinking:
I would suggest bumping the minimum version to 3.15 (at least, it's getting to be a bit old at this point) and then adding a single CI job that tests CMake 3.15.

PR #1851, opened 2025-05-31
Bump CMake minimum version

wermos:
Addresses issues #1839 and #1724.

I bumped the CMake minimum version to 3.15 as suggested in those issues.

I noticed that the version of CMake used in the CI is [`3.16.2`](https://github.com/ericniebler/range-v3/blob/ca1388fb9da8e69314dda222dc7b139ca84e092f/.github/workflows/range-v3-ci.yml#L7), so we can also change the minimum required version to that instead. That would also be more robust.


ericniebler (approved):

jmcarcell:
Can this be merged? CMake 4.X will emit warnings if the minimum required is less than 3.10.

PR #1859, opened 2025-11-20
Suppress clang deprecation warning in make_compressed_tuple.

TryExceptElse:
Previously, modern clang versions threw warnings due to the use of the deprecated `compressed_tuple` within `make_compressed_tuple`. This caused problems when ranges-v3 or a dependant project included the `ranges/v3/utility/compressed_pair.hpp` header, either directly or indirectly, via `ranges/v3/all.hpp`. This commit suppresses the deprecated-declarations warning within the `make_compressed_tuple` implementation, and adds a deprecation annotation to the former struct, achieving the intended goal of deprecating these structs without disrupting projects that do not use `compressed_tuple` or `make_compressed_tuple`.

The deprecation warning fixed by this commit seems to be unique to newer clang versions, not present in clang 10, but introduced before (or in) clang-19. It does not occur when using gcc (tested 7 and 13).

### DIFF
--- a/include/range/v3/utility/compressed_pair.hpp
+++ b/include/range/v3/utility/compressed_pair.hpp
@@ -105,6 +105,7 @@ namespace ranges
     {
         // clang-format off
         template<typename... Args>
+        RANGES_DEPRECATED("ranges::compressed_tuple is deprecated.")
         constexpr auto CPP_auto_fun(operator())(Args &&... args) (const)
         (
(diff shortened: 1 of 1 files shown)


TryExceptElse:
I've rebased this on top of c1ed9bf9 since that commit predated this work and because it compiles cleanly with both gcc and clang (tested using gcc 13 and clang 19).

PR #1860, opened 2026-01-28
declare dependency on rules_cc and import cc_library

marcushultman:
This is required in Bazel 9


marcushultman:
@ericniebler

ericniebler:
thanks, sorry for the lag

PR #1864, opened 2026-07-27
Avoid turning off warnings in external code.

vchigrin:


include/concepts/swap.hpp
vchigrin, line 160: Here some compiler warnings are turned off. Without `CPP_DIAGNOSTIC_PUSH`/`CPP_DIAGNOSTIC_POP` around this file, that caused unintended warning suppresion in external, user code, that used this library.

